PR #67, opened 2022-08-31
[Setup] Swap `AnyView` to `@ViewBuilder`

yo1995:
## Description

This PR proposes to remove `AnyView` and use `@ViewBuilder` instead, so that we don't need to deal with type-erased `AnyView` which Apple suggests against: https://developer.apple.com/videos/play/wwdc2021/10022/?time=701

## Linked Issue(s)

- `swift/issues/2300`

## How To Test

App works the same as before.

## To Discuss

See below

### DIFF
--- a/Scripts/GenerateSampleViewSourceCode.swift
+++ b/Scripts/GenerateSampleViewSourceCode.swift
@@ -112,7 +112,7 @@ private let sampleStructs = sampleMetadata
 private let sampleViewCases = sampleMetadata
     .map { sample in
         return """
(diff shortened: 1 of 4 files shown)


yo1995:
After some experiments in `swift/issues/2300`, I decided this change is not worth making in the app, as it brings minimal benefits, if beneficial at all. Using `AnyView`, while is not ideal in from a code smell perspective, actually work OK in our situation and keeps the code easy to reason about.

I'll close this PR for now.

PR #89, opened 2022-12-01
[Fix] Raster file load issue

### LINKED ISSUE #48: Fix raster file download problem in `Add raster from file`
Problem was in #46 

See https://github.com/Esri/arcgis-runtime-samples-ios/pull/1036

The problem's behavior is that when the raster file is downloaded for the first time, it would throw a build failure due to the file not found error. In the following builds, the error is gone.

It may be due to the raster file rely on a certain folder structure, and we have to move back to the old raster portal item. See `common-samples/issues/3618` for the data change.

yo1995:
## Description

This PR fixes #48.

## Linked Issue(s)

- #48 

## How To Test

1. Remove the app from the simulator and clean build folder, to make sure it is a fresh install
2. Open the `Add raster from file` sample
3. It crashes on `v.next`, doesn't crash on this branch

## To Discuss

I realized the static method might be initialized too early, which causes the view struct not to have access to the file.

### DIFF
--- a/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
+++ b/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
@@ -16,23 +16,8 @@ import ArcGIS
 import SwiftUI
 
 struct AddRasterFromFileView: View {
-    /// Makes a map with a raster layer.
-    private static func makeMap() -> Map {
-        /// A map with a standard imagery basemap style.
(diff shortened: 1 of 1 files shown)


yo1995:
@dg0yal encountered the crash for the TestFlight beta, but I didn't. So it might be something related to timing. 🤔

yo1995:
This will be part of #107

PR #96, opened 2022-12-07
[Fix] Screenshot path for `Project geometry`

yo1995:
## Description

This PR fixes the path to the screenshot. Thanks @maryharvey !
preview: https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Ting/Fix-Screenshot/Shared/Samples/Project%20geometry/README.md

### DIFF
--- a/Shared/Samples/Project geometry/README.md
+++ b/Shared/Samples/Project geometry/README.md
@@ -2,7 +2,7 @@
 
 Project a point to another spatial reference.
 
-![Screenshot of project sample](project.png)
+![Screenshot of project geometry sample](project-geometry.png)
 
 ## Use case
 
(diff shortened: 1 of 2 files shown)


yo1995:
I realized that I didn't check the image paths because they are populated from the actual file paths on disk into the metadata. But there is a case like this if the metadata is generated first, and later on the sample name is changed, the checker will miss the image names.

I consider this as a lower risk edge case, so will fix when we need to do other stuff with the status checker.

vquach2404 (approved):
LGTM

saratchandrakarumuri (approved):

yo1995:
Here is another occurrence of this situation: https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/151#discussion_r1156558804

PR #107, opened 2023-01-04
[Update] Use model to manage geo view and data

yo1995:
## Description

This PR changes the `@StateObject` pattern in most of the samples, to use an observable model object to manage the data.

## Linked Issue(s)

- `swift/issues/3271`
- `swift/issues/3250`
- https://github.com/ArcGIS/arcgis-maps-sdk-swift-toolkit/issues/200
- Potentially #48 and #89 

## How To Test

To run the app, you'll need to

1. replace the toolkit dependency line in Package.swift locally with `.package(name: "arcgis-maps-sdk-swift", path: "../swift/ArcGIS")`.
2. drag the toolkit into the samples project, so it overrides the remote resolved packages and uses the local toolkit repo instead.

No behavioral change should present. Verify all changed sample to work as expected.

## To Discuss

- What should be a good generic docstring for the view model objects?

### DIFF
--- a/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
+++ b/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
@@ -16,23 +16,25 @@ import ArcGIS
 import SwiftUI
 
 struct AddRasterFromFileView: View {
-    /// Makes a map with a raster layer.
-    private static func makeMap() -> Map {
-        /// A map with a standard imagery basemap style.
-        let map = Map(basemapStyle: .arcGISImageryStandard)
-        // Gets the Shasta.tif file URL.
-        let shastaURL = Bundle.main.url(forResource: "Shasta", withExtension: "tif", subdirectory: "raster-file/raster-file")!
-        // Creates a raster with the file URL.
-        let raster = Raster(fileURL: shastaURL)
-        // Creates a raster layer using the raster object.
-        let rasterLayer = RasterLayer(raster: raster)
-        // Adds the raster layer to the map's operational layer.
-        map.addOperationalLayer(rasterLayer)
-        return map
+    private class Model: ObservableObject {
+        /// A map with imagery basemap and a raster layer.
+        let map: Map = {
+            /// A map with a standard imagery basemap style.
+            let map = Map(basemapStyle: .arcGISImageryStandard)
+            // Gets the Shasta.tif file URL.
+            let shastaURL = Bundle.main.url(forResource: "Shasta", withExtension: "tif", subdirectory: "raster-file/raster-file")!
+            // Creates a raster with the file URL.
+            let raster = Raster(fileURL: shastaURL)
+            // Creates a raster layer using the raster object.
+            let rasterLayer = RasterLayer(raster: raster)
+            // Adds the raster layer to the map's operational layer.
+            map.addOperationalLayer(rasterLayer)
+            return map
+        }()
     }
     
-    /// A map with a standard imagery basemap style.
-    @StateObject private var map = makeMap()
+    /// The view model for the sample.
+    @StateObject private var model = Model()
     
     /// A Boolean value indicating whether to show an alert.
     @State private var isShowingAlert = false
@@ -47,12 +49,12 @@ struct AddRasterFromFileView: View {
     
     var body: some View {
         // Creates a map view with a viewpoint to display the map.
-        MapView(map: map, viewpoint: viewpoint)
+        MapView(map: model.map, viewpoint: viewpoint)
             .onViewpointChanged(kind: .centerAndScale) { viewpoint = $0 }
             .alert(isPresented: $isShowingAlert, presentingError: error)
             .task {
                 do {
-                    let rasterLayer = map.operationalLayers.first!
+                    let rasterLayer = model.map.operationalLayers.first!
                     try await rasterLayer.load()
                     viewpoint = Viewpoint(center: rasterLayer.fullExtent!.center, scale: 8e4)
                 } catch {
--- a/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
+++ b/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
@@ -18,7 +18,9 @@ import SwiftUI
 struct AddRasterFromFileView: View {
     private class Model: ObservableObject {
         /// A map with imagery basemap and a raster layer.
-        let map: Map = {
+        let map: Map
+        
+        init() {
             /// A map with a standard imagery basemap style.
             let map = Map(basemapStyle: .arcGISImageryStandard)
             // Gets the Shasta.tif file URL.
@@ -29,8 +31,8 @@ struct AddRasterFromFileView: View {
             let rasterLayer = RasterLayer(raster: raster)
             // Adds the raster layer to the map's operational layer.
             map.addOperationalLayer(rasterLayer)
-            return map
-        }()
+            self.map = map
+        }
     }
     
     /// The view model for the sample.
--- a/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
+++ b/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
@@ -18,21 +18,7 @@ import SwiftUI
 struct AddRasterFromFileView: View {
     private class Model: ObservableObject {
         /// A map with imagery basemap and a raster layer.
-        let map: Map
-        
-        init() {
-            /// A map with a standard imagery basemap style.
-            let map = Map(basemapStyle: .arcGISImageryStandard)
-            // Gets the Shasta.tif file URL.
-            let shastaURL = Bundle.main.url(forResource: "Shasta", withExtension: "tif", subdirectory: "raster-file/raster-file")!
-            // Creates a raster with the file URL.
-            let raster = Raster(fileURL: shastaURL)
-            // Creates a raster layer using the raster object.
-            let rasterLayer = RasterLayer(raster: raster)
-            // Adds the raster layer to the map's operational layer.
-            map.addOperationalLayer(rasterLayer)
-            self.map = map
-        }
+        let map = Map(basemapStyle: .arcGISImageryStandard)
     }
     
     /// The view model for the sample.
@@ -56,8 +42,15 @@ struct AddRasterFromFileView: View {
             .alert(isPresented: $isShowingAlert, presentingError: error)
             .task {
                 do {
-                    let rasterLayer = model.map.operationalLayers.first!
+                    // Gets the Shasta.tif file URL.
+                    let shastaURL = Bundle.main.url(forResource: "Shasta", withExtension: "tif", subdirectory: "raster-file/raster-file")!
+                    // Creates a raster with the file URL.
+                    let raster = Raster(fileURL: shastaURL)
+                    // Creates a raster layer using the raster object.
+                    let rasterLayer = RasterLayer(raster: raster)
                     try await rasterLayer.load()
+                    // Adds the raster layer to the map's operational layer.
+                    model.map.addOperationalLayer(rasterLayer)
                     viewpoint = Viewpoint(center: rasterLayer.fullExtent!.center, scale: 8e4)
                 } catch {
                     // Presents an error message if the raster fails to load.
--- a/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
+++ b/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
@@ -16,6 +16,8 @@ import ArcGIS
 import SwiftUI
 
 struct AddRasterFromFileView: View {
+    /// The model used to store the geo model and other expensive objects
+    /// used in this view.
     private class Model: ObservableObject {
         /// A map with imagery basemap and a raster layer.
         let map = Map(basemapStyle: .arcGISImageryStandard)
--- a/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
+++ b/Shared/Samples/Add raster from file/AddRasterFromFileView.swift
@@ -43,6 +43,7 @@ struct AddRasterFromFileView: View {
             .onViewpointChanged(kind: .centerAndScale) { viewpoint = $0 }
             .alert(isPresented: $isShowingAlert, presentingError: error)
             .task {
+                guard model.map.operationalLayers.isEmpty else { return }
                 do {
                     // Gets the Shasta.tif file URL.
                     let shastaURL = Bundle.main.url(forResource: "Shasta", withExtension: "tif", subdirectory: "raster-file/raster-file")!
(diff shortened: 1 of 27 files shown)


Shared/Samples/Add raster from file/AddRasterFromFileView.swift
```diff
@@ -16,23 +16,27 @@ import ArcGIS
...
-        return map
+    private class Model: ObservableObject {
+        /// A map with imagery basemap and a raster layer.
+        let map: Map
+        
+        init() {
+            /// A map with a standard imagery basemap style.
+            let map = Map(basemapStyle: .arcGISImageryStandard)
+            // Gets the Shasta.tif file URL.
+            let shastaURL = Bundle.main.url(forResource: "Shasta", withExtension: "tif", subdirectory: "raster-file/raster-file")!
```
yo1995, line 27: Instead of [setting](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html#ID232) the default property value with a closure (which is effectively similar to a static func), I moved the raster into the initializer. It might solve the problem in #48 .

Shared/Samples/Clip geometry/ClipGeometryView.swift
```diff
@@ -91,6 +39,70 @@ struct ClipGeometryView: View {
...
+        )
+        
+        /// The graphics overlay containing clipped graphics.
+        private let clippedGraphicsOverlay = GraphicsOverlay()
+        
+        /// The unclipped graphic of Colorado.
+        private var coloradoGraphic: Graphic { coloradoGraphicsOverlay.graphics.first! }
+        
+        /// The graphics overlays used in this sample.
+        var graphicsOverlays: [GraphicsOverlay] {
```
yo1995, line 73: In the following samples, if one uses multiple graphics overlays, I use a computed property to expose them to the view, instead of enumerate them such as model.overlay1, model.overlay2, etc.

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
```diff
@@ -96,11 +116,20 @@ struct DownloadVectorTilesToLocalCacheView: View {
                                 guard let extent = mapView.envelope(fromViewRect: viewRect) else { return }
                                 
                                 // Downloads the vector tiles.
-                                await model.downloadVectorTiles(extent: extent)
-                                // Sets downloading to false when the download finishes.
-                                isDownloading = false
+                                do {
```
yo1995, line 119: I move the error handling into the view's task modifier so that the model doesn't need to know about a view. See discussion here: https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/42#discussion_r942747195

Shared/Samples/Show callout/ShowCalloutView.swift
```diff
@@ -40,8 +45,8 @@ struct ShowCalloutView: View {
                     Text("Location")
                         .font(.headline)
                     Text(
-                        CoordinateFormatter.toLatitudeLongitude(
-                            point: callout.location,
+                        CoordinateFormatter.latitudeLongitudeString(
```
yo1995, line 48: This is a recent change in the API. In order to make this branch build I included it in this PR. Let me know if you want it in a separate PR.

Shared/Samples/Show device location/ShowDeviceLocationView.swift
```diff
@@ -17,15 +17,32 @@ import CoreLocation
...
+    
     /// The view model for this sample.
     @StateObject private var model = Model()
     
     var body: some View {
         MapView(map: model.map)
             .locationDisplay(model.locationDisplay)
             .task {
-                await model.startLocationDataSource()
-                await model.updateAutoPanMode()
```
yo1995, line 28: This was a trivial bug in this sample. Before this change the task modifier will get stuck, waiting for the auto pan mode stream to end indefinitely. Now it is wrapped in a Task block and has the same lifespan as the model object.

Shared/Samples/Style graphics with renderer/StyleGraphicsWithRendererView.swift
```diff
@@ -16,188 +16,195 @@ import SwiftUI
...
-            spatialReference: spatialReference
-        )!
+        /// The graphics overlay rendering an ellipse.
+        private let ellipseGraphicsOverlay: GraphicsOverlay = {
+            // Creates a simple fill symbol.
+            let symbol = SimpleFillSymbol(color: .purple)
+            // Defines the center point of the ellipse.
+            let center = Point(x: 40e5, y: 25e5, spatialReference: .webMercator)
+            // Creates the parameters for the ellipse.
+            let parameters = GeodesicEllipseParameters<Polygon>(
```
yo1995, line 110: This is a change in `swift/pull/2413`. Without specifying `Polygon` type constraint the returned geometry is a `Multipoint`. Will discuss the reason.

Shared/Samples/Show callout/ShowCalloutView.swift
philium: It would be better to create a PR branch with fixes for API changes and then base this branch off of that one. That way, those fixes get in sooner since this PR will take longer to review.

Shared/Samples/Add raster from file/AddRasterFromFileView.swift
dfeinzimer: I got a crash here opening the sample for the first time.

dfeinzimer:
I'm seeing two close buttons in the Floor Filter UI in "Browse building floors" sample. Not sure what would cause that but I will investigate it on the toolkit side under [#224](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/244).

Shared/Samples/Add raster from file/AddRasterFromFileView.swift
yo1995: 🤦  My speculation is wrong again.

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
```diff
@@ -22,6 +22,20 @@ struct DownloadVectorTilesToLocalCacheView: View {
...
     @State private var isCancellingJob = false
     
+    /// A Boolean value indicating whether to show an alert.
+    @State private var isShowingAlert = false
+    
+    /// A Boolean value indicating whether the download button is disabled.
+    @State private var isDownloadDisabled = true
+    
+    /// A Boolean value indicating whether to show the result map.
+    @State private var isShowingResults = false
```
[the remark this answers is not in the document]
yo1995: I'll address this suggestion in this PR, but leave other boolean renaming in #111 , as it is not totally related to the view model change.

Shared/Samples/Show device location/ShowDeviceLocationView.swift
yo1995: I made some changes here, please take another look in the next round of review.

Shared/Samples/Show result of spatial relationships/ShowResultOfSpatialRelationshipsView.swift
```diff
@@ -31,82 +33,34 @@ struct ShowResultOfSpatialRelationshipsView: View {
...
+              let allRelationships = model.allRelationships(for: selectedGeometry),
               let mapPoint = mapPoint else {
             return
         }
         
         // Updates the relationships.
-        relationships = allRelationships
+        model.relationships = allRelationships
         // Updates the location callout placement.
         calloutPlacement = LocationCalloutPlacement(location: mapPoint)
```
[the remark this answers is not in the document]
yo1995: Agreed. Even though it doesn't visually appear, not updating the `relationships` array is not a good practice as the data source and UI is out of sync. Will make adjustments here.

Shared/Samples/Select features in feature layer/SelectFeaturesInFeatureLayerView.swift
```diff
@@ -89,6 +75,26 @@ struct SelectFeaturesInFeatureLayerView: View {
...
+        let gdpPerCapitaLayer = FeatureLayer(
+            item: PortalItem(
+                portal: .arcGISOnline(connection: .anonymous),
+                id: .gdpPerCapita
+            )
+        )
+
+        /// A map with a topographic basemap style and a feature layer.
+        let map: Map
+
```
dfeinzimer, line 90:
```suggestion
        
```

Shared/Samples/Select features in feature layer/SelectFeaturesInFeatureLayerView.swift
```diff
@@ -89,6 +75,26 @@ struct SelectFeaturesInFeatureLayerView: View {
...
+private extension SelectFeaturesInFeatureLayerView {
+    class Model: ObservableObject {
+        /// A feature layer visualizing GDP per capita.
+        let gdpPerCapitaLayer = FeatureLayer(
+            item: PortalItem(
+                portal: .arcGISOnline(connection: .anonymous),
+                id: .gdpPerCapita
+            )
+        )
+
```
dfeinzimer, line 87:
```suggestion
        
```

dfeinzimer:
> ## To Discuss
> * What should be a good generic docstring for the view model objects?

I propose something along the lines of:

"The model used to store the geo model and other expensive objects used in this view."

Which uses inspiration from doc added in Swift #3284 by @philium and the Apple doc referenced therein: [Managing model data in your app](https://developer.apple.com/documentation/swiftui/managing-model-data-in-your-app)

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
```diff
@@ -77,7 +92,7 @@ struct DownloadVectorTilesToLocalCacheView: View {
                             Button("Download Vector Tiles") {
                                 isDownloading = true
                             }
-                            .disabled(model.isDownloadDisabled || isDownloading)
+                            .disabled(model.exportVectorTilesTask == nil || isDownloading)
```
dfeinzimer, line 95: This is an edge case but I can reliably crash the app here if I tap the button before the preview tiles have loaded. Is there another condition that we can use to disable this button until the preview has fully loaded?

Shared/Samples/Show device location/ShowDeviceLocationView.swift
```diff
@@ -42,10 +65,10 @@ struct ShowDeviceLocationView: View {
                             }
                         }
                     }
-                    .disabled(model.areSettingsDisabled)
+                    .disabled(settingsButtonIsDisabled)
```
dfeinzimer, line 68: I'm not sure this is working as intended, or even needed. The button seems to dim itself automatically when the menu is presented.

yo1995: I think this was intended for failing to create the location data source. Say the user denied location permission for the app, or some other reasons that prevent the LDS from starting. Unfortunately Simulator has a [bug](https://developer.apple.com/forums/thread/693317) right now that doesn't allow us to test it.

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
yo1995: Are you using an iPad simulator to test? I made a small change in af8f130 , let me know if it solves the problem. vector tile download is super slow at my end… 🐌

dfeinzimer: Hmm, not sure. I think I may have been on a simulator. I just tried to reproduce with the latest updates on my physical iPad and crashed it here. 
Screenshot 2023-01-27 at 16 33 13

Shared/Samples/Show device location/ShowDeviceLocationView.swift
dfeinzimer: Ah okay, I see that it's performing as intended then on my physical iPad. Thanks!

dfeinzimer (commented):
Good to go from my perspective after [this open discussion](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/107#discussion_r1089568365) has some resolution.

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
yo1995: Thank you for reporting this. I made a commit a2b7028 as a quick patch. There were a few things wrong with this sample

- No safeguard for the export vector tiles task load status. Previously the task is directly assigned to the model's instance property before it is loaded, which allows the button to crash the sample
- Instead of using the `model.exportVectorTilesTask == nil` as the condition, a better one could be `model.exportVectorTilesTask.loadStatus == LoadStatus.loaded`. After all it doesn't make sense to allow using tapping the button before the task is loaded. I need to test if the task's load status is also published and can trigger the UI recompute though.
- If `vectorTileSourceInfo.allowsExportingTiles` is false, the download shouldn't be started. I'll break the `Model.downloadVectorTiles(extent:)` method into 2 part: condition check and async throw download. So no UI would be changed if the download task doesn't allow download in the first place.

Since they are not related to the view model change, I'll fix them in a separate PR. Issue logged at #114

dfeinzimer (dismissed):
The cash I reported last time appears to be resolved with the latest changes.

dfeinzimer (approved):

Shared/Samples/Show device location/ShowDeviceLocationView.swift
```diff
@@ -17,15 +17,38 @@ import CoreLocation
...
+    /// A Boolean value indicating whether the settings button is disabled.
+    @State private var settingsButtonIsDisabled = true
+    
     /// The view model for this sample.
     @StateObject private var model = Model()
     
     var body: some View {
         MapView(map: model.map)
             .locationDisplay(model.locationDisplay)
             .task {
```
philium, line 37:
```suggestion
            .task {
                guard model.locationDisplay.dataSource.status != .started else {
                    return
                }
```

Shared/Samples/Show result of spatial operations/ShowResultOfSpatialOperationsView.swift
```diff
@@ -46,71 +56,64 @@ struct ShowResultOfSpatialOperationsView: View {
...
-        case .difference:
-            resultGeometry = GeometryEngine.difference(.polygon1, .polygon2)
-        case .symmetricDifference:
-            resultGeometry = GeometryEngine.symmetricDifference(.polygon1, .polygon2)
-        case .intersection:
-            resultGeometry = GeometryEngine.intersection(.polygon1, .polygon2)
+        /// A graphic representing the result of the spatial operation.
+        var resultGraphic: Graphic {
+            return graphicsOverlay.graphics.last!
         }
```
philium, line 96:
```suggestion
        var resultGraphic: Graphic { graphicsOverlay.graphics.last! }
```

philium (approved):

philium (approved):

dfeinzimer (approved):

yo1995:
Thanks for the suggestions. I'll fix other issues in the upcoming PRs.

PR #118, opened 2023-02-10
[New] Show device location history

yo1995:
## Description

This PR implements `Show device location history` in `Maps` category.

URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Ting/ShowLocationHistory/Shared/Samples/Show%20device%20location%20history

## Linked Issue(s)

- `common-samples/issues/3820`

## How To Test

Run the sample and see if the track always works, and the objects get deallocated correctly when the sample quits.

## Screenshots

show-device-location-history

## To Discuss

1. Do we need to weakly capture self in the standalone tasks?
2. Make sure the `for await newLocation in locationDisplay.dataSource.locations` section is correct. This sample's key part is  to demonstrate async stream is the equivalent of `locationChangedHandler` pattern.


yo1995:
This is low priority, no hurry to review.

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
```diff
@@ -0,0 +1,234 @@
...
+        
+        init() {
+            locationDisplay = LocationDisplay(dataSource: Self.makeSimulatedLocationDataSource())
+            locationDisplay.autoPanMode = .recenter
+            
+            trackOverlay = GraphicsOverlay(graphics: [trackGraphic])
+            trackOverlay.renderer = SimpleRenderer(
+                symbol: SimpleLineSymbol(
+                    style: .solid,
+                    color: UIColor(Color.accentColor.opacity(0.5)),
```
dfeinzimer, line 113: This makes me think.. did we ever discuss swapping out `UIColor` with `Color` in the SDK? I imagine we did but stayed with `UIColor` for a reason?

dfeinzimer (approved):
Nice work!

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
yo1995: There was a discussion in `swift/issues/1491`. I think the concluded rule of thumb is to use `Color` when it's working with a SwiftUI `View` (like setting a color via view modifier), and use `UIColor` for other stuff that's displayed within the map/scene view.

dfeinzimer: Hmm, interesting. Thanks!

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
```diff
@@ -0,0 +1,234 @@
...
+        /// A Boolean value indicating whether to track the device location.
+        @Published var isTracking = false {
+            didSet {
+                if isTracking {
+                    startProcessingLocationChanges()
+                } else {
+                    stopProcessingLocationChanges()
+                }
+            }
+        }
```
philium, line 76: The model doesn't really use this property, so it doesn't make a lot of sense for it to be on the model. Instead, it could be a state property of the view itself:

```swift
@State private var isTracking = false
```

Then, instead of using a property observer to start/stop processing location changes, a task modifier could be used instead:

```swift
.task(id: isTracking) {
    guard isTracking else { return }
    await model.processingLocationChanges()
}

// In ShowDeviceLocationHistoryView.Model

func processingLocationChanges() async {
    for await location in locationDisplay.dataSource.locations {
        processLocationUpdate(lastPosition: location.position)
    }
}
```

That eliminates the need for `locationProcessingTask` and `stopProcessingLocationChanges()`. It also means that a toggle can be used:

```swift
Toggle(isTracking ? "Stop Tracking" : "Start Tracking", isOn: $isTracking)
    .toggleStyle(.button)
    .disabled(trackingButtonIsDisabled)
```

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
```diff
@@ -0,0 +1,234 @@
...
+            locationProcessingTask = nil
+        }
+        
+        /// Makes the simulated location data source for the sample.
+        /// - Returns: A simulated location data source in Los Angeles, CA.
+        private static func makeSimulatedLocationDataSource() -> SimulatedLocationDataSource {
+            let routePolyline = try? Polyline.fromJSON(polylineJSONString)
+            // Densify the simulated path to make it smoother.
+            let densifiedRoute = GeometryEngine.geodeticDensify(
+                routePolyline!,
```
philium, line 173:
```suggestion
            let routePolyline = try! Polyline.fromJSON(polylineJSONString)
            // Densify the simulated path to make it smoother.
            let densifiedRoute = GeometryEngine.geodeticDensify(
                routePolyline,
```

Just noticed that this generates a linting error, so this suggestion shouldn't be accepted. However, the suggested code makes more sense that conditionally trying something and just forcing it later, so perhaps we should reconsider having the `force_try` rule enabled.

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
philium:
```suggestion
                    color: .tintColor.withAlphaComponent(0.5),
```

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
```diff
@@ -0,0 +1,234 @@
...
+        private let locationGraphic = Graphic()
+        
+        /// The line graphic for traveled path.
+        private let trackGraphic = Graphic()
+        
+        /// A polyline builder to make the line graphic.
+        private let trackBuilder = PolylineBuilder(spatialReference: .webMercator)
+        
+        /// A multipoint builder to make the point graphics.
+        private let pointBuilder = MultipointBuilder(spatialReference: .webMercator)
```
philium, line 100: This name is confusing because there is a type called `PointBuilder`. A better name would be `multipointBuilder`.

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
yo1995: I don't want to leave `force_try` out - in most cases it is code smell. This is some of the rare cases that we actually need to `try!`, so instead I'll add a line to disable it. // swiftlint:disable:next force_try

Shared/Samples/Show device location history/ShowDeviceLocationHistoryView.swift
yo1995: Thank you. I like how the view's task modifier has the same lifetime as the view, so there is no need to manage a separate task by the user.

philium (approved):

yo1995:
Thanks!

PR #124, opened 2023-03-03
[Update] Use latest API signature

yo1995:
## Description

This PR fixes the build with latest API.

### DIFF
--- a/Shared/Samples/Add scene layer from service/AddSceneLayerFromServiceView.swift
+++ b/Shared/Samples/Add scene layer from service/AddSceneLayerFromServiceView.swift
@@ -33,7 +33,7 @@ private extension AddSceneLayerFromServiceView {
             let scene = Scene(basemapStyle: .arcGISTopographic)
             let point = Point(x: -4.4978, y: 48.3828, z: 62.0133, spatialReference: .wgs84)
             let camera = Camera(location: point, heading: 41.65, pitch: 71.2, roll: 0)
-            scene.initialViewpoint = Viewpoint(targetExtent: point, camera: camera)
+            scene.initialViewpoint = Viewpoint(boundingGeometry: point, camera: camera)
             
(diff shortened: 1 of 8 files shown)


Shared/Samples/Add scene layer from service/AddSceneLayerFromServiceView.swift
```diff
@@ -33,7 +33,7 @@ private extension AddSceneLayerFromServiceView {
             let scene = Scene(basemapStyle: .arcGISTopographic)
             let point = Point(x: -4.4978, y: 48.3828, z: 62.0133, spatialReference: .wgs84)
             let camera = Camera(location: point, heading: 41.65, pitch: 71.2, roll: 0)
-            scene.initialViewpoint = Viewpoint(targetExtent: point, camera: camera)
+            scene.initialViewpoint = Viewpoint(boundingGeometry: point, camera: camera)
```
yo1995, line 36: Although this works, it doesn't read so right. A point doesn't have a dimension, thus "bounding" doesn't make sense. "extent" had the same problem for both point an polyline, and we had a similar [question](https://community.esri.com/t5/arcgis-runtime-sdk-for-ios-questions/arcgis-error-quot-target-extent-has-zero-for-width/m-p/1218505#M7460) before. I'd like to hear some thoughts.

des12437 (approved):
Looks good.

Shared/Samples/Add scene layer from service/AddSceneLayerFromServiceView.swift
dfeinzimer: Though it does seem like an odd parameter name when used with a point, I think a point is technically considered a geometry with zero dimension. Might be worth discussing in an issue over on the Swift side.

dfeinzimer (approved):

PR #130, opened 2023-03-20
[New] Trace utility network

dfeinzimer:
## Description

This PR implements `Trace utility network`.
URL to README [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/df/AddTraceUtilityNetwork/Shared/Samples/Trace%20utility%20network/README.md)

## Linked Issue(s)

- `swift/issues/2259`
- `common-samples/pull/3835`

## How To Test

- Tap "Star a new trace"
- Select a trace type
- Tap various network elements to add them as starting points or barriers.
  - Note, to test the terminal picker look for elements with a triangle icon. You may need to zoom way in to get identify to hit them properly
- Tap "Trace"  

## Screenshots

trace-utility-network


Shared/Samples/Trace utility network/trace-utility-network.png
yo1995, line 1: Please grab the screenshot in light mode. Follow the requirements in the wiki.

zkline101 (commented):
Nice work. Here are some edits, I am still thinking through a few things but this is what I got thus far.

Shared/Samples/Trace utility network/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+![Image of trace utility network](trace-utility-network.png)
+
+## Use case
+
+You can use a trace to visualize and validate the network topology of a utility network for quality assurance. Subnetwork traces are used for validating whether subnetworks, such as circuits or zones, are defined or edited appropriately.
+
+## How to use the sample
+
+Tap "Start a new trace" to select the type of trace using the confirmation dialogue. Tap on one or more features while "Start" or "Barrier" is selected. When a junction feature is identified, you may be prompted to select a terminal. When an edge feature is identified, the distance from the tapped location to the beginning of the edge feature will be computed. Tap "Trace" to initiate a trace on the network. Tap "Reset" to clear the trace parameters and start over.
```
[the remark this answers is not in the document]
dfeinzimer: I changed the button to "Reset" to match the ObjC sample

Shared/Samples/Trace utility network/TraceUtilityNetworkView.swift
```diff
@@ -0,0 +1,539 @@
...
+        map.operationalLayers.forEach { ($0 as? FeatureLayer)?.clearSelection() }
+        points.removeAllGraphics()
+        pendingTrace = nil
+        tracingActivity = .none
+    }
+    
+    /// Performs important tasks including adding credentials, loading and adding operational layers.
+    func setup() async {
+        try? await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+        try? await network?.load()
```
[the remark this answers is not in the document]
dfeinzimer: Good call, done

Screenshot 2023-03-24 at 18 11 27

Shared/Samples/Trace utility network/TraceUtilityNetworkView.swift
```diff
@@ -0,0 +1,539 @@
...
+    /// the network.
+    var featureLayers: [URL] {
+        return [
+            URL.featureService.appendingPathComponent("0"),
+            URL.featureService.appendingPathComponent("3")
+        ]
+    }
+    
+    /// The utility tier for this sample.
+    var mediumVoltageRadial: UtilityTier? {
```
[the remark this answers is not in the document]
dfeinzimer: I thought making the extension `private` would have done that but it looks like that only makes them `fileprivate`.

.github/PULL_REQUEST_TEMPLATE.md
yo1995, line 1: Please delete this file from this PR. It should only be on `main` branch.

Shared/Samples/Trace utility network/README.md
```diff
@@ -0,0 +1,59 @@
...
+3. Add relevant `FeatureLayer`(s) to the map as operation layers, optionally applying custom styling.
+4. Allow users to choose a trace type from the set of `UtilityTraceParameters.TraceType`(s).
+5. Create `UtilityTraceParameters` with the selected trace type.
+6. Set the `UtilityTraceConfiguration` with the utility tier's `defaultTraceConfiguration` property.
+7. Use the `onSingleTapGesture` modifier to listen for tap events on the map view.
+8. Identify tapped features on the map and add a `Graphic` that represents its purpose (starting point or barrier) at the tapped location.
+9. Create a `UtilityElement` for the identified feature.
+10. Determine the type of the identified feature using `UtilityNetworkSource.sourceType`.
+11. If the type is `junction`, display a terminal picker when more than one terminal is found.
+12. If the type is `edge`, compute how far along the edge the user tapped using `enum GeometryEngine.polyline(_:fractionalLengthClosestTo:tolerance:)`.
```
yo1995, line 28: 🤔 I think I mentioned somewhere - when grabbing an API name use "qualified symbol name"
image

Shared/Samples/Trace utility network/README.md
```diff
@@ -0,0 +1,59 @@
...
+* UtilityTier
+* UtilityTraceConfiguration
+* UtilityTraceParameters
+* UtilityTraceResult
+* UtilityTraceType
+* UtilityTraversability
+
+## Additional information
+
+The [Naperville electrical](https://sampleserver7.arcgisonline.com/server/rest/services/UtilityNetwork/NapervilleElectric/FeatureServer) network feature service, hosted on ArcGIS Online, contains a utility network used to run the subnetwork-based trace shown in this sample.
```
yo1995, line 55: The name of the feature server is `NapervilleElectric`, but I think an electrical network sounds more correct…

Shared/Samples/Trace utility network/README.metadata.json
```diff
@@ -0,0 +1,58 @@
+{
+    "category": "",
```
yo1995, line 2:
```suggestion
    "category": "Utility network",
```

Shared/Samples/Trace utility network/TraceUtilityNetworkView.swift
```diff
@@ -0,0 +1,541 @@
...
+        }
+    }
+    
+    /// The view at the bottom of the screen that guides the user through the various stages of a
+    /// tracing activity.
+    var traceManager: some View {
+        HStack(spacing: 5) {
+            switch tracingActivity {
+            case .none:
+                Button("Start a new trace") {
```
yo1995, line 319:
```suggestion
                Button("Start a New Trace") {
```
We use title case in the sample viewer for most situations

- https://developer.apple.com/design/human-interface-guidelines/foundations/writing/
- https://developer.apple.com/library/archive/documentation/FinalCutProX/Conceptual/FxPlugHIG/TextStyleGuidelines/TextStyleGuidelines.html

yo1995 (commented):
I'm hoping UN logics can split into a view model, or at least somehow to keep the view file within 300 lines, instead of all in the view itself. But I'm on the fence 🤔 

On the one hand,

- Using a view model can draw a clearer line between GUI related code (updating graphics, user interaction handling), and geodatabase related code (configure trace parameter, run trace).
- Easier to follow for potential developers/audience.
- It doesn't need to be `TraceUtilityNetworkView.Model.swift` - can be sth like `TraceUtilityNetworkModel.swift` as a standalone type.

On the other hand,

- All in the view is easier to write, and all the types, extensions, and properties are readily available all in one place.
- Having a view model will [introduce another layer](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering), which inevitably add more glue code to the sample, which is already pretty long.

Can you give the reasons why you didn't split it? Maybe I'll be convinced if there is other benefits I haven't seen. Thanks.

Shared/Samples/Trace utility network/README.md
dfeinzimer: Since the electrical network has a name association, I believe this is correct. The old sample [uses this language](https://github.com/Esri/arcgis-runtime-samples-ios/blob/main/arcgis-ios-sdk-samples/Utility%20network/Trace%20utility%20network/README.md#:~:text=The%20Naperville%20electrical%20network%20feature%20service%2C%20hosted%20on%20ArcGIS%20Online%2C%20contains%20a%20utility%20network%20used%20to%20run%20the%20subnetwork%2Dbased%20trace%20shown%20in%20this%20sample.) as well.

dfeinzimer:
> Can you give the reasons why you didn't split it? Maybe I'll be convinced if there is other benefits I haven't seen. Thanks.

Mainly for simplicity, brevity and it's my understanding that view models are good for making SwiftUI code more testable. Since this code won't be programmatically tested I opted to write everything all in the view. I think it might also be a bit easier for someone to follow with everything in one file.

That being said, if you prefer a view model, I'd be happy to go that route.

dfeinzimer:
Merge conflicts resolved

dfeinzimer:
> I would say whatever layout the samples app is currently using (a view model or not), I would follow that.

I took a look at the existing samples and interestingly it's quite split, 21 with a model and 13 without. However it's quite clear that a model is used whenever a sample is longer and so this would be an outlier if it didn't so I'll go ahead and implement one. Reverting to draft.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,349 @@
...
+                    )
+                }
+                map.addOperationalLayer(layer)
+            }
+        }
+        
+        /// Runs a trace with the pending trace configuration and selects features in the map that
+        /// correspond to the element results.
+        ///
+        /// Note that elements are grouped by network source prior to selection so that all selections
```
zkline101, line 224:
```suggestion
        /// - Note: Elements are grouped by network source prior to selection so that all selections
```

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+        .disabled(model.pendingTraceParameters?.startingLocations.isEmpty ?? true)
+        .padding()
+        Button("Reset", role: .destructive) {
+            model.reset()
+        }
+        .padding()
+    }
+    
+    /// A binding to the type of starting point being set.
```
zkline101, line 40:
```suggestion
    /// A binding to the type of the starting point being set.
```

zkline101:
> For some reason, only the main and model files show in the picker.

I see them all on iPhone sim:
Screenshot 2023-03-29 at 10 42 00 AM

zkline101:
Also, when I start a new trace on iPad I get weird UI behavior, where the select trace type is cut off at the top.
Screenshot 2023-03-29 at 10 47 00 AM

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -239,30 +239,31 @@ extension TraceUtilityNetworkView {
...
-                switch tracingActivity {
-                case .none:
-                    hint = ""
-                case .settingPoints(let pointType):
-                    switch pointType {
-                    case .start:
-                        hint = "Tap on the map to add a Start Location."
-                    case .barrier:
-                        hint = "Tap on the map to add a Barrier."
+            DispatchQueue.main.async { [weak self] in
```
zkline101, line 247: We could probably mark this method as `@MainActor` instead of using dispatch queue.

dfeinzimer: I tried that out but landed on this because it introduced additional complexity elsewhere as they method became async.

zkline101: Hmm it looks like one spot it is getting called is already marked async and the other spot I think we could just wrap this call to this method in a task. @yo1995 What are your thoughts? I think it would be good to advertise us taking advantage of the new swift technology.

dfeinzimer: Ah, actually that works for me. I had tried `Task` with a priority and that still caused issues. Using `Task` without a priority works fine. I'll go that route.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -201,7 +205,9 @@ extension TraceUtilityNetworkView {
                 try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
                 try await network?.load()
             } catch {
-                updateUserHint(withMessage: "An error occurred while loading the network.")
+                Task {
+                    await updateUserHint(withMessage: "An error occurred while loading the network.")
+                }
                 return
```
zkline101: This method is already marked as async so you don't need a task here.

dfeinzimer: Oops, good catch, thank you

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,353 @@
...
+            if let message {
+                hint = message
+            } else {
+                switch tracingActivity {
+                case .none:
+                    hint = ""
+                case .settingPoints(let pointType):
+                    switch pointType {
+                    case .start:
+                        hint = "Tap on the map to add a Start Location."
```
zkline101, line 260:
```suggestion
                        hint = "Tap on the map to add a start location."
```

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,353 @@
...
+                    case .barrier:
+                        hint = "Tap on the map to add a Barrier."
```
zkline101, line 262:
```suggestion
                        hint = "Tap on the map to add a barrier."
```

zkline101 (commented):
Just found a few minor edits.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
dfeinzimer: I was matching the [capitalization of ObjC](https://github.com/Esri/arcgis-runtime-samples-ios/blob/b2694f9e5a97cd81ea042c1078708215ab4f3f48/arcgis-ios-sdk-samples/Utility%20network/Trace%20utility%20network/TraceUtilityNetworkViewController.swift#L351-L358) here but I do like the lowercase forms better. I'll take the suggestions and leave it for @yo1995 to comment on if he likes.

zkline101 (approved):

zkline101 (dismissed):

yo1995 (commented):
I've left some initial UI workflow thoughts. You can take a look at other sample viewers (such as [Qt](https://www.arcgis.com/home/search.html?t=content&q=tags%3A%22QmlSampleViewer%22)) to see their behaviors.

However, no need to make changes yet if you feel the current workflow is fine. I'd encourage you to present this sample at the Sample Design Council meeting on Wednesday morning next week and get some feedback, as this is quite a huge sample.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
```diff
@@ -0,0 +1,124 @@
...
+                        isPresented: $model.terminalSelectorIsOpen,
+                        actions: { terminalPickerButtons }
+                    )
+            case .tracing:
+                ProgressView()
+                    .progressViewStyle(CircularProgressViewStyle())
+                    .padding()
+            case .viewingResults:
+                Button("Reset", role: .destructive) {
+                    model.reset()
```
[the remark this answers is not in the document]
dfeinzimer: One of the main reasons I put trace type selection first in the workflow was because of the way [UtilityTraceParameter's initializer](https://next.sites.afd.arcgis.com/swift/api-reference/documentation/arcgis/utilitytraceparameters/init(tracetype:startinglocations:)) is.

By choosing the trace type up front we can initializer a trace parameters instance right away with the intended trace type and no additional properties are needed to hold intermediate values.

If we instead defer picking the trace type to later on in the flow, we have to defer the creation of trace parameters and instead hold the chosen starting points and barriers in two additional properties in the model.

In the name of keeping the sample as light as possible while using the API in a way that feels natural (creating parameters right await and then adding barriers and starting points to it) I went this route.

Thoughts on this? I can still change the workflow if you like, I thought it'd just be worth highlighting the downsides first.

Note that a parameters [starting points](https://next.sites.afd.arcgis.com/swift/api-reference/documentation/arcgis/utilitytraceparameters/addstartinglocation(_:)) and [barriers](https://next.sites.afd.arcgis.com/swift/api-reference/documentation/arcgis/utilitytraceparameters/addbarrier(_:)) are mutable where as [trace type](https://next.sites.afd.arcgis.com/swift/api-reference/documentation/arcgis/utilitytraceparameters/tracetype-swift.property) is not.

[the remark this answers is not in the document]
dfeinzimer: One option would be to keep the workflow as is and add buttons to clear the starting points and barriers.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.swift
```diff
@@ -0,0 +1,102 @@
...
+                    MapView(
+                        map: model.map,
+                        viewpoint: .initialViewpoint,
+                        graphicsOverlays: [model.points]
+                    )
+                    .onSingleTapGesture { screenPoint, mapPoint in
+                        model.lastSingleTap = (screenPoint, mapPoint)
+                    }
+                    .selectionColor(.yellow)
+                    .onChange(of: model.traceTypeSelectorIsOpen) { _ in
```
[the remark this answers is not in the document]
dfeinzimer: Not sure I follow here. What steps did you take to get back to trace type selection with graphics still leftover from a previous trace?

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
```diff
@@ -0,0 +1,124 @@
...
+import ArcGIS
+import SwiftUI
+
+extension TraceUtilityNetworkView {
+    /// The buttons and picker shown to the user while setting points.
+    @ViewBuilder
+    private var controlsForSettingPoints: some View {
+        Picker("Add starting points & barriers", selection: pointType) {
+            ForEach([PointType.start, PointType.barrier], id: \.self) { type in
+                Text(type.rawValue.capitalized).tag(type)
```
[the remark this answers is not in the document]
dfeinzimer: Changed to `Text(String(describing: type).capitalized).tag(type)`. Should that work okay?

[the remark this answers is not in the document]
dfeinzimer: Update: I added `PointType.label`. I think that's more what you were suggesting. This is still ready for review.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,353 @@
...
+        ///   - proxy: The map view proxy to perform the identify operation with.
+        /// - Returns: The first discoverable feature or `nil` if none were identified.
+        func identifyFeatureAt(_ screenPoint: CGPoint, with proxy: MapViewProxy) async -> ArcGISFeature? {
+            guard let feature = try? await proxy.identifyLayers(
+                screenPoint: screenPoint,
+                tolerance: 10
+            ).first?.geoElements.first as? ArcGISFeature else {
+                return nil
+            }
+            return feature
```
zkline101, line 181:
```suggestion
            try? await proxy.identifyLayers(
                screenPoint: screenPoint,
                tolerance: 10
            ).first?.geoElements.first as? ArcGISFeature
```

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,353 @@
...
+        }
+        
+        /// Makes new utility trace parameters with the provided trace type.
+        /// - Parameter type: The trace type.
+        func makeTraceParameters(with type: UtilityTraceParameters.TraceType) {
```
zkline101, line 186: This method name is pretty confusing because it starts with "make" like it is returning some trace parameters so maybe it would be good to rename or just return the `UtilityTraceParameters`.

dfeinzimer: Since this method is called from the UI and not the model itself, I found that it didn't really make sense to return the params so I've instead opted for a rename to `initializeTraceParameters`, keeping the existing behavior and adjusting the doc to match.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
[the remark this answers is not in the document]
yo1995: common-samples/pull/3835

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
dfeinzimer: Hmm ok, I'm weary of taking transferring too much responsibility from the model to a button in the view (this would also involve opening up access to the model's `mediumVoltageRadial` property).

I'd like to strike the right balance here. Can you think of any names or other approaches that might make sense?

dfeinzimer: Would something like `prepareForTrace(ofType:)` work for you? 

Then we could also move `model.tracingActivity = .settingPoints(pointType: .start)` from the button into that method.

zkline101: What about `setTraceParameters(of:)` ?

dfeinzimer: I'll take it!

zkline101 (dismissed):

yo1995 (commented):
A few UI feedback provided in Slack

dfeinzimer:
I've made some UI improvements to closer match the old sample. This is ready for re-review.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,357 @@
...
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        func setup() async {
+            do {
+                try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+                try await network?.load()
+            } catch {
+                await updateUserHint(withMessage: "An error occurred while loading the network.")
+                return
+            }
+            featureLayerURLs.forEach { url in
```
yo1995, line 213: Here is an idea you may take or ignore. Instead of adding the 2 feature layers to the map again, you can retrieve them from the portal item itself, such as 

```swift
private var featureLayers: [FeatureLayer] {
    map.operationalLayers
        .compactMap {
            ($0 as? GroupLayer)?.layers.compactMap { $0 as? FeatureLayer }
        }
        .flatMap { $0 }
}
```

Benefits:

- No need to create additional replicas of these 2 layers
- All the feature layers come in the order of how they are authored in the portal item
- Can get rid of a few variables such as `featureService` and `featureLayerURLs`

Cons:

- This group layer structure might not apply to users' own portal items
- Takes time to loop through layers. If an item contains many layers, it is slow

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
yo1995: I like the current approach.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
yo1995: No preference 🙈

Shared/Samples/Trace utility network/TraceUtilityNetworkView.swift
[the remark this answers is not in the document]
yo1995: Resolved in Slack.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -0,0 +1,357 @@
...
+        /// Runs a trace with the pending trace configuration and selects features in the map that
+        /// correspond to the element results.
+        ///
+        /// - Note: Elements are grouped by network source prior to selection so that all selections
+        /// per operational layer can be made at once.
+        func trace() async throws {
+            guard let pendingTraceParameters = pendingTraceParameters else { return }
+            let traceResults = try await network?.trace(using: pendingTraceParameters)
+                .filter { $0 is UtilityElementTraceResult }
+            for result in traceResults as? [UtilityElementTraceResult] ?? [] {
```
yo1995, line 236: I cannot think of a way to write it more concisely, so I blame Swift not being expressive enough! 😂

yo1995 (commented):
Thanks for writing this sample. Overall the structure is very clear and easy to follow. I have some thoughts and suggestions below.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
```diff
@@ -0,0 +1,123 @@
...
+    /// The buttons and picker shown to the user while setting points.
+    @ViewBuilder
+    private var controlsForSettingPoints: some View {
+        Group {
+            Picker("Add starting points & barriers", selection: pointType) {
+                ForEach([PointType.start, PointType.barrier], id: \.self) { type in
+                    Text(type.label).tag(type)
+                }
+            }
+            .padding()
```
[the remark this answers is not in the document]
dfeinzimer: Static values seem odd, and I played around with some geometry proxy solutions but the static values I choose performed best across all the scenarios I tested against (full screen catalyst, iPad, 14 Pro Max, SE 3rd Gen.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
```diff
@@ -0,0 +1,123 @@
...
+            case .none:
+                Button("Start a New Trace") {
+                    withAnimation {
+                        model.traceTypeSelectorIsOpen.toggle()
+                    }
+                }
+                .padding()
+                .confirmationDialog("Trace Type", isPresented: $model.traceTypeSelectorIsOpen) {
+                    traceTypePickerButtons
+                }
```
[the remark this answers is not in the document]
dfeinzimer: I'll take it! Lets use remove a stateful property and some other related code.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
```diff
@@ -0,0 +1,123 @@
...
+            Button(terminal.name) {
+                model.lastAddedElement?.terminal = terminal
+                model.updateUserHint(withMessage: "terminal: \(terminal.name)")
+            }
+        }
+    }
+    
+    /// The menu at the bottom of the screen that guides the user through running a trace.
+    var traceMenu: some View {
+        HStack(spacing: 5) {
```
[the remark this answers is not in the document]
dfeinzimer: Something is odd about this, removing the `HStack` causes no UI to be drawn at all, even with `controlsForSettingPoints` being an `HStack` itself now. It might be something with `controlsForSettingPoints` needing to anchor itself to another view inside of the toolbar. I can revisit and spend more time on this if you like.

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
dfeinzimer: Not sure I fully understand the suggestion here, can you clarify a bit? 

> Instead of adding the 2 feature layers to the map again

I think these need to be added no matter what?

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
dfeinzimer: Yes, this is cleaner, thank you!

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
yo1995: The portal item https://www.arcgis.com/home/item.html?id=471eb0bf37074b1fbb972b1da70fb310 contains a group layer, which bundles all the feature layers.

In theory, we can directly use the feature layers in that group layer because they are already part of the map. But that involves filtering them from the `operationalLayers` and create separate references. Which is why we added them as duplicates using the service feature table (with layer ID) URLs, just to display them and makes it easy to get to the 2 specific layers.

What I suggest is to retrieve the 2 feature layers from the group layer, so no duplicate layers that use the same feature table in the map. But there are some drawbacks as I listed, so if you think the current approach is good, I don't think it needs to be changed.

dfeinzimer: I see. I think I prefer the way it's currently written as I find it a bit more digestible and [it's similar to the Objective C approach](https://github.com/Esri/arcgis-runtime-samples-ios/blob/b2694f9e5a97cd81ea042c1078708215ab4f3f48/arcgis-ios-sdk-samples/Utility%20network/Trace%20utility%20network/TraceUtilityNetworkViewController.swift#L72-L102).

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Views.swift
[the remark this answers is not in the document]
yo1995: This might be yet another SwiftUI bug (or feature). On iOS 15, `Group` works just fine.

I actually thought making both `Group` (i.e., `controlsForSettingPoints` and `traceMenu`) makes more sense. Because the `traceMenu` is in a `ToolbarItemGroup`, the layout should already be managed by the toolbar, thus we don't need to use an `HStack`.

yo1995 (dismissed):
LGTM! The comment on using `Group` is up to you.

yo1995 (approved):

yo1995 (approved):

zkline101 (approved):
Tested and the code looks good, nice job!

dfeinzimer:
Thank you both!

Shared/Samples/Trace utility network/TraceUtilityNetworkView.swift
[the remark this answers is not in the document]
yo1995: It's rare that all of us missed this while reviewing this PR, but there is a case that the network itself is `nil` because the map fails to load, thus it doesn't throw an error while being in an invalid state.

It led to the issue in `swift/issues/4401`

PR #133, opened 2023-03-22
[New] Navigate route

dfeinzimer:
## Description

This PR implements `Navigate route`. It was originally written by Christopher but not fully completed/merged.
[README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/df/New-NavigateARoute/Shared/Samples/Navigate%20a%20route) 

## Linked Issue(s)

- `swift/issues/2254`

## How To Test

- 1: Tap "Navigate"
- 2: Optionally pan the map away and press "Recenter" to recenter on the current location
- 3: Optionally press "Reset" to stop navigation

## Screenshots

navigate-route


mhdostal:
Folder/Group needs to be put in the sorted order.
image

mhdostal:
The app crashes when tapping the "info" button in the "Navigate a route" sample; the `NavigateARouteView.swift` file needs to be added to the "Copy Source Code Files" section of the project.

Shared/Samples/Navigate a route/NavigateARouteView.swift
```diff
@@ -0,0 +1,349 @@
...
+        @Published var isNavigateDisabled = true
+        
+        /// A Boolean value indicating whether the sample is navigating the route.
+        @Published var isNavigatingRoute = false
+        
+        /// A Boolean value indicating whether navigation is being reset.
+        @Published var isResettingRoute = false
+        
+        /// A Boolean value indicating whether to show an alert.
+        @Published var isShowingAlert = false
```
mhdostal, line 91: This is never used by the view. Is it still coming?

mhdostal (changes requested):
Just the one question and the conversations about the crash and folder order.

mhdostal (approved):

Samples.xcodeproj/project.pbxproj
```diff
@@ -57,6 +57,7 @@
 		00E5401C27F3CCA200CF66D5 /* SamplesApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00E5400C27F3CCA100CF66D5 /* SamplesApp.swift */; };
 		00E5401E27F3CCA200CF66D5 /* ContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00E5400D27F3CCA100CF66D5 /* ContentView.swift */; };
 		00E5402027F3CCA200CF66D5 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 00E5400E27F3CCA200CF66D5 /* Assets.xcassets */; };
+		75E92C6129CCBCF900111226 /* NavigateARouteView.swift in Copy Source Code Files */ = {isa = PBXBuildFile; fileRef = E008221428760157002AD138 /* NavigateARouteView.swift */; };
```
yo1995, line 60: Sorry for the late notice. Can you help rename the sample to "Navigate route" according to the common design? There was an effort to standardize the names across platforms, and the old iOS sample name deviates from the common design. Thanks.

- the folder name
- README title
- title in metadata
- swift filename
- view type name (NavigateARouteView -> NavigateRouteView)

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+                routeTracker: routeTracker,
+                locationDataSource: mockDataSource
+            )
+            
+            // Sets the location display's data source.
+            locationDisplay.dataSource = routeTrackerLocationDataSource
+            
+            // Updates the graphics and viewpoint.
+            routeRemainingGraphic.geometry = routeGeometry
+            viewpoint = Viewpoint(boundingGeometry: routeGeometry)
```
yo1995, line 244:
```suggestion
            viewpoint = Viewpoint(boundingGeometry: routeGeometry, rotation: 0)
```
Like to see the map rotate back to the initial state.

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+        }
+        
+        /// Start monitoring multiple asynchronous streams of information.
```
yo1995, line 247:
```suggestion
        /// Starts monitoring multiple asynchronous streams of information.
```

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+                group.addTask { await self.trackVoiceGuidance() }
+            }
+        }
+        
+        /// Monitor the asynchronous stream of tracking statuses.
+        ///
+        /// When new statuses are delivered, update the route's traversed and remaining graphics.
+        private func trackStatus() async {
+            guard let routeTracker else { return }
+            for try await status in routeTracker.$trackingStatus {
```
yo1995, line 261: I attempted to remove the `try` keyword and the code still works. 🤔  Out of curiosity, can you find out why/when the tracking status can be `nil`?

In iOS API, the delegate method `func routeTracker(_ routeTracker: AGSRouteTracker, didUpdate trackingStatus: AGSTrackingStatus)` doesn't give a `nil` status.

mhdostal (dismissed):

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+                if let status {
+                    routeTraversedGraphic.geometry = status.routeProgress.traversedGeometry
+                    routeRemainingGraphic.geometry = status.routeProgress.remainingGeometry
+                    
+                    switch status.destinationStatus {
+                    case .notReached, .approaching:
+                        statusText = """
+                        Distance remaining: \(status.routeProgress.remainingDistance.distanceRemainingText)
+                        Time remaining: \(timeFormatter.string(from: status.routeProgress.remainingTime) ?? "")
```
yo1995, line 270:
```suggestion
                        Time remaining: \(timeFormatter.string(from: status.routeProgress.remainingTime)!)
```
From the doc I think it is safe to force unwrap.

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+                        if status.currentManeuverIndex + 1 < directions.count {
+                            statusText.append("\nNext direction: \(directions[status.currentManeuverIndex + 1].text)")
+                        }
+                    case .reached:
+                        if status.remainingDestinationCount > 1 {
+                            statusText = "Intermediate stop reached, continue to next stop."
+                            try? await routeTracker.switchToNextDestination()
+                        } else {
+                            isNavigatingRoute = false
+                            isResettingRoute = true
```
yo1995, line 281: I prefer not to reset the navigation to the start when it ends. Rather, I'd leave it to the user to tap on the reset button.

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+        /// Start monitoring multiple asynchronous streams of information.
+        private func startTracking() async {
+            await withTaskGroup(of: Void.self) { group in
+                group.addTask { await self.trackAutoPanMode() }
+                group.addTask { await self.trackStatus() }
+                group.addTask { await self.trackVoiceGuidance() }
+            }
+        }
+        
+        /// Monitor the asynchronous stream of tracking statuses.
```
yo1995, line 256:
```suggestion
        /// Monitors the asynchronous stream of tracking statuses.
```

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+                    await model.startNavigation()
+                }
+                .disabled(model.isNavigateDisabled || model.isNavigatingRoute)
+                
+                Spacer()
+                
+                Button("Recenter") {
+                    model.autoPanMode = .navigation
+                }
+                .disabled(!model.isNavigatingRoute || model.autoPanMode == .navigation)
```
yo1995, line 62: Seems to me that `model.autoPanMode` only changes when the navigation is ongoing.
```swift
.disabled(model.autoPanMode == .navigation)
```
If you don't set the `model.autoPanMode` to `.off` in the initializer, but to `.navigation` either in the initializer or directly on the property's declaration, then I think this condition can be simplified.

It is OK if you don't think it can be removed.

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+                try await locationDisplay.dataSource.start()
+                locationDisplay.autoPanMode = .navigation
+                await startTracking()
+            } catch {
+                self.error = error
+            }
+        }
+        
+        func resetNavigation() async {
+            locationDisplay.autoPanMode = .off
```
yo1995, line 323: If the suggestion above is applied, then this line can be removed.

yo1995 (commented):
First pass comments. I'll take a closer look at the booleans once some other suggestions are addressed.

Shared/Samples/Navigate route/NavigateRouteView.swift
dfeinzimer: `Model.autoPanMode` can be initialized to `.navigation` but the button will still be enabled then when there's no location to snap to. I think the condition should remain.

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,354 @@
...
+        }
+        
+        /// Sets the route tracker and location display's data source with a solved route. Updates
+        /// the list of directions, the route ahead graphic, and the map's viewpoint.
+        func setNavigation() {
+            guard let routeResult else { return }
+            
+            // Creates a route tracker from the route results.
+            routeTracker = RouteTracker(routeResult: routeResult, routeIndex: 0, skipsCoincidentStops: true)
+            guard let routeTracker else { return }
```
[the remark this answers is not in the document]
dfeinzimer: swift/issues/3842

dfeinzimer:
Conflicts resolved, feedback applied - this is ready for review again.

Shared/Samples/Navigate route/NavigateRouteView.swift
yo1995: Hmm interesting. I see Kotlin generates the same code (navigation/RouteTracker.kt#L124), while in cocoa it only passes non optional status. If the status is `nil` only for the first empty location, then I think it's worth logging an issue to improve this API.

mhdostal (dismissed):

Shared/Samples/Navigate route/NavigateRouteView.swift
yo1995: After discussion, it is an expected behavior.

yo1995:
Initially I thought `isNavigateDisabled` and `isNavigatingRoute` can be combined together as 1 bool to control the various states of the buttons. But I found that combined bool will take too much responsibility and is vague in meaning, so I like what you have here.

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,352 @@
...
+                Spacer()
+                
+                Button("Recenter") {
+                    model.autoPanMode = .navigation
+                }
+                .disabled(!model.isNavigatingRoute || model.autoPanMode == .navigation)
+                
+                Spacer()
+                
+                Button("Reset") {
```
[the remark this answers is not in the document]
dfeinzimer: I think it may have something to do with the viewpoint being set in WGS84 from the route geometry and it coming back out at as web mercator in `onViewpointChanged` but I'd have to debug further. Is it a blocker for this sample?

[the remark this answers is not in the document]
dfeinzimer: I dug some more; on line 248 the viewpoint is created without a scale (`viewpoint = Viewpoint(boundingGeometry: routeGeometry, rotation: 0)`. If a scale is passed, (e.g. `viewpoint = Viewpoint(center: routeGeometry.extent.center, scale: 34182, rotation: 0)`) then the provided scale is preserved when `onViewpointChanged` fires.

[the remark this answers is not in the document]
yo1995: It doesn't block this sample. I think it might be worth logging an issue to take a further look though.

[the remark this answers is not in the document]
dfeinzimer: `swift/issues/3854`

yo1995 (commented):
Will approve after the merge conflict is solved.

dfeinzimer:
Feedback applied, conflicts resolved

mhdostal (approved):

Shared/Samples/Navigate route/NavigateRouteView.swift
```diff
@@ -0,0 +1,351 @@
...
+        private let speechSynthesizer = AVSpeechSynthesizer()
+        
+        /// The graphics overlay for the stops.
+        private let stopGraphicsOverlay: GraphicsOverlay
+        
+        /// The graphics overlay for the route.
+        private let routeGraphicsOverlay: GraphicsOverlay
+        
+        /// Formats the time remaining for display.
+        private let timeFormatter: DateComponentsFormatter = {
```
yo1995, line 138: This formatter is still needed for iOS 15: https://stackoverflow.com/a/72945284/14369688

PR #141, opened 2023-03-23
[Update] 200.1.0 ArcGIS Bundle ID

zkline101:
## Description

This PR updates the ArcGIS Bundle ID which prevents a crash when clicking on the about button.


yo1995 (approved):

saratchandrakarumuri (commented):
App is still crashing for me when I open the info pane

2023-03-23 at 5 47 PM

yo1995:
Sorry Sarat I should mention, this need to be tested with the local toolkit folder and ArcGIS daily build binary, because…

- 200.0.0-beta doesn't have the same bundle ID
- The source file uses a different bundle ID ("ArcGIS" instead of "com.esri.ArcGIS", for some reason). Only the binary is using the "com.esri.ArcGIS" bundle ID.

saratchandrakarumuri:
@yo1995 Ahhh... Makes sense... I tried with the source code swift package... I will try with daily build now

saratchandrakarumuri (approved):
Changes looking good

zkline101:
Thank you both

PR #154, opened 2023-04-04
[Update] Adjust `Compass` API signature

yo1995:
## Description

This PR updates `Set viewpoint rotation`.

The autohide setting is now a view modifier.

## Linked Issue(s)

- https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/291

## How To Test

The sample behaves exactly the same as before.

## To Discuss

To match the old code, should we explicitly write `.autoHideDisabled(true)`? I don't think so because…
1. its default value is true (which means the compass doesn't autohide)
2. there are other view modifiers on the compass. A user can explore them with other methods, while the sample can keep simple.

### DIFF
--- a/Shared/Samples/Set viewpoint rotation/SetViewpointRotationView.swift
+++ b/Shared/Samples/Set viewpoint rotation/SetViewpointRotationView.swift
@@ -39,9 +39,9 @@ struct SetViewpointRotationView: View {
                 }
                 .overlay(alignment: .topTrailing) {
                     Compass(
-                        viewpointRotation: $rotation,
-                        autoHide: false
+                        viewpointRotation: $rotation
                     )
+                    .autoHideDisabled(true)
                     .padding()
                 }
             
--- a/Shared/Samples/Set viewpoint rotation/SetViewpointRotationView.swift
+++ b/Shared/Samples/Set viewpoint rotation/SetViewpointRotationView.swift
@@ -38,11 +38,9 @@ struct SetViewpointRotationView: View {
                     rotation = viewpoint.rotation
                 }
                 .overlay(alignment: .topTrailing) {
-                    Compass(
-                        viewpointRotation: $rotation
-                    )
-                    .autoHideDisabled()
-                    .padding()
+                    Compass(viewpointRotation: $rotation)
+                        .autoHideDisabled()
+                        .padding()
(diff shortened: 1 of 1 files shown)


Shared/Samples/Set viewpoint rotation/SetViewpointRotationView.swift
```diff
@@ -39,8 +39,7 @@ struct SetViewpointRotationView: View {
                 }
                 .overlay(alignment: .topTrailing) {
                     Compass(
-                        viewpointRotation: $rotation,
-                        autoHide: false
+                        viewpointRotation: $rotation
                     )
                     .padding()
```
mhdostal, line 44: The old `autoHide` argument is now a modifier, but the meaning has flipped, it's now whether you want to disable the auto-hide behavior, which for this case would be true (in order to match previous behavior).
```
                    Compass(viewpointRotation: $rotation)
                        .autoHideDisabled(true)
                        .padding()
```

mhdostal (changes requested):

dfeinzimer (commented):
Yes, just the one suggestion from Mark and this should be good to go. Note `true` can be omitted thanks to the presence of a default value.

Shared/Samples/Set viewpoint rotation/SetViewpointRotationView.swift
```diff
@@ -39,9 +39,9 @@ struct SetViewpointRotationView: View {
                 }
                 .overlay(alignment: .topTrailing) {
                     Compass(
-                        viewpointRotation: $rotation,
-                        autoHide: false
+                        viewpointRotation: $rotation
                     )
+                    .autoHideDisabled()
                     .padding()
```
dfeinzimer, line 45: We should probably single line the init since it only has 1 param now

```swift
                    Compass(viewpointRotation: $rotation)
                        .autoHideDisabled()
                        .padding()
```

dfeinzimer (changes requested):
We should probably single line the init since it only has 1 param now

```swift
Compass(viewpointRotation: $rotation)
    .autoHideDisabled()
    .padding()
```

dfeinzimer (approved):

mhdostal (approved):

PR #160, opened 2023-04-06
[Fix] Deletion precondition in Sketch on map

### LINKED ISSUE #158: [Sketch on map] Crash when deleting mid vertex
1. Create a new line or area based sketch
2. Attempt to delete a mid vertex

> Fatal error: Precondition failed. Illegal state. A mid vertex element can't be deleted.

Should we add more protection against a user triggering this condition?

dfeinzimer:
Closes #158 

- Disables "Delete selected element" when the selected element cannot be deleted.

### DIFF
--- a/Shared/Samples/Sketch on map/SketchOnMapView.swift
+++ b/Shared/Samples/Sketch on map/SketchOnMapView.swift
@@ -144,7 +144,7 @@ extension GeometryEditorMenu {
             } label: {
                 Label("Delete Selected Element", systemImage: "xmark.square.fill")
             }
-            .disabled(model.selection == nil || !(model.selection?.canBeDeleted ?? false))
+            .disabled(deleteButtonIsDisabled)
             
             Button(role: .destructive) {
                 model.geometryEditor.clearGeometry()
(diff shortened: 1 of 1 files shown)


Shared/Samples/Sketch on map/SketchOnMapView.swift
```diff
@@ -144,7 +144,7 @@ extension GeometryEditorMenu {
             } label: {
                 Label("Delete Selected Element", systemImage: "xmark.square.fill")
             }
-            .disabled(model.selection == nil)
+            .disabled(model.selection == nil || !(model.selection?.canBeDeleted ?? false))
```
dfeinzimer, line 147: @rolson, this precondition isn't documented on `deleteSelectedElement()`, should that be added?

philium: FYI: Instead of (essentially) checking for `nil` twice, it would be better to move the logic into a computed property:

```swift
var deleteButtonDisabled: Bool {
    guard let selection = model.selection else { return true }
    return !selection.canBeDeleted
}
```

yo1995 (approved):
Thanks for catching this issue! 

🤜🐛🤛

Shared/Samples/Sketch on map/SketchOnMapView.swift
rolson: > @rolson, this precondition isn't documented on deleteSelectedElement(), should that be added?

Sure

rolson (approved):

Shared/Samples/Sketch on map/SketchOnMapView.swift
rolson: > @rolson, this precondition isn't documented on deleteSelectedElement(), should that be added?

Actually does it fail silently or crash? If crash, then yes it needs to be documented in GeometryEditor.

dfeinzimer: @rolson It was crashing:

capture

PR #161, opened 2023-04-06
[Fix] Compile error related to `Item.id`

zkline101:
## Description

ting: See Swift 3906 - it involves a bug in Item.ID's optionality, and the final solution is hard to explain in a few words.

### DIFF
--- a/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
+++ b/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
@@ -170,7 +170,7 @@ class OfflineMapModel: ObservableObject, Identifiable {
     let offlineMapTask: OfflineMapTask
     
     /// The directory where the mmpk will be stored.
-    let mmpkDirectory: URL
+    let mmpkDirectory: URL?
     
     /// The currently running download job.
     @Published private(set) var job: DownloadPreplannedOfflineMapJob?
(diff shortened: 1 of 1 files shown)


Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
```diff
@@ -181,9 +181,17 @@ class OfflineMapModel: ObservableObject, Identifiable {
...
-            .appendingPathComponent(preplannedMapArea.portalItem.id.rawValue)
-            .appendingPathExtension("mmpk")
+        
+        let mmpkDirectory: URL?
+        if let itemID = preplannedMapArea.portalItem.id {
+            mmpkDirectory = temporaryDirectory
+                .appendingPathComponent(itemID.rawValue)
+                .appendingPathExtension("mmpk")
+        } else {
+            mmpkDirectory = nil
```
yo1995, line 191: Does this means if the preplanned map area doesn't have a portal ID, then the map doesn't support downloading mmpk at all?

If so, I think we can use a precondition instead of making the directory optional, to alert user this fact.

dfeinzimer: Another idea would be to make this init failable and use `compactMap` instead of `map` at the call site.

dfeinzimer (commented):
Just the one open discussion from Ting

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
zkline101: > Does this means if the preplanned map area doesn't have a portal ID, then the map doesn't support downloading mmpk at all?

I think it just means the preplanned map area is not hosted on AGOL.



> Another idea would be to make this init failable and use compactMap instead of map at the call site.

I like that idea. Because if there is no item ID then we should just not create the model.

yo1995 (approved):

dfeinzimer (approved):

yo1995:
Should be ready to merge. Will create the dummy app after this.

Issue #110, opened 2023-01-09
"Set basemap" sample - Double border


dfeinzimer:
In the "Set basemap" sample, [a second border is set](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/5992329590720aa0da6e7caf68770949a001c808/Shared/Samples/Set%20basemap/SetBasemapView.swift#L38) on the basemap gallery. This leads to a bit of odd looking UI.

![Simulator Screen Shot - iPhone 14 Pro Max - 2023-01-09 at 14 01 07](https://user-images.githubusercontent.com/16397058/211418341-a37a98f0-070c-454d-ad2d-6bf00cc8ce96.png)
![Simulator Screen Shot - iPhone 14 Pro Max - 2023-01-09 at 14 01 38](https://user-images.githubusercontent.com/16397058/211418345-69ab462d-ba58-41de-b4fd-5f22d5aeb1d7.png)

PR #163, opened 2023-04-10
[Update] Basemap gallery API usage

dfeinzimer:
Related Swift #3931

- Updates basemap gallery example for changes made in [toolkit #283](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/283).


Shared/Samples/Set basemap/SetBasemapView.swift
```diff
@@ -32,19 +32,24 @@ struct SetBasemapView: View {
...
-                        .style(.automatic())
+        GeometryReader { geometryProxy in
+            MapView(map: map)
+                .overlay(alignment: .topTrailing) {
+                    if isShowingBasemapGallery {
+                        BasemapGallery(geoModel: map)
+                            .style(.automatic())
+                            .frame(maxWidth: geometryProxy.size.width / 3)
+                            .esriBorder()
+                            .padding()
```
yo1995, line 43: Is this padding for setting the gallery view away from the screen edge?

Shared/Samples/Set basemap/SetBasemapView.swift
```diff
@@ -32,19 +32,24 @@ struct SetBasemapView: View {
...
-                    BasemapGallery(geoModel: map)
-                        .style(.automatic())
+        GeometryReader { geometryProxy in
+            MapView(map: map)
+                .overlay(alignment: .topTrailing) {
+                    if isShowingBasemapGallery {
+                        BasemapGallery(geoModel: map)
+                            .style(.automatic())
+                            .frame(maxWidth: geometryProxy.size.width / 3)
+                            .esriBorder()
```
yo1995, line 42: In #110 this was removed. Can you add a comment here on why it is used in the new version? So we can refer to here when we review the code in the future.

yo1995 (approved):

Shared/Samples/Set basemap/SetBasemapView.swift
dfeinzimer: Yes

Shared/Samples/Set basemap/SetBasemapView.swift
dfeinzimer: In 110, it was removed because the component was adding its own border. That's no longer true, its up to a user of the component to supply their own border now.

mhdostal (approved):

PR #170, opened 2023-04-13
[Fix] Accessory button and help text is not clearly visible with selected sample

des12437:
## Description

This PR fixes the issue described [here](https://devtopia.esri.com/runtime/swift/issues/3878) where in iOS 16+ a selected list item's help text color does not contrast well with the purple background color, and the button is no longer visible since it's tint color is purple.

Now the help text and button are white when the list item is selected and the background color is purple. If the list item's background is not purple the color of the help text and button is not modified.

## Linked Issue(s)

- `swift/issues/3878`

## How To Test
Test with iOS 16+

## Screenshots

Before

After (dark mode)

After (light mode)

Question - Is the help text contrast in dark mode still too light? The accent color for dark mode may need to be changed to one with less brightness. Perhaps a lighter shade of the accessory color for light mode.

### DIFF
--- a/Shared/Supporting Files/Views/SampleList.swift
+++ b/Shared/Supporting Files/Views/SampleList.swift
@@ -90,7 +90,6 @@ private extension SampleList {
                                 Text(sample.description)
                                     .foregroundColor(selection == sample.name ? .white.opacity(0.8) : .secondary)
                                     .font(.caption)
-                                    .foregroundColor(.secondary)
                                     .transition(.move(edge: .top).combined(with: .opacity))
                             } else {
                                 Text(sample.description)
--- a/Shared/Supporting Files/Views/SampleList.swift
+++ b/Shared/Supporting Files/Views/SampleList.swift
@@ -19,6 +19,12 @@ struct SampleList: View {
     /// A Boolean value that indicates whether the user is searching.
     @Environment(\.isSearching) private var isSearching
     
+    /// The horizontal size class of the enviroment.
+    @Environment(\.horizontalSizeClass) var horizontalSizeClass
+    
+    /// The vertical size class of the enviroment.
+    @Environment(\.verticalSizeClass) var verticalSizeClass
+    
     /// All samples that will be displayed in the list.
     let samples: [Sample]
     
@@ -46,7 +52,7 @@ struct SampleList: View {
     
     var body: some View {
         List(displayedSamples, id: \.name, selection: $selection) { sample in
-            SampleRow(sample: sample, selection: selection)
+            SampleRow(sample: sample, selection: selection, horizontalSizeClass: horizontalSizeClass, verticalSizeClass: verticalSizeClass)
         }
         .navigationTitle("Samples")
         .toolbar {
@@ -75,6 +81,12 @@ private extension SampleList {
         /// The current device type.
         let deviceType = UIDevice.current.userInterfaceIdiom
         
+        /// The horizontal size class of the device.
+        let horizontalSizeClass: UserInterfaceSizeClass?
+        
+        /// The vertical size class of the device.
+        let verticalSizeClass: UserInterfaceSizeClass?
+        
         /// A Boolean value indicating whether to show the sample's description
         @State private var isShowingDescription = false
         
@@ -88,6 +100,8 @@ private extension SampleList {
         
         var selectionColorIsAccentColor: Bool {
             if #available(iOS 16, *), deviceType != .phone {
+                return true
+            } else if #available(iOS 16, *), horizontalSizeClass == .compact, verticalSizeClass == .compact {
                 return true
             } else {
                 return false
--- a/Shared/Supporting Files/Views/SampleList.swift
+++ b/Shared/Supporting Files/Views/SampleList.swift
@@ -19,12 +19,6 @@ struct SampleList: View {
     /// A Boolean value that indicates whether the user is searching.
     @Environment(\.isSearching) private var isSearching
     
-    /// The horizontal size class of the enviroment.
-    @Environment(\.horizontalSizeClass) var horizontalSizeClass
-    
-    /// The vertical size class of the enviroment.
-    @Environment(\.verticalSizeClass) var verticalSizeClass
-    
     /// All samples that will be displayed in the list.
     let samples: [Sample]
     
@@ -52,7 +46,7 @@ struct SampleList: View {
     
     var body: some View {
         List(displayedSamples, id: \.name, selection: $selection) { sample in
-            SampleRow(sample: sample, selection: selection, horizontalSizeClass: horizontalSizeClass, verticalSizeClass: verticalSizeClass)
+            SampleRow(sample: sample, selection: selection)
         }
         .navigationTitle("Samples")
         .toolbar {
@@ -72,6 +66,12 @@ struct SampleList: View {
 
 private extension SampleList {
     struct SampleRow: View {
+        /// The horizontal size class of the enviroment.
+        @Environment(\.horizontalSizeClass) var horizontalSizeClass
+        
+        /// The vertical size class of the enviroment.
+        @Environment(\.verticalSizeClass) var verticalSizeClass
+        
         /// The sample displayed in the row.
         let sample: Sample
         
@@ -80,12 +80,6 @@ private extension SampleList {
         
         /// The current device type.
         let deviceType = UIDevice.current.userInterfaceIdiom
-        
-        /// The horizontal size class of the device.
-        let horizontalSizeClass: UserInterfaceSizeClass?
-        
-        /// The vertical size class of the device.
-        let verticalSizeClass: UserInterfaceSizeClass?
         
         /// A Boolean value indicating whether to show the sample's description
         @State private var isShowingDescription = false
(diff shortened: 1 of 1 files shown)


Shared/Supporting Files/Views/SampleList.swift
```diff
@@ -65,6 +69,12 @@ private extension SampleList {
         /// The sample displayed in the row.
         let sample: Sample
         
+        /// A string representation of the selected sample.
+        let selection: String?
+        
+        /// The current device type.
+        let deviceType = UIDevice.current.userInterfaceIdiom
```
philium, line 76: When it comes to UI. design, checking the user interface idiom is almost never the right choice. The reason is because of split screen on iPad. The better option is to think in terms of size classes and use the horizontal and vertical size class environment values to determine how the UI should be displayed. That will ensure that the UI looks great on not just iPhone and iPad full screen, but also on iPad split screen.

Shared/Supporting Files/Views/SampleList.swift
```diff
@@ -76,17 +86,29 @@ private extension SampleList {
...
+                                Text(sample.description)
+                                    .foregroundColor(selection == sample.name ? .white.opacity(0.8) : .secondary)
+                                    .font(.caption)
+                                    .transition(.move(edge: .top).combined(with: .opacity))
+                            } else {
+                                Text(sample.description)
+                                    .font(.caption)
+                                    .foregroundColor(.secondary)
+                                    .transition(.move(edge: .top).combined(with: .opacity))
+                            }
```
philium, line 99: There is a lot of repeated information in both branches. If this were just an availability check, just the necessary modifier(s) themselves could go in the branch. For example:

```swift
#if ... {
    .someModifier(...)
}
```

In this case, since there's a non-availability check as well, consider adding a computed property to determine the conditioned value:

```swift
var sampleForegroundColor: Color {
    ...
}
```

And then pass that to the `foregroundColor` modifier.

Remember: When it comes to branching (if, switch, etc.), a good rule of thumb is this: keep them as succinct as possible. If the same thing is in all branches, find a way to move it outside so that it only exists in one place.

saratchandrakarumuri (dismissed):
Tested on Mac Catalyst and iPad. This is now looking good on both light and dark modes

Shared/Supporting Files/Views/SampleList.swift
```diff
@@ -76,17 +86,29 @@ private extension SampleList {
...
                     Spacer()
                     Button {
                         isShowingDescription.toggle()
                     } label: {
-                        Image(systemName: isShowingDescription ? "info.circle.fill" : "info.circle")
+                        if #available(iOS 16, *), deviceType != .phone {
+                            Image(systemName: isShowingDescription ? "info.circle.fill" : "info.circle")
+                                .foregroundColor(selection == sample.name ? .white : .accentColor)
+                        } else {
+                            Image(systemName: isShowingDescription ? "info.circle.fill" : "info.circle")
```
philium, line 110: Since both of these images are variants of one another, it would be better to specify the image name in one place and then use the `symbolVariant(_:)` modifier:

```swift
```

Shared/Supporting Files/Views/SampleList.swift
des12437: Okay, can size classes be used to distinguish an iPad from an iPhone? Both iPhone and iPad have a compact horizontalSizeClass and regular verticalSizeClass (in the default orientation). 

There's a check on the device type since in iOS 16+ the List item selection color is the app's accent color on iPad and Mac Catalyst, and gray on iPhone.

yo1995:
While I'm still experimenting with the changes, here are some discussions 

- SwiftUI list highlight on MacOS https://developer.apple.com/forums/thread/719507
- [Change the stroke/fill color of SF Symbol icon in SwiftUI?](https://stackoverflow.com/questions/56709463/change-the-stroke-fill-color-of-sf-symbol-icon-in-swiftui)
- [Selection color of a NavigationLink in List was different in Simulator and Device (SwiftUI)](https://stackoverflow.com/questions/69993084/selection-color-of-a-navigationlink-in-list-was-different-in-simulator-and-devic)
- App accent colors https://developer.apple.com/design/human-interface-guidelines/foundations/color#app-accent-colors

It is definitely an inconsistency introduced by Apple roughly between iOS 15 and 16, and even their own apps have different behaviors (e.g., System Settings vs Finder). Will try to find the solution that requires minimal changes.

yo1995 (commented):
I'm proposing a shorter, but hacky-er approach here. 😬 

1. Instead of detecting selection, we leverage the imperfect SwiftUI's native behavior. Currently, it seems only `Label` can adapt to both light/dark variants and the select/unselect states, according to a stack overflow post [here](https://stackoverflow.com/a/74353338/14369688).
2. Hence, replace the info button with a label, such as below. Please notice, for some unknown reason, `.labelStyle(.iconOnly)` doesn't work correctly for this label. And I'm not sure why it has the purple tint in the first place.
```swift
Label("", systemImage: isShowingDescription ? "info.circle.fill" : "info.circle")
    .onTapGesture {
        isShowingDescription.toggle()
    }
```
3. Due to the lack of variety of `SwiftUI.Color`, we don't have the [semantic UIColor](https://developer.apple.com/documentation/uikit/uicolor/ui_element_colors) as before. Therefore, I'd suggest adding a computed property in a private extension called `Color.secondaryLabel`, and either use the UIColor equivalent's RGB values, or come up with our own light and dark variants.
4. Replace the `.foregroundColor(.secondary)` with this custom color.
5. Add detailed explanation on why the hack is like this.

The only problem I haven't figured out yet is why the separator line is partial on iPhone.

Shared/Supporting Files/Views/SampleList.swift
philium: I didn't fully understand the problem. Almost every time I've seen a user interface idiom check, size classes should have been checked instead. This seems to be one of the rare exceptions where a user interface idiom check might be appropriate.

philium:
Have you tried overriding the environment's [color scheme](https://developer.apple.com/documentation/swiftui/environmentvalues/colorscheme) to get the desired tint color? Define an environment variable for the tint color and then use an `environment(_:_:)` modifier on the element(s) you want to have the opposite color scheme.

des12437:
> The only problem I haven't figured out yet is why the separator line is partial on iPhone.

Nice solution using the label, it works well except for the iPhone where the separator line is partially covered by the label. If I can't figure out how to fix the separator line then I may keep my latest iteration.

des12437:
> Have you tried overriding the environment's [color scheme](https://developer.apple.com/documentation/swiftui/environmentvalues/colorscheme) to get the desired tint color? 

No, since the color scheme (light or dark) does not affect this issue. The only checks made are for iOS version and device type, since a List's selected item will look different depending on these factors where the selection color is either gray or the app's accent color.

yo1995:
Adding Phil to the reviewers since he already shared some thoughts.

Shared/Supporting Files/Views/SampleList.swift
yo1995: I haven't tried, but does this work with iPhone 14PM horizontal orientation?

des12437: @yo1995 Good catch. It does not. The List item selection color goes from grey to the app's accent color when switching to landscape mode.

Simulator Screen Shot - iPhone 14 Pro Mac iOS 16 2 - 2023-04-14 at 12 09 30

philium: It sounds like checking the horizontal and vertical size class is the right approach.

yo1995 (commented):
https://user-images.githubusercontent.com/9660181/232620165-ef052c29-9301-492d-996c-64742bb0349f.mp4

While I was trying these changes, I spotted a UI issue. When I repetitively tap `Display map` sample row, the first time it works fine. The subsequent taps doesn't show the selection nor the "push" animation, and when navigate back, the UI doesn't look right near the top of the screen. Can you confirm and try to determine what is the cause? Thanks.

Shared/Supporting Files/Views/SampleList.swift
```diff
@@ -65,9 +75,39 @@ private extension SampleList {
...
+        var sampleForegroundColor: Color {
+            selection == sample.name ? .white.opacity(0.8) : .secondary
+        }
+        
+        var descriptionTextColor: Color {
+            selection == sample.name ? .white : .accentColor
+        }
+        
+        var selectionColorIsAccentColor: Bool {
+            if #available(iOS 16, *), deviceType != .phone {
```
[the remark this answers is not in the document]
des12437: Yes, both iPhone in Portrait orientation and and iPad for iOS 16+ have the same size classes (compact horizontally and regular vertically), so the `deviceType` condition is needed.

des12437:
> While I was trying these changes, I spotted a UI issue. When I repetitively tap `Display map` sample row, the first time it works fine. The subsequent taps doesn't show the selection nor the "push" animation, and when navigate back, the UI doesn't look right near the top of the screen. Can you confirm and try to determine what is the cause? (Might be sth to do with the selection state, I guess). Thanks.

Yes I see the same issue where the segue animation only happens the first time only in iOS 16+. Looking into this.

yo1995:
With the changes in #227 , this problem is partially mitigated. On iPad landscape, when a row is first tapped, it shows a translucent gray background. The second tap will make it the accent color.

yo1995:
With the changes in #226 , this PR may be closed if we decide to use `DisclosureGroup`. Feel free to leave a comment in that PR on what you think about the UI.

PR #175, opened 2023-04-20
[Update] Get rid of iCloud SF Symbols usage

yo1995:
## Description

This PR replaces 2 iCloud SF Symbols. Per Apple Use Restrictions

> This symbol may not be modified and may only be used to refer to Apple’s iCloud service.


rolson (approved):

dfeinzimer (approved):
Nice catch

Issue #157, opened 2023-04-05
Crash when pressing main info button


dfeinzimer:
<img width="1130" alt="Screenshot 2023-04-05 at 12 04 13" src="https://user-images.githubusercontent.com/16397058/230180020-49fec570-4ed2-4ed6-ba42-3f513b4c43bb.png">
<img width="596" alt="Screenshot 2023-04-05 at 12 04 57" src="https://user-images.githubusercontent.com/16397058/230180026-4008d6a3-0220-4fd9-9d02-1c432ad0cc3a.png">

PR #177, opened 2023-04-20
[Setup] Fix local bundle ID caused crash

yo1995:
## Description

This PR fixes the info button crash when developing using a local source code package. 

## Linked Issue(s)

- #157 

## How To Test

1. Build without changing the project file. It resolves to the 200.1 released binary package
2. Build after changing the project file to use the local source package.
3. Tap on info button and no crash.
4. You can also test build with swift-daily.

### DIFF
--- a/Shared/Supporting Files/Views/AboutView.swift
+++ b/Shared/Supporting Files/Views/AboutView.swift
@@ -58,7 +58,10 @@ private struct VersionRow: View {
 }
 
 private extension Bundle {
-    static let arcGIS = Bundle(identifier: "com.esri.ArcGIS")!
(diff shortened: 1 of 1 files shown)


dfeinzimer (approved):
Crash appears to be gone!

The SDK version number is still wrong when using the GitHub package
Screenshot 2023-04-20 at 16 55 15

But okay when using binaries. Any idea why?
Screenshot 2023-04-20 at 16 53 32

saratchandrakarumuri (approved):

yo1995:
> The SDK version number is still wrong when using the GitHub package, but okay when using binaries.

Please see CreatePackage.swift around line 268. The version and build numbers are injected when the package is built. By default it is 1.0 I believe.

PR #181, opened 2023-05-22
[New] Set min and max scale

CalebRas:
## Description

This PR implements `Set min and max scale` in `Maps` category.
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-SetMinMaxScale/Shared/Samples/Set%20min%20and%20max%20scale

## Linked Issue(s)

- `swift/issues/4086`

## How To Test

- Spread to zoom in to the min scale.
- Pinch to zoom out to the max scale.

## Screenshots

set-min-and-max-scale


Shared/Samples/Set min and max scale/README.metadata.json
```diff
@@ -0,0 +1,31 @@
...
+        "Basemap",
+        "MapView",
+        "ViewPoint"
+    ],
+    "redirect_from": [],
+    "relevant_apis": [
+        "ArcGISMap",
+        "Basemap",
+        "MapView",
+        "ViewPoint"
```
yo1995, line 25:
```suggestion
        "Basemap",
        "Map",
        "MapView",
        "Viewpoint"
```

Shared/Samples/Set min and max scale/README.metadata.json
```diff
@@ -0,0 +1,31 @@
...
+        "area of interest",
+        "level of detail",
+        "maximum",
+        "minimum",
+        "scale",
+        "viewpoint",
+        "ArcGISMap",
+        "Basemap",
+        "MapView",
+        "ViewPoint"
```
yo1995, line 18:
```suggestion
        "Basemap",
        "Map",
        "MapView",
        "Viewpoint"
```

yo1995 (commented):
First batch of comments on the metadata. Good job! Waiting for you adding the graphics overlay to the view.

Samples.xcodeproj/project.pbxproj
```diff
@@ -1145,6 +1156,7 @@
 				0074ABC428174F430037244A /* Sample.swift in Sources */,
 				E004A6F0284E4B9B002A1FE6 /* DownloadVectorTilesToLocalCacheView.swift in Sources */,
 				4D2ADC4329C26D05003B367F /* AddDynamicEntityLayerView.swift in Sources */,
+				D7EAF35A2A1C023800D822C4 /* SetMinAndMaxScaleView.swift in Sources */,
```
yo1995, line 1159: Don't forget to add the source code to `Copy Source Code Files` in the Build Phases. You can test if the README is correctly added by tapping on the info button in the sample.

image

Samples.xcodeproj/project.pbxproj
```diff
@@ -190,6 +192,7 @@
 			dstPath = "";
 			dstSubfolderSpec = 7;
 			files = (
+				D7E9EF292A1D2219000C4865 /* SetMinAndMaxScaleView.swift in Copy Source Code Files */,
```
dfeinzimer, line 195: I'll let @yo1995 comment on this but it looks like most of the files in this list are alphabetized, should this be as well?

dfeinzimer: Screenshot 2023-05-23 at 11 33 10

yo1995: Dang! I wasn't paying attention to the order of this list until now. 🤦 
It seems Xcode doesn't have a way to sort it in GUI: https://stackoverflow.com/questions/24326147 

However, I don't think the order matters too much for a few reasons

- Xcode doesn't allow us to do so automatically 🤪 and it's error-prone to change order manually
- The source code file names don't strictly follows the sample name, so if sorted alphabetically, it is possible that the swift files for 1 sample is scattered around multiple places
- the order doesn't matter. We only need a "unordered set" of filenames so they are copied to the bundle.
- (a small benefit) Adding the files using the default order preserves the order of when they were added, though this info is not important at all.

so no need to worry about the order.

yo1995 (approved):

dfeinzimer (approved):

PR #182, opened 2023-05-23
[New] Style point with picture marker symbols

CalebRas:
## Description

This PR implements `Style point with picture marker symbols` in `Display information` category.
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-PictureMarkerSymbols/Shared/Samples/Style%20point%20with%20picture%20marker%20symbols

## Linked Issue(s)

- `swift/issues/4087`

## Screenshots

style-point-with-picture-marker-symbols


Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,90 @@
...
+        self.graphicsOverlay.addGraphic(campsiteGraphic)
+    }
+    
+    init() {
+        // Add picture marker symbol using a remote image.
+        self.addPictureMarkerSymbolFromURL()
+        
+        // Add picture marker symbol using image in assets.
+        self.addPictureMarkerSymbolFromImage()
+    }
```
yo1995, line 84: It should not be a problem, but until Apple provides a doc that explains the lifecycle of SwiftUI (like [UIViewController](https://developer.apple.com/documentation/uikit/uiviewcontroller#1652793)), I would refrain setting anything up that depends on a SwiftUI component, in the initializer of a SwiftUI View. Here because the graphics overlay is a state variable, it might have unknown side effects. Putting it in the task block is a bit safer.

```swift
        MapView(map: map, graphicsOverlays: [graphicsOverlay])
            .task {
                // Add picture marker symbol using a remote image.
                addPictureMarkerSymbolFromURL()
                
                // Add picture marker symbol using image in assets.
                addPictureMarkerSymbolFromImage()
            }
```

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    var body: some View {
+        // Creates a map view to display the map and point graphics.
```
yo1995, line 87: One more thing: we use [3rd person](https://stackoverflow.com/questions/8910214/comment-style-imperative-or-third-person) for 3-slash comments. We don't have requirement for 2-slash, but better keep them consistent.

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
philium: As a value type, a SwiftUI view instance doesn't have a lifecycle per se. A view is instantiated whenever the parent's body is computed, which can happen any number of times. As part of the initializer, this code would add the two graphics to the graphics overlay whenever the parent view is re-computed, which isn't great. While using a `task` modifier would prevent that from happening, but since these aren't asynchronous methods, that would be an abuse of that modifier as it is designed for performing asynchronous operations. Instead, the actions performed by these two methods should just be part of creating the graphics overlay. Something like this:

```swift
@State private var graphicsOverlay: GraphicsOverlay = {
    let graphics = [
        makePictureMarkerSymbolGraphicFromImage(),
        makePictureMarkerSymbolGraphicFromURL()
    ]
    return GraphicsOverlay(graphics: graphics)
}()
```

private static func makePictureMarkerSymbolGraphicFromImage() -> Graphic { ... }

private static func makePictureMarkerSymbolGraphicFromURL() -> Graphic { ... }

yo1995: Agreed. Using type property or type factory method make sense.

yo1995 (commented):
Please see Phil's comment.

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,84 @@
...
+    @State private var map: Map = {
+        let map = Map(basemapStyle: .arcGISTopographic)
+        map.initialViewpoint = Viewpoint(
+            center: Point(x: -225166.5, y: 6551249, spatialReference: .webMercator),
+            scale: 1e5
+        )
+        return map
+    }()
+    
+    /// Create a picture marker symbol from an image in the project assets.
```
yo1995, line 38:
```suggestion
    /// Creates a picture marker symbol from an image in the project assets.
```

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,84 @@
...
+        // Create the location for pin.
+        let pinPoint = Point(x: -226773, y: 6550477, spatialReference: .webMercator)
+        
+        // Create the graphic for pin.
+        let pinGraphic = Graphic(geometry: pinPoint, symbol: pinSymbol)
+        
+        return pinGraphic
+    }
+    
+    /// Create a picture marker symbol using a remote image.
```
yo1995, line 57:
```suggestion
    /// Creates a picture marker symbol using a remote image.
```

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,84 @@
...
+        let map = Map(basemapStyle: .arcGISTopographic)
+        map.initialViewpoint = Viewpoint(
+            center: Point(x: -225166.5, y: 6551249, spatialReference: .webMercator),
+            scale: 1e5
+        )
+        return map
+    }()
+    
+    /// Create a picture marker symbol from an image in the project assets.
+    private static func makePictureMarkerSymbolFromImage() -> Graphic {
```
yo1995, line 39:
```suggestion
    /// Create a picture marker symbol from an image in the project assets.
    /// - Returns: A picture marker symbol.
```

In Xcode you can add a 3-slash documentation template (Xcode Markup, or commonly referred as docstring) by Editor > Selection > Add Documentation. Shortcut is `cmd + option + /`.

We add docstring for all non-trivial methods and properties.

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,84 @@
...
+        let pinPoint = Point(x: -226773, y: 6550477, spatialReference: .webMercator)
+        
+        // Create the graphic for pin.
+        let pinGraphic = Graphic(geometry: pinPoint, symbol: pinSymbol)
+        
+        return pinGraphic
+    }
+    
+    /// Create a picture marker symbol using a remote image.
+    private static func makePictureMarkerSymbolFromURL() -> Graphic {
```
yo1995, line 58: Add one for this method as well.

yo1995 (dismissed):
Thanks for writing this sample! Can you do one more re-indent of the whole file? For some reason the whitespaces aren't consistent in it.

yo1995 (dismissed):

yo1995 (approved):

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,86 @@
...
+    /// Creates a picture marker symbol from an image in the project assets.
+    /// - Returns: A picture marker symbol
+    private static func makePictureMarkerSymbolFromImage() -> Graphic {
+        let imageName = "PinBlueStar"
+        
+        // Create pin symbol using the image.
+        let pinSymbol = PictureMarkerSymbol(image: UIImage(named: imageName)!)
+        
+        // Change offsets, so the symbol aligns properly to the point.
+        pinSymbol.offsetY = pinSymbol.image!.size.height / 2
```
mhdostal, line 47: You should probably at least mention the fact that `PictureMarkerSymbol` is `Loadable`, as this same code, if it was used in `makePictureMarkerSymbolFromURL()` would crash, because `image` is `nil` until loaded. Or maybe an `if let image = pinSymbol.image`.

mhdostal (changes requested):

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
CalebRas: I added a guard statement in both the makePictureMarkerSymbolFromImage() and makePictureMarkerSymbolFromURL() methods. Let me know if that is what you were thinking or if you want something different.

mhdostal: Since the user can overcome this limitation by simply calling `load()` on the symbol, I'd prefer not to explicitly check to see if the symbol is loaded (even though `load()` is `async throws` and not so easy to implement here).

I added a suggestion below, instead of the `guard` you added.

mhdostal (changes requested):

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,103 @@
...
+    private static func makePictureMarkerSymbolFromImage() -> Graphic {
+        let imageName = "PinBlueStar"
+        
+        // Create pin symbol using the image.
+        let pinSymbol = PictureMarkerSymbol(image: UIImage(named: imageName)!)
+        
+        // Ensure the pin symbol loaded.
+        guard pinSymbol.loadStatus != .failed else {
+            return Graphic()
+        }
```
mhdostal, line 49: I'd just remove this, as it's not necessary.
```suggestion
```

Shared/Samples/Style point with picture marker symbols/StylePointWithPictureMarkerSymbolsView.swift
```diff
@@ -0,0 +1,103 @@
...
+        // `PictureMarkerSymbol(url: URL)` initializer, the `image` property
+        // will be `nil` until the symbol is loaded using the
+        // `func load() async throws` method. We are not accessing that
+        // property here, so we don't need to explicitly load it. The symbol
+        // will be loaded automatically prior to drawing.
+        
+        // Ensure the campsite symbol loaded.
+        guard campsiteSymbol.loadStatus != .failed else {
+            return Graphic()
+        }
```
mhdostal, line 83: Same here. You're not accessing `.image`, so no issues with the code without this guard.
```suggestion
```

mhdostal (approved):
Nice work!

yo1995 (approved):

PR #185, opened 2023-05-26
[New] Create buffers around points

CalebRas:
## Description

This PR implements `Create buffers around points` in `Geometry` category.
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-CreateBuffersAroundPoints/Shared/Samples/Create%20buffers%20around%20points

## Linked Issue(s)

- `swift/issues/4100`

## How To Test

* Tap outside the boundary to see "Tap with the boundary to add buffer" message at the top.
* Tap inside the boundary to add a point. 
* Enter a value, such as 100.
* Tap done. (See "Buffer created." message)
* Add another buffer close enough to the first that they will overlap.
* Toggle "Union" on and off to see the buffers merge and un-merge.
* Tap "Clear" to remove all the buffers and points from the map. (See "Tap on the map to add buffers" message.)

## Screenshots

create-buffers-around-points

### DIFF
--- a/Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
+++ b/Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
@@ -61,7 +61,7 @@ struct CreateBuffersAroundPointsView: View {
                 TextField("100", text: $model.radiusInput)
                     .keyboardType(.numberPad)
                 Button("Done") {
-                    // Check to ensure the tapPoint is within the boundary
+                    // Check to ensure the tapPoint is within the boundary.
                     guard model.boundaryContains(model.tapPoint) else {
                         status = .outOfBoundsTap
                         return


Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,295 @@
...
+    /// An Enum for the sample status.
+    enum Status {
+        case addPoints
+        case bufferCreated
+        case outOfBoundsTap
+        case noPoints
+    }
+    
+    /// Get the text message associated with the current status for the overlay.
+    /// - Returns: A 'String' message.
```
yo1995, line 282:
```suggestion
    /// - Returns: A status message.
```
Also we use single ticks for symbol names, so they look like code in the quick help.
image

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,295 @@
...
+        private let boundaryPolygon: Polygon
+        
+        init() {
+            /// The spatial reference for this sample.
+            let statePlaneNorthCentralTexas = SpatialReference(wkid: WKID(32038)!)!
+            
+            // Create boundary polygon.
+            boundaryPolygon = {
+                let boundaryPoints = [
+                    Point(x: -103.070, y: 31.720, spatialReference: .wgs84),
```
yo1995, line 126:
```swift
                    Point(latitude: 31.720, longitude: -103.070),
```
We use this convenience initializer when it comes to WGS84, so it is more obvious to non-GIS users.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,295 @@
...
+                    Point(x: -103.070, y: 34.580, spatialReference: .wgs84),
+                    Point(x: -94.000, y: 34.580, spatialReference: .wgs84),
+                    Point(x: -94.00, y: 31.720, spatialReference: .wgs84)
+                ]
+                let polygon = GeometryEngine.project(Polygon(points: boundaryPoints), into: statePlaneNorthCentralTexas)
+                return polygon!
```
yo1995, line 132:
```suggestion
                let polygon = GeometryEngine.project(Polygon(points: boundaryPoints), into: statePlaneNorthCentralTexas)!
                return polygon
```
In this type of scenarios, we unwrap as soon as possible.

yo1995 (commented):
First round of comments.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,296 @@
...
+                                model.drawBuffers()
+                            }
+                        }
+                    // Clear button.
+                    Button("Clear") {
+                        model.clearBufferPoints()
+                        model.status = .addPoints
+                    }
+                    .disabled(model.bufferPoints.isEmpty)
+                }
```
des12437, line 56: Should the "Undo" button be added @yo1995 ? The undo button was included in the Objective-C Sample and [removes](https://github.com/Esri/arcgis-runtime-samples-ios/blob/main/arcgis-ios-sdk-samples/Geometry/Buffer%20list/BufferListViewController.swift#L134) the previously added radius.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,296 @@
...
+            .alert("Buffer Radius", isPresented: $model.inputBoxIsPresented, actions: {
+                TextField("100", text: $model.radiusInput)
+                    .keyboardType(.numberPad)
+                // Input alert done button.
+                Button("Done") {
+                    model.addBufferPoint()
+                    model.radiusInput = ""
+                    model.drawBuffers()
+                    model.status = Status.bufferCreated
+                }
```
des12437, line 67: Please add input validation here (`radiusInput` > 0 && `radiusInput` < 300). Pressing the "Done" button with a radius less than or equal to zero will add a point to the map with no buffer which is unexpected, and will also add a buffer with a radius of any size when the radius is greater than 300 which is also unexpected behavior.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
CalebRas: I think we decided that to forgo the Undo button since there was nothing in the [common design](https://devtopia.esri.com/runtime/common-samples/tree/main/designs/CreateBuffersAroundPoints) about it.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,306 @@
...
+                boundaryGeometry: boundaryPolygon
+            )
+            bufferGraphicsOverlay = CreateBuffersAroundPointsView.Model.makeBufferGraphicsOverlay()
+            tapPointsGraphicsOverlay = CreateBuffersAroundPointsView.Model.makeTappedPointsGraphicsOverlay()
+            graphicsOverlays.append(
+                contentsOf: [boundaryGraphicsOverlay, bufferGraphicsOverlay, tapPointsGraphicsOverlay]
+            )
+        }
+        
+        /// Create a 'Map' with image layers from a 'SpatialReference'.
```
des12437, line 150:
```suggestion
        /// Creates a map with image layers from a spatial reference.
```
Applies to other make methods as well.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,306 @@
...
+        func clearBufferPoints() {
+            bufferPoints.removeAll()
+            bufferGraphicsOverlay.removeAllGraphics()
+            tapPointsGraphicsOverlay.removeAllGraphics()
+        }
+    }
+}
+
+private extension CreateBuffersAroundPointsView {
+    /// An 'Enum' for the sample status.
```
des12437, line 281:
```suggestion
    /// An enumeration for the sample status.
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,306 @@
...
+    enum Status {
+        case addPoints
+        case bufferCreated
+        case outOfBoundsTap
+        case invalidInput
+        case noPoints
+    }
+    
+    /// Get the text message associated with the current status for the overlay.
```
des12437, line 290:
```suggestion
    /// Gets the text message associated with the current status for the overlay.
```

des12437 (commented):
The documentation reads well, but the style is not consistent with other samples in terms of using code voice to refer to a type. See the documentation in the other samples for reference.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,306 @@
...
+                ToolbarItemGroup(placement: .bottomBar) {
+                    // Union toggle switch.
+                    Toggle("Union", isOn: $model.shouldUnion)
+                        .toggleStyle(.switch)
+                        .onChange(of: model.shouldUnion) { _ in
+                            if !model.bufferPoints.isEmpty {
+                                model.drawBuffers()
+                            }
+                        }
+                    // Clear button.
```
des12437, line 50:
```suggestion
```
This documentation is probably not necessary.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,306 @@
...
+                        model.clearBufferPoints()
+                        model.status = .addPoints
+                    }
+                    .disabled(model.bufferPoints.isEmpty)
+                }
+            }
+            .alert("Buffer Radius", isPresented: $model.inputBoxIsPresented, actions: {
+                TextField("100", text: $model.radiusInput)
+                    .keyboardType(.numberPad)
+                // Input alert done button.
```
des12437, line 61:
```suggestion
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,306 @@
...
+                Button("Done") {
+                    model.addBufferPoint()
+                    model.radiusInput = ""
+                    model.drawBuffers()
+                }
+                // Input alert cancel button.
```
des12437, line 67:
```suggestion
```

des12437 (approved):

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,303 @@
...
+            bufferGraphicsOverlay = CreateBuffersAroundPointsView.Model.makeBufferGraphicsOverlay()
+            tapPointsGraphicsOverlay = CreateBuffersAroundPointsView.Model.makeTappedPointsGraphicsOverlay()
+            graphicsOverlays.append(
+                contentsOf: [boundaryGraphicsOverlay, bufferGraphicsOverlay, tapPointsGraphicsOverlay]
+            )
+        }
+        
+        /// Create a map with image layers from a spatial reference.
+        /// - Parameters:
+        ///   - spatialReference: The 'SpatialReference' the 'Map' is derived from.
```
yo1995, line 149:
```suggestion
        ///   - spatialReference: The `SpatialReference` the `Map` is derived from.
```
|Before|After|
|:-:|:-:|
|image|image|

Use backtick or plain English for markup docstring. Please fix other places as well.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,303 @@
...
+            let graphicsOverlay = GraphicsOverlay()
+            let circleSymbol = SimpleMarkerSymbol(style: .circle, color: .red, size: 10)
+            graphicsOverlay.renderer = SimpleRenderer(symbol: circleSymbol)
+            return graphicsOverlay
+        }
+        
+        /// Check if a point is within the valid area of use for this sample.
+        /// - Parameter point: A 'Point' to validate.
+        /// - Returns: A 'Bool' indicating whether it is within bounds.
+        func isWithinBounds(_ point: Point) -> Bool {
```
yo1995, line 199: The method name reads a bit unnatural to me. I feel it should be similar to the geometry engine method name, sth like "boundaryContains(_:)". Will leave it open to thoughts.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,303 @@
...
+                }
+                // Input alert message.
+            }, message: {
+                Text("Please enter a number between 0 and 300 miles.")
+            })
+    }
+}
+
+private extension CreateBuffersAroundPointsView {
+    // The view model for this sample.
```
yo1995, line 76:
```suggestion
    /// The view model for this sample.
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,303 @@
...
+                }
+                
+                // Update the buffer radius with the text value.
+                let radiusInMiles = Measurement(value: radius, unit: UnitLength.miles)
+                
+                // The spatial reference in this sample uses US feet as its unit.
+                let radiusInFeet = radiusInMiles.converted(to: .feet).value
+                
+                // Add point with radius to bufferPoints Array.
+                bufferPoints.append((point: tapPoint!, radius: radiusInFeet))
```
yo1995, line 263:
```suggestion
                bufferPoints.append((point: tapPoint, radius: radiusInFeet))
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,303 @@
...
+        case addPoints
+        case bufferCreated
+        case outOfBoundsTap
+        case invalidInput
+        case noPoints
+    }
+    
+    /// Gets the text message associated with the current status for the overlay.
+    /// - Returns: A 'String' status message.
+    func getStatusText(_ status: Status) -> String {
```
yo1995, line 289: We typically put this descriptive string as part of the enum itself, i.e.,

See an example in `ShowResultOfSpatialOperationsView.SpatialOperation`

```swift
        var label: String {
            switch self {
            case .addPoints:
                return "Tap on the map to add buffers."
            case .bufferCreated:
                return "Buffer created."
            case .outOfBoundsTap:
                return "Tap within the boundary to add buffer."
            case .invalidInput:
                return "Enter a value between 0 and 300 to create a buffer."
            case .noPoints:
                return "Add a point to draw the buffers."
            }
        }
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,303 @@
...
+                    .disabled(model.bufferPoints.isEmpty)
+                }
+            }
+            .alert("Buffer Radius", isPresented: $model.inputBoxIsPresented, actions: {
+                TextField("100", text: $model.radiusInput)
+                    .keyboardType(.numberPad)
+                Button("Done") {
+                    model.addBufferPoint()
+                    model.radiusInput = ""
+                    model.drawBuffers()
```
yo1995, line 63: The status string's state transition can be improved. For instance, the `addBufferPoint` is called upon `Done` button tap, and the done button is only shown when the `inputBoxIsPresented` is set to `true`; the out of bound check is already performed when the boolean is togged, so no need to check again.

I'd suggest either keep all the status checks close to the UI inputs, or to where the data is processed.

yo1995 (commented):
2nd batch of comments. Good work!

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
CalebRas: I originally tried to do that but the "Done" button won't show at all if the .disabled modifier is used on it since it is disabled with the Alert loads. I did move all of the status checks/updates to the UI, though. Let me know if that was not what you were thinking.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -0,0 +1,281 @@
...
+                        status = .addPoints
+                    }
+                    .disabled(model.bufferPoints.isEmpty)
+                }
+            }
+            .alert("Buffer Radius", isPresented: $model.inputBoxIsPresented, actions: {
+                TextField("100", text: $model.radiusInput)
+                    .keyboardType(.numberPad)
+                Button("Done") {
+                    // Check to ensure the tapPoint is within the boundary
```
yo1995, line 64:
```suggestion
                    // Check to ensure the tapPoint is within the boundary.
```

yo1995 (approved):

des12437 (approved):

yo1995 (approved):

PR #190, opened 2023-05-31
[New] Show coordinates in multiple formats

CalebRas:
## Description

This PR implements `Show coordinates in multiple formats` in `Geometry` category.
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-FormatCoordinates/Shared/Samples/Show%20coordinates%20in%20multiple%20formats

## Linked Issue(s)

- `swift/issues/4129`

## How To Test

* Tap on the map to see the different coordinate formats for the tapped point.

## Screenshots

show-coordinates-in-multiple-formats


Shared/Samples/Show coordinates in multiple formats/ShowCoordinatesInMultipleFormatsView.swift
```diff
@@ -0,0 +1,167 @@
...
+        var map = Map(basemapStyle: .arcGISImageryStandard)
+        
+        /// The graphics overlay for the point graphic.
+        lazy var graphicsOverlay = GraphicsOverlay(graphics: [pointGraphic])
+        
+        /// A yellow cross graphic for the map point.
+        private var pointGraphic: Graphic = {
+            let yellowCrossSymbol = SimpleMarkerSymbol(style: .cross, color: .yellow, size: 20)
+            return Graphic(symbol: yellowCrossSymbol)
+        }()
```
yo1995, line 114: These can be set in the initializer. Also, in most cases we can use `let`.

Shared/Samples/Show coordinates in multiple formats/ShowCoordinatesInMultipleFormatsView.swift
```diff
@@ -0,0 +1,167 @@
...
+        
+        /// The point on the map.
+        @Published var mapPoint: Point!
+        
+        /// The decimal degrees text.
+        @Published var latLongDDTextField = ""
```
yo1995, line 120: Instead of "…TextField", change their names to "…String" to make them read better

Shared/Samples/Show coordinates in multiple formats/ShowCoordinatesInMultipleFormatsView.swift
```diff
@@ -0,0 +1,167 @@
...
+        
+        /// The UTM text.
+        @Published var utmString = ""
+        
+        /// the USNG text.
+        @Published var usngString = ""
+        
+        init() {
+            map = Map(basemapStyle: .arcGISImageryStandard)
+            pointGraphic = Graphic(symbol: SimpleMarkerSymbol(style: .cross, color: .yellow, size: 20))
```
zkline101, line 130: These declarations can both be used where the property is declared they don't need to be in the init.

Shared/Samples/Show coordinates in multiple formats/ShowCoordinatesInMultipleFormatsView.swift
```diff
@@ -0,0 +1,167 @@
...
+            
+            utmString = CoordinateFormatter.utmString(
+                from: mapPoint,
+                conversionMode: .latitudeBandIndicators,
+                addSpaces: true)
+            
+            usngString = CoordinateFormatter.usngString(
+                from: mapPoint,
+                precision: 4,
+                addSpaces: true)
```
zkline101, line 164:
```suggestion
            latLongDMSString = CoordinateFormatter.latitudeLongitudeString(
                from: mapPoint,
                format: .degreesMinutesSeconds,
                decimalPlaces: 1
            )
            
            utmString = CoordinateFormatter.utmString(
                from: mapPoint,
                conversionMode: .latitudeBandIndicators,
                addSpaces: true
            )
            
            usngString = CoordinateFormatter.usngString(
                from: mapPoint,
                precision: 4,
                addSpaces: true
            )
```

zkline101 (changes requested):

Shared/Samples/Show coordinates in multiple formats/ShowCoordinatesInMultipleFormatsView.swift
```diff
@@ -0,0 +1,179 @@
...
+        var map = Map(basemapStyle: .arcGISImageryStandard)
+        
+        /// The graphics overlay for the point graphic.
+        let graphicsOverlay: GraphicsOverlay
+        
+        /// A yellow cross graphic for the map point.
+        private let pointGraphic = Graphic(symbol: SimpleMarkerSymbol(style: .cross, color: .yellow, size: 20))
+        
+        /// The point on the map.
+        @Published var mapPoint: Point!
```
yo1995, line 125: A second thought: I feel it is better to be a state variable in the view than a published property in the model.

1. It doesn't drive UI updates, so it doesn't need to be published
2. I feel in most cases the published properties are private(set), public get. If it is both public set and get, then just staying in the view is fine.
3. If this suggestion is applied, method signature can change to `private func updateCoordinateFields(mapPoint: Point)`.

yo1995 (commented):
Another problem I noticed is when the sample is in iPhone 14 PM landscape orientation, the textfields take up the entire screen. I don't think it is an issue or a need to add a scroll view as no one gonna use the sample that way. 😼

yo1995 (dismissed):

Shared/Samples/Show coordinates in multiple formats/ShowCoordinatesInMultipleFormatsView.swift
```diff
@@ -0,0 +1,181 @@
...
+
+import ArcGIS
+import SwiftUI
+
+struct ShowCoordinatesInMultipleFormatsView: View {
+    /// The view model for the sample.
+    @StateObject private var model = Model()
+    
+    /// The tapped point on the map.
+    @State private var mapPoint = Point(latitude: 0, longitude: 0)
```
zkline101, line 23: I think `tappedPoint` would better describe this property.

zkline101:
> Another problem I noticed is when the sample is in iPhone 14 PM landscape orientation, the textfields take up the entire screen. I don't think it is an issue or a need to add a scroll view as no one gonna use the sample that way. 😼

Shouldn't we fix that though? We can't really predict how the user is going to use the app but I think we should make sure the sample looks right in all scenrios.

yo1995:
I was afraid that might be too much for Caleb to implement, and add a lot of adaptive UI code that isn't related to the sample. But I'm happy if you want to try that out, either with an adaptive stack view like the old sample, or sth like a LazyVGrid in a scroll view.

zkline101 (approved):

yo1995 (approved):
If you want, you can add a note and rework the UI in the second half of the internship.

Issue #137, opened 2023-03-23
Use `ObservedObject` instead of `EnvironmentObject` for settings


yo1995:
https://www.hackingwithswift.com/quick-start/swiftui/whats-the-difference-between-observedobject-state-and-environmentobject

In most of the samples, the complexity doesn't need an environment object as there are only a few views.

yo1995:
This is fixed in #314

PR #191, opened 2023-05-31
[New] Densify and generalize geometry

CalebRas:
## Description

This PR implements `Densify and generalize geometry` in `Geometry` category.
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-DensifyAndGeneralize/Shared/Samples/Densify%20and%20generalize%20geometry

## Linked Issue(s)

- `swift/issues/4132`

## How To Test

* Tap the Options button to bring up the sheet.
* Toggle Generalize and move the slider to adjust the Max Deviation.
* Toggle Densify and move the slider to adjust the Max Segment Length.
* Toggle both off to remove the result line and points.
* Tap Reset to reset the Option sheet values.

## Screenshots

densify-and-generalize-geometry


Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.SettingsView.swift
```diff
@@ -0,0 +1,69 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+extension DensifyAndGeneralizeGeometryView {
+    struct OptionsView: View {
+        /// The view model for the sample.
+        @EnvironmentObject private var model: Model
```
yo1995, line 21: Instead of using `EnvironmentObject`, can you change to `ObservedObject`? Both works but I'm thinking of improving it. #137

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.SettingsView.swift
```diff
@@ -0,0 +1,69 @@
...
+                        }
+                    VStack {
+                        HStack {
+                            Text("Max Deviation")
+                            Spacer()
+                            Text(String(Int(model.maxDeviation)))
+                        }
+                        Slider(value: $model.maxDeviation, in: 1...250) { _ in
+                            model.updateGraphics()
+                        }
```
yo1995, line 39: The slider's [`onEditingChanged`](https://developer.apple.com/documentation/swiftui/slider) closure doesn't match the old behavior. The `onEditingChanged` closure is only called twice: touch down (true) and touch up (false) on the knob. Whereas the old one changes constantly when the values are updated.

I'll leave it open for the other reviewer's thoughts. I'd like to have the old behavior but the new one isn't a big problem.

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
```diff
@@ -0,0 +1,184 @@
...
+        /// A map with a Streets (Night) basemap centered on the polyline.
+        lazy var map: Map = {
+            let map = Map(basemapStyle: .arcGISStreetsNight)
+            
+            // Set the initial viewpoint to show the extent of the graphics.
+            map.initialViewpoint = Viewpoint(
+                center: originalPolyline.extent.center,
+                scale: 65907)
+            return map
+        }()
```
yo1995, line 53: A general rule of thumb of mine, if a property is guaranteed to be used, don't use `lazy` to manipulate its initialization order. lazy should be conservatively used for those resource consuming types that might not be used (FileManager as an example). There should be a way to play with the order to make everything load.

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
```diff
@@ -0,0 +1,184 @@
...
+            shouldDensify = false
+            maxDeviation = 10.0
+            shouldGeneralize = false
+            maxSegmentLength = 100.0
+            updateGraphics()
+        }
+        
+        /// Update the result polyline and mulitpoint graphics after densifying
+        /// and generalizing.
+        func updateGraphics() {
```
yo1995, line 146: It may not follow the design, but I felt like generalize and densify should be separated into 2 methods, and the UI controls should be mutually exclusive. Currently it is "first generalize then densify" if both toggles are on. I'll leave it open for other reviewer's thoughts.

rolson:
Sorry, I missed this PR.  I will get to it ASAP.

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
```diff
@@ -0,0 +1,188 @@
...
+            originalPolyline = Polyline(points: pointCollection)
+            
+            // Create map.
+            map = Map(basemapStyle: .arcGISStreetsNight)
+            
+            // Set the initial viewpoint to show the extent of the graphics.
+            map.initialViewpoint = Viewpoint(
+                center: originalPolyline.extent.center,
+                scale: 65907
+            )
```
rolson, line 115: I'm not sure what the sample coding standards are, but this could happen as part of the declaration of the `map` variable if you wanted it to.

something like:

```swift
let map = {
    ...
    return map
}()
```

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
```diff
@@ -0,0 +1,188 @@
...
+            if shouldGeneralize {
+                resultPolyline = GeometryEngine.generalize(
+                    resultPolyline,
+                    maxDeviation: maxDeviation,
+                    removeDegenerateParts: true
+                ) as! Polyline
+            }
+            // Densify the points of the polyline with the specified max segment
+            if shouldDensify {
+                resultPolyline = GeometryEngine.densify(
```
rolson, line 171: not sure why this isn't generic in the Swift API. I'm assuming there is a reason, but I'd like to know why.  Same question for generalize(...) above.

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
rolson: FWIW, I agree.

CalebRas: I tried to implement it in two different methods, and I am not sure it is the right approach. Since each method only either densifies or generalizes and then updates the graphics accordingly, they each throw away the result of the other. That's great if that's what you want, but want to see the results combined, it would be much less complex and make more sense to have them in the same method since work they on the same polyline. Let me know what you guys think though.

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
CalebRas: I had it done in the constructor since you can't set the initial viewpoint in a property initializer because you use the original polyline to do so and the property initializers run before 'self' is available. I moved map creation up there though and kept the setting of the initial viewpoint in the constructor.

Shared/Samples/Densify and generalize geometry/DensifyAndGeneralizeGeometryView.swift
CalebRas: I am not totally sure.

rolson: looks like it's because you can pass an envelope to these methods and get a polygon.

rolson (dismissed):

Shared/Samples/Densify and generalize geometry/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Use case
+
+The sample shows a polyline representing a ship's location at irregular intervals. The density of vertices along the ship's route is appropriate to represent the path of the ship at the sample map view's initial scale. However, that level of detail may be too great if you wanted to show a polyline of the ship's movement down the whole of the Willamette river. Then, you might consider generalizing the polyline to still faithfully represent the ship's passage on the river without having an overly complicated geometry.
+
+Densifying a multipart geometry can be used to more accurately represent curved lines or to add more regularity to the vertices making up a multipart geometry.
+
+## How to use the sample
+
+Tap the Option button to open the settings. Use the sliders to control the parameters of the densify and generalize methods. You can deselect the toggles for either method to remove its effect from the result polyline.
```
yo1995, line 15:
```suggestion
Tap the "Option" button to open the settings. Use the sliders to control the parameters of the densify and generalize methods. You can deselect the toggles for either method to remove its effect from the result polyline.
```

Shared/Samples/Densify and generalize geometry/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## How it works
+
+1. Use the static method `GeometryEngine.densify(polyline, maxSegmentLength)` to densify the polyline object. The resulting polyline object will have more points along the line, so that there are no points greater than `maxSegmentLength` from the next point.
```
yo1995, line 19:
```suggestion
1. Use the static method `static GeometryEngine.densify(_:maxSegmentLength:)` to densify the polyline object. The resulting polyline object will have more points along the line, so that there are no points greater than `maxSegmentLength` from the next point.
```

Shared/Samples/Densify and generalize geometry/README.md
```diff
@@ -0,0 +1,35 @@
...
+2. Use the static method `GeometryEngine.generalize(polyline, maxDeviation, true)` to generalize the polyline object. The resulting polyline object will have points shifted from the original line to simplify the shape. None of these points can deviate farther from the original line than `maxDeviation`. The last parameter, `removeDegenerateParts`, will clean up extraneous parts of a multipart geometry. This will have no effect in this sample as the polyline does not contain extraneous parts.
```
yo1995, line 20:
```suggestion
2. Use the static method `static GeometryEngine.generalize(_:maxDeviation:removeDegenerateParts:)` to generalize the polyline object. The resulting polyline object will have points shifted from the original line to simplify the shape. None of these points can deviate farther from the original line than `maxDeviation`. The last parameter, `removeDegenerateParts`, will clean up extraneous parts of a multipart geometry. This will have no effect in this sample as the polyline does not contain extraneous parts.
```

yo1995 (commented):
Good stuff!

yo1995 (approved):

rolson (approved):

PR #201, opened 2023-06-12
[New] Measure distance in scene

CalebRas:
## Description

This PR implements `Measure distance in scene` in `Analysis` category.
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-MeasureDistanceInScene/Shared/Samples/Measure%20distance%20in%20scene

## Linked Issue(s)

- `swift/issues/4158`

## How To Test

* Tap on the map to set the location distance measurement start location.
* Tap again to set the location distance measurement end location.
* Tap on "Imperial" to change the location distance measurement unit system.

## Screenshots

measure-distance-in-scene

## To Discuss

* As of PR creation, this sample does not implement any drag functionality.


CalebRas:
@pgruenler I figured out the issue with points and added drag functionality so it should me good to go now. Let me know if you have any more problems.

CalebRas:
@yo1995 Can you let me know if you think there is a better way to use the task blocks and if you think that this sample needs a view model?

yo1995:
> Can you let me know if you think there is a better way to use the task blocks 

Will comment when I review.

> if you think that this sample needs a view model

This is a tough call. 4 objects (scene, analysis overlay, formatter, measurement) typically drive me towards a data model, but since the sample is only 150 lines I'll leave it up to you.

CalebRas:
Sounds good. I ended up adding a view model.

Shared/Samples/Measure distance in scene/README.md
```diff
@@ -0,0 +1,34 @@
...
+## How to use the sample
+
+Choose a unit system for the measurement in the segmented control. Tap any location in the scene to start measuring. Then tap and drag to an end location, and lift your finger to complete the measure. Tap a new location to start a new measurement.
+
+## How it works
+
+1. Create an `AnalysisOverlay` object and add it to the analysis overlay collection of the `SceneView` object.
+2. Specify the start location and end location to create a `LocationDistanceMeasurement` object. Initially, the start and end locations can be the same point.
+3. Add the location distance measurement analysis to the analysis overlay.
+4. The `measurementChanged` callback will trigger if the distances change. You can get the new values for the `directDistance`, `horizontalDistance`, and `verticalDistance` from the `MeasurementChangedEvent` object returned by the callback.
```
pgruenler, line 20: I'm not exactly sure how to talk about these values but they are more of streamed properties in the Swift API rather that objects returned from a callback.

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
```diff
@@ -0,0 +1,183 @@
...
+        let measurementFormatter: MeasurementFormatter = {
+            let measurementFormatter = MeasurementFormatter()
+            measurementFormatter.unitOptions = .providedUnit
+            measurementFormatter.numberFormatter.minimumFractionDigits = 2
+            measurementFormatter.numberFormatter.maximumFractionDigits = 2
+            return measurementFormatter
+        }()
+        
+        init() {
+            // Set scene the viewpoint specified by the location distance measurement.
```
pgruenler, line 151:
```suggestion
            // Set scene to the viewpoint specified by the location distance measurement.
```

pgruenler (commented):
Looks good and works well. Just a couple comments.

Samples.xcodeproj/project.pbxproj
```diff
@@ -368,9 +371,11 @@
 		D75101802A2E493600B8FA48 /* ShowLabelsOnLayerView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ShowLabelsOnLayerView.swift; sourceTree = "<group>"; };
 		D751018D2A2E962D00B8FA48 /* IdentifyLayerFeaturesView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = IdentifyLayerFeaturesView.swift; sourceTree = "<group>"; };
 		D752D9452A3A6F7F003EB25E /* MonitorChangesToMapLoadStatusView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MonitorChangesToMapLoadStatusView.swift; sourceTree = "<group>"; };
+		D752D9512A3BB309003EB25E /* arcgis-maps-sdk-swift-toolkit */ = {isa = PBXFileReference; lastKnownFileType = wrapper; name = "arcgis-maps-sdk-swift-toolkit"; path = "../arcgis-maps-sdk-swift-toolkit"; sourceTree = "<group>"; };
```
yo1995, line 374: Please remove the local toolkit reference from the project file.

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
```diff
@@ -0,0 +1,183 @@
...
+    static let startMessage = "Tap on the map to set the start location."
+    
+    /// The user instruction message for setting the end location.
+    static let endMessage = "Tap and drag on the map to set the end location."
+}
+
+private extension URL {
+    /// A elevation image service URL for Brest, France.
+    static var brestElevationService: URL {
+        URL(string: "https://scene.arcgis.com/arcgis/rest/services/BREST_DTM_1M/ImageServer")!
```
yo1995, line 176: `common-samples/issues/3852`. If you like you can lend a help hand on figuring out why the data is down and help to replace affected designs. 🙂

Shared/Samples/Measure distance in scene/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Relevant API
+
+* AnalysisOverlay
+* LocationDistanceMeasurement
+* MeasurementChangedEvent
```
yo1995, line 26: To add to Phil's comment, `MeasurementChangedEvent` is an internal symbol (LocationDistanceMeasurementMeasurementChangedEvent) so please remove it.

Shared/Samples/Measure distance in scene/measure-distance-in-scene.png
pgruenler, line 1: The latest screen shot has all zeros for the distances; I think it should show the correct distances. Is that a bug that the distances are wrong?

CalebRas: No, sorry. I must have tapped it.

pgruenler (dismissed):

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
```diff
@@ -0,0 +1,183 @@
...
+        )
+        
+        /// A measurement formatter for converting the distances to strings.
+        let measurementFormatter: MeasurementFormatter = {
+            let measurementFormatter = MeasurementFormatter()
+            measurementFormatter.unitOptions = .providedUnit
+            measurementFormatter.numberFormatter.minimumFractionDigits = 2
+            measurementFormatter.numberFormatter.maximumFractionDigits = 2
+            return measurementFormatter
+        }()
```
philium, line 148: Formatter objects [should not be used](https://developer.apple.com/documentation/foundation/data_formatting) in Swift:

> When working in Swift, use `formatted` methods directly on the types you want to format, optionally using [FormatStyle](https://developer.apple.com/documentation/foundation/formatstyle) and its subtypes to customize formatter output. This approach supports dates, integers, floating-point numbers, measurements, sequences, and person name components. Foundation caches identically-configured formatter instances internally, allowing you to focus on your app’s formatting needs.

Also, localized strings/values should not be stored. They should be computed on-demand when needed since (parts of) the locale can change while the app is running, affecting how the information is displayed. In this case, the measurements themselves should be stored and then formatted on demand like so:

```swift
Text("Direct: \(directDistance, format: ...)")
...
```

As for the format, one way you can reuse the same format for each measurement is by extending format style:

```swift
extension FormatStyle where Self == Measurement<UnitLength>.FormatStyle {
    static var distance: Self {
        .measurement(width: .abbreviated, usage: .asProvided, numberFormatStyle: .number.precision(.fractionLength(2)))
    }
}
```

And then passing `.distance` as the format: `format: .distance`.

CalebRas: @philium I updated the sample to use the format style, but can't get it work by storing the measurements. In using the following: 
```
for await measurements in model.locationDistanceMeasurement.measurements {
    directDistance = measurements.directDistance
    horizontalDistance = measurements.horizontalDistance
    verticalDistance = measurements.verticalDistance
}
```
The measurements updates come through the async stream just fine but in set them to the corresponding State variable, they wouldn't always update with the correct units when the unit system is toggled. Also, Changing the location distance measurement can cause some of distances to update when toggling the unit system but not all them. The aysnc stream and the API seem to be working fine but using `Measurement` State variables seems to be super buggy for some reason. Both Ting and looked at it and were at a loss on how to fix it so I just reverted back to using the localized strings. Let me know if there is better solution.

philium: That is surprising. I can't think of a better option at the moment, so what you have will have to do for now. Thank you for making those changes.

Samples.xcodeproj/project.pbxproj
```diff
@@ -379,6 +382,8 @@
 		D75362D12A1E886700D83028 /* ApplyUniqueValueRendererView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ApplyUniqueValueRendererView.swift; sourceTree = "<group>"; };
 		D754E3222A1D66820006C5F1 /* StylePointWithPictureMarkerSymbolsView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = StylePointWithPictureMarkerSymbolsView.swift; sourceTree = "<group>"; };
 		D78666AC2A2161F100C60110 /* FindNearestVertexView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = FindNearestVertexView.swift; sourceTree = "<group>"; };
+		D7ABA2F82A32579C0021822B /* MeasureDistanceInSceneView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MeasureDistanceInSceneView.swift; sourceTree = "<group>"; };
+		D7B3C5BF2A43A336001DA4D8 /* arcgis-maps-sdk-swift-toolkit */ = {isa = PBXFileReference; lastKnownFileType = wrapper; name = "arcgis-maps-sdk-swift-toolkit"; path = "../arcgis-maps-sdk-swift-toolkit"; sourceTree = "<group>"; };
```
yo1995, line 386: Oops, please remove the local toolkit reference.

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
```diff
@@ -0,0 +1,175 @@
...
+    /// The overlay instruction message text.
+    @State private var instructionText: String = .startMessage
+    
+    var body: some View {
+        VStack {
+            SceneViewReader { sceneViewProxy in
+                SceneView(scene: model.scene, analysisOverlays: [model.analysisOverlay])
+                    .onSingleTapGesture { screenPoint, _ in
+                        // Set the start and end locations when the screen is tapped.
+                        Task {
```
yo1995, line 43: > Can you let me know if you think there is a better way to use the task blocks

I tried a few ideas but since this sample supports both drag and single tap gestures, I can't find a good way to differentiate one from another without adding a few states. 

Thus I think in-situ task blocks are good.

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
```diff
@@ -0,0 +1,175 @@
...
+                                instructionText = .startMessage
+                            }
+                            model.locationDistanceMeasurement.endLocation = location
+                        }
+                    }
+                    .onDragGesture { _, _ in
+                        if model.locationDistanceMeasurement.startLocation == model.locationDistanceMeasurement.endLocation {
+                            return true
+                        }
+                        return false
```
yo1995, line 61:
```suggestion
                        if model.locationDistanceMeasurement.startLocation == model.locationDistanceMeasurement.endLocation {
                            return true
                        } else {
                            return false
                        }
```

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
```diff
@@ -0,0 +1,175 @@
...
+        )
+        
+        init() {
+            // Set scene to the viewpoint specified by the location distance measurement.
+            let lookAtPoint = Envelope(
+                min: locationDistanceMeasurement.startLocation,
+                max: locationDistanceMeasurement.endLocation
+            ).center
+            let camera = Camera(lookingAt: lookAtPoint, distance: 200, heading: 0, pitch: 45, roll: 0)
+            scene.initialViewpoint = Viewpoint(boundingGeometry: lookAtPoint, camera: camera)
```
yo1995, line 142:
```suggestion
            scene.initialViewpoint = Viewpoint(latitude: .nan, longitude: .nan, scale: .nan, camera: camera)
```
I don't like a point being the bounding geometry of a map or scene view as "bounding" sounds like an area or polygon to me. I like to explicitly set the unrelated parameter to `nan` to emphasize that a scene viewpoint can be solely determined by the camera, but it is up to you.

yo1995 (commented):
Generally looking good!

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
philium: Or just `return model.locationDistanceMeasurement.startLocation == model.locationDistanceMeasurement.endLocation`.

Samples.xcodeproj/project.pbxproj
CalebRas: Let me know if I didn't do it right.

pgruenler (approved):

yo1995 (approved):
Thanks for writing this sample! 

@Esri/swift-samples-reviewers After this PR is merged, Samples `v.next` depends on 200.2+ version.

Shared/Samples/Measure distance in scene/MeasureDistanceInSceneView.swift
yo1995: Phil's suggestion is good, let's apply that and add a comment to explain when the drag gesture is active.

yo1995 (dismissed):

pgruenler (approved):

yo1995 (approved):

PR #199, opened 2023-06-08
[New] Display content of utility network container

yo1995:
## Description

This PR implements `Display content of utility network container` in `Utility network` category.

URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Ting/UNContainer/Shared/Samples/Display%20content%20of%20utility%20network%20container/README.md

## Linked Issue(s)

- `swift/issues/4066`

## How To Test

1. Open the sample. Try a few times. If it gives a network error more than once, please add a comment. edit: this is a timing bug I already fixed, due to viewDidLoad vs viewDidAppear order.
2. Tap to show the legend list.
3. Tap a few features in the map. Find at least 1 with multiple contained elements, and another 1 with no containment, only itself.
4. Check memory usage, see if the model is correctly released.
5. I reworded the README a bit. Let me know if you want me to adjust the order in `How it works` section.

## To Discuss

- [ ] Core panic `Thread 1: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)`

PR #205, opened 2023-06-16
[New] Set visibility of subtype feature layer

des12437:
## Description

This PR implements `Set visibility of subtype sublayer` in `Layers` category.
URL to README: [URL](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/des12437/set-visibility-sample/Shared/Samples/Set%20visibility%20of%20subtype%20sublayer)

## Linked Issue(s)

- `swift/issues/4077`

## How To Test
- Toggle sublayer visibility with first switch.
- Toggle sublayer's original and alternate renderer with second switch.
- Tap the "Set Current to Minimum Scale" button to set the sublayer's minimum scale to the current map scale.

## Screenshots
set-visibility-of-subtype-feature-layer

## To Discuss

The information button does not show the readme and source code. The view shows for a second and then dismisses with this error:
```
Samples[87547:5050058] [assertion] Error acquiring assertion: <Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}>
```

https://github.com/Esri/arcgis-maps-sdk-swift-samples/assets/117859673/75acc93b-9f7f-4cf8-95f4-397c2d239a24


yo1995:
Please follow the latest common design to rename the sample to `Set visibility of subtype sublayer`

Samples.xcodeproj/project.pbxproj
```diff
@@ -698,12 +711,12 @@
...
 			children = (
-				0074ABAF281742420037244A /* Supporting Files */,
+				001C6DD827FE585A00D472C2 /* AppSecrets.swift.masque */,
+				00E5400E27F3CCA200CF66D5 /* Assets.xcassets */,
 				0074ABB228174B830037244A /* Samples */,
 				00E5400C27F3CCA100CF66D5 /* SamplesApp.swift */,
-				00E5400E27F3CCA200CF66D5 /* Assets.xcassets */,
-				001C6DD827FE585A00D472C2 /* AppSecrets.swift.masque */,
 				0074ABCA2817B8DB0037244A /* SamplesApp+Samples.swift.tache */,
+				0074ABAF281742420037244A /* Supporting Files */,
```
yo1995, line 719: No need to sort the whole folder, only the "Samples" group is fine. 😬

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+    /// The view model for this sample.
+    class Model: ObservableObject {
+        /// A map with a streets basemap style.
+        let map = Map(basemapStyle: .arcGISStreetsNight)
+        
+        /// A feature table for the electrical network in this sample.
+        private let featureTable = ServiceFeatureTable(url: .featureServiceURL)
+        
+        /// A subtype feature layer created from the service feature table.
+        private var subtypeFeatureLayer: SubtypeFeatureLayer!
```
pgruenler, line 28: Are there guidelines on when to use implicitly unwrapped optionals in the samples app code? I would rather the 3 iuo's here be optionals since they don't get their values in the init. However, if it's common practice to use iuo's like this then that's okay.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+        
+        /// The subtype sublayer of the subtype feature layer in this sample.
+        private var subtypeSublayer: SubtypeSublayer!
```
rolson, line 31: This should be optional for the same reason mentioned about `originalRenderer`.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.swift
```diff
@@ -0,0 +1,50 @@
...
+                        model.isShowingSettings.toggle()
+                    }
+                    .sheet(isPresented: $model.isShowingSettings, detents: [.medium], dragIndicatorVisibility: .visible) {
+                        SettingsView()
+                            .environmentObject(model)
+                    }
+                }
+            }
+            .task {
+                try? await model.setup()
```
rolson, line 47: What if this fails? What will the user see? What will the values of the IUO variables be?

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        
+        deinit {
+            ArcGISEnvironment.authenticationManager.arcGISCredentialStore.removeAll()
+        }
+        
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        func setup() async throws {
+            try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+            guard isAuthenticated else { return }
```
rolson, line 96: this just stops the setup with no warning or error. Is that what we want?  Should this be a precondition?

rolson: You actually do'nt need this because if it fails to add on the line above then the method throws and that is propagated and the method exits.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.swift
des12437: Fixed. An error message will be presented to the user if the setup fails.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+        private let featureTable = ServiceFeatureTable(url: .featureServiceURL)
+        
+        /// A subtype feature layer created from the service feature table.
+        private var subtypeFeatureLayer: SubtypeFeatureLayer!
+        
+        /// The subtype sublayer of the subtype feature layer in this sample.
+        private var subtypeSublayer: SubtypeSublayer!
+        
+        /// The renderer of the subtype feature layer.
+        private var originalRenderer: Renderer!
```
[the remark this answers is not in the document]
des12437: These variables can't be declared as constants and set in the init since their initialization requires a layer to be loaded asynchronously, so they are set in the setup function instead.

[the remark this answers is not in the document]
rolson: No, but they should be optional.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,171 @@
...
+        /// The renderer of the subtype feature layer.
+        private var originalRenderer: Renderer?
+        
+        /// The status text to display to the user.
+        @Published var statusText = ""
+        
+        /// A Boolean value indicating if the sample is authenticated.
+        private var isAuthenticated: Bool {
+            !ArcGISEnvironment.authenticationManager.arcGISCredentialStore.credentials.isEmpty
+        }
```
rolson, line 42: can this be removed now that it's not used?

des12437: Yes.

pgruenler (approved):

philium:
Is this ready for review? I was waiting for a review request after the latest changes were pushed, but @pgruenler went ahead and reviewed it, so I will too.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,166 @@
...
+        /// The renderer of the subtype feature layer.
+        private(set) var originalRenderer: Renderer?
+        
+        /// The status text to display to the user.
+        @Published private(set) var statusText = ""
+        
+        /// The  subtype sublayer's label definition.
+        private(set) var labelDefinition: LabelDefinition = {
+            // Make and stylize the text symbol.
+            let textSymbol = TextSymbol()
```
yo1995, line 42:
```suggestion
            let textSymbol = TextSymbol(color: .blue, size: 10.5)
```
Moving some parameters to the initializer. Maybe this is an overkill, up to you.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,166 @@
...
+        @Published var isShowingSettings = false
+        
+        /// A Boolean value indicating whether to show the subtype sublayer.
+        @Published var showsSublayer = true
+        
+        /// A Boolean value indicating whether to show the subtype sublayer's renderer.
+        @Published var showsOriginalRenderer = true
+        
+        /// The current scale of the map.
+        @Published var currentScale: Double = .zero
```
yo1995, line 75: It seems the view doesn't depend on this value, so no need to be published? However if the format style suggestion is applied, then the value can be formatted in place.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,166 @@
...
+        }()
+        
+        /// A Boolean value indicating whether the settings should be presented.
+        @Published var isShowingSettings = false
+        
+        /// A Boolean value indicating whether to show the subtype sublayer.
+        @Published var showsSublayer = true
+        
+        /// A Boolean value indicating whether to show the subtype sublayer's renderer.
+        @Published var showsOriginalRenderer = true
```
yo1995, line 72: Ideally they should be state properties on the view.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,166 @@
...
+        
+        /// The current scale of the map.
+        @Published var currentScale: Double = .zero
+        
+        /// The map's current scale value in text.
+        @Published var currentScaleText: String = ""
+        
+        /// The subtype sublayer's minimum scale value in text.
+        @Published var minimumScaleText: String = "Not Set"
```
yo1995, line 81:
```suggestion
        @Published private(set) var minimumScaleText: String = "Not Set"
```

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,166 @@
...
+        }
+        
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        @MainActor
+        func setup() async {
+            do {
+                try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+                try await subtypeFeatureLayer.load()
+                map.addOperationalLayer(subtypeFeatureLayer)
+                subtypeSublayer = subtypeFeatureLayer.sublayer(withSubtypeName: "Street Light")
```
yo1995, line 100: This can be guarded, as it is optional chained 3 times below. Implicitly unwrapped optional or not is up to you.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Views.swift
```diff
@@ -0,0 +1,55 @@
...
+                }
+                Section("Sublayer Minimum Scale") {
+                    VStack {
+                        HStack {
+                            Text("Minimum Scale")
+                            Spacer()
+                            Text(model.minimumScaleText)
+                                .foregroundColor(.secondary)
+                        }
+                    }
```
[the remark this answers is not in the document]
yo1995: Scratch that… iOS 16+ only 💔

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
des12437: With `guard` the return will execute and then the subtype sublayer's `isVisible` and `renderer` properties can't be updated. It seems unsafe to force unwrap.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Views.swift
```diff
@@ -0,0 +1,55 @@
...
+                Section("Layers") {
+                    Toggle("Show Sublayer", isOn: $model.showsSublayer)
+                        .onChange(of: model.showsSublayer) { _ in
+                            model.toggleSublayer()
+                        }
+                    Toggle("Show Original Renderer", isOn: $model.showsOriginalRenderer)
+                        .onChange(of: model.showsOriginalRenderer) { _ in
+                            model.toggleRenderer()
+                        }
+                }
```
yo1995, line 34: Sorry, I still want to put the booleans on the view. Such as below. Feel free to use a better method signature.

```swift
        /// A Boolean value indicating whether to show the subtype sublayer.
        @State private var showsSublayer = true
        
        /// A Boolean value indicating whether to show the subtype sublayer's renderer.
        @State private var showsOriginalRenderer = true
        
        var body: some View {
            Form {
                Section("Layers") {
                    Toggle("Show Sublayer", isOn: $showsSublayer)
                        .onChange(of: showsSublayer) { newValue in
                            model.toggleSublayer(isVisible: newValue)
                        }
                    Toggle("Show Original Renderer", isOn: $showsOriginalRenderer)
                        .onChange(of: showsOriginalRenderer) { newValue in
                            model.toggleRenderer(showsOriginalRenderer: newValue)
                        }
                }
```

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.swift
```diff
@@ -0,0 +1,63 @@
...
+                    .padding(8)
+                    .background(.ultraThinMaterial, ignoresSafeAreaEdges: .horizontal)
+                    .multilineTextAlignment(.center)
+            }
+            .toolbar {
+                ToolbarItemGroup(placement: .bottomBar) {
+                    Button("Settings") {
+                        model.isShowingSettings.toggle()
+                    }
+                    .sheet(isPresented: $model.isShowingSettings, detents: [.medium], dragIndicatorVisibility: .visible) {
```
yo1995, line 48: Please put the boolean in this view.

```swift
    /// A Boolean value indicating whether the settings should be presented.
    @State private var isShowingSettings = false
```

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+                subtypeSublayer?.renderer = originalRenderer
+            } else {
+                let symbol = SimpleMarkerSymbol(style: .diamond, color: .systemPink, size: 20)
+                let alternativeRenderer = SimpleRenderer(symbol: symbol)
+                subtypeSublayer?.renderer = alternativeRenderer
+            }
+        }
+        
+        func formatCurrentScaleText() {
+            currentScaleText = "1:\(currentScale.formatted(.decimal))"
```
yo1995, line 115: no need to change, another way can be `1:\(currentScale, specifier: "%.0f")`, although it has a comma separator.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+        @Published var showsOriginalRenderer = true
+        
+        /// The current scale of the map.
+        var currentScale: Double = .zero
+        
+        /// The map's current scale value in text.
+        @Published var currentScaleText: String = ""
+        
+        /// The subtype sublayer's minimum scale value in text.
+        @Published private(set) var minimumScaleText: String = "Not Set"
```
yo1995, line 76: Instead of holding the scale texts, let's generate the text in the views.

The `currentScale` of the map view can be a state to the main view, and the `minScale` in the settings row can be computed from `subtypeSublayer?.minScale`.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,164 @@
...
+            subtypeFeatureLayer = SubtypeFeatureLayer(featureTable: featureTable)
+            subtypeFeatureLayer.scalesSymbols = false
+        }
+        
+        deinit {
+            ArcGISEnvironment.authenticationManager.arcGISCredentialStore.removeAll()
+        }
+        
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        @MainActor
```
yo1995, line 89: Phil [suggested](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/199#discussion_r1232571109) that the observable objects should have the `@MainActor` attribute added to the type, instead of certain methods, according to Apple.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
des12437: I did not apply this suggestion since the scale information is needed on the view and SettingsView, in which case the data is more easily accessible on the model.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,147 @@
...
+        let map = Map(basemapStyle: .arcGISStreetsNight)
+        
+        /// A feature table for the electrical network in this sample.
+        private let featureTable = ServiceFeatureTable(url: .featureServiceURL)
+        
+        /// A subtype feature layer created from the service feature table.
+        private let subtypeFeatureLayer: SubtypeFeatureLayer
+        
+        /// The subtype sublayer of the subtype feature layer in this sample.
+        var subtypeSublayer: SubtypeSublayer?
```
yo1995, line 32:
```suggestion
        private var subtypeSublayer: SubtypeSublayer?
```

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
yo1995: OK

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
yo1995: Hmm I don't quite get it. If `subtypeSublayer` is `nil`, then all the optional chainings will stop at the `nil`. It seems to me that all the settings UI should be disabled if it is `nil`.

des12437:
> Hmm I don't quite get it. If subtypeSublayer is nil, then all the optional chainings will stop at the nil. It seems to me that all the settings UI should be disabled if it is nil.

I'm not sure why, but unwrapping the subtypeSublayer is not successful so `subtypeSublayer.labelsAreEnabled` and the renderer are not set in the init which causes unexpected behavior in the sample. The current implementation causes the values to be set correctly which is odd.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
des12437: Okay I unwrapped the layer and also made it an implicitly unwrapped optional which will call `fatalError()` if not initialized. I can make the layer optional again if preferred.

yo1995: Thank you. I'm 50/50 on whether or not to use IUO. The reasons that it may become `nil` as I can think of are

- `subtypeFeatureLayer` is not loaded due to auth or network issues. `fatalError` might not be the best here. However, I have good faith in sampleserver7's stability.
- `sublayer(withSubtypeName:)` method returns `nil` due to a programmer error of using a wrong name. `fatalError` in this case makes sense.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Views.swift
```diff
@@ -0,0 +1,58 @@
...
+extension SetVisibilityOfSubtypeSublayerView {
+    struct SettingsView: View {
+        /// The view model for the sample.
+        @ObservedObject var model: Model
+        
+        /// A Boolean value indicating whether to show the subtype sublayer.
+        @State var showsSublayer = true
+        
+        /// A Boolean value indicating whether to show the subtype sublayer's renderer.
+        @State var showsOriginalRenderer = true
```
yo1995, line 27:
```suggestion
        /// A Boolean value indicating whether to show the subtype sublayer.
        @State private var showsSublayer = true
        
        /// A Boolean value indicating whether to show the subtype sublayer's renderer.
        @State private var showsOriginalRenderer = true
```

yo1995 (approved):

yo1995 (dismissed):

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
rolson: What is the variable in question about the IUO?  If it is [this variable](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/205/files#diff-ab27dee8370a163cb44caef7203e764ac165300ec47dcb7ad31b1fbcc00b427eR32), then I don't think it should be IUO.

This line could throw:
```
try await subtypeFeatureLayer.load()
```

and then the variable `subtypeSublayer` would be `nil` and unsafe.  It needs to be an optional.

des12437: Fixed.

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,148 @@
...
+        deinit {
+            ArcGISEnvironment.authenticationManager.arcGISCredentialStore.removeAll()
+        }
+        
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        func setup() async throws {
+            try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+            try await subtypeFeatureLayer.load()
+            map.addOperationalLayer(subtypeFeatureLayer)
+            guard let subtypeSublayer = subtypeFeatureLayer.sublayer(withSubtypeName: "Street Light") else { fatalError("Failed to initialize subtype sublayer") }
```
rolson, line 77: Now that subtypeSublayer is optional, you don't need to crash with a fatalError here. The question is should you return or throw an error?  If you simply return in the case where the sublayer cannot be found then `setup()` basically fails and gives no indication of the failure. That means the code that calls `setup()` won't catch an error and it will think that setup succeeded when it didn't.

What I think should be done is something like this:

```swift
            guard let subtypeSublayer = subtypeFeatureLayer.sublayer(withSubtypeName: "Street Light") else {
                throw SetupError.cannotFindSublayer
            }
```

where you have an error defined something like this:

```swift
extension SetVisibilityOfSubtypeSublayerView.Model {
    enum SetupError: Error {
        case cannotFindSublayer
    }
}
```

You should test that to see what it shows the user in the case of the error here:
```
            .alert(isPresented: $isShowingAlert, presentingError: error)
```

If it doesn't show a nice message the error may need to conform to something like LocalizedError.

des12437: Here is the presented error message:

rolson: Try conforming to LocalizedError and see if that presents a nice error message.

des12437: Yes that works

Shared/Samples/Set visibility of subtype sublayer/SetVisibilityOfSubtypeSublayerView.Model.swift
```diff
@@ -0,0 +1,168 @@
...
+    var errorDescription: String? {
+        switch self {
+        case .cannotFindSublayer:
+            return NSLocalizedString(
+                "Cannot find subtype sublayer.",
+                comment: "Error thrown when subtype sublayer cannot be found."
+            )
+        }
+    }
+}
```
yo1995, line 127:
```suggestion
extension SetVisibilityOfSubtypeSublayerView.Model {
    enum SetupError: LocalizedError {
        case cannotFindSublayer
        
        var errorDescription: String? {
            switch self {
            case .cannotFindSublayer:
                return NSLocalizedString(
                    "Cannot find subtype sublayer.",
                    comment: "Error thrown when subtype sublayer cannot be found."
                )
            }
        }
    }
}
```
Should it be simplified like this?

yo1995 (approved):

pgruenler (approved):

des12437:
Thank you both @yo1995 @pgruenler

Samples.xcodeproj/project.pbxproj
```diff
@@ -1383,11 +1407,13 @@
 				D71099702A2802FA0065A1C1 /* DensifyAndGeneralizeGeometryView.SettingsView.swift in Sources */,
 				E004A6ED2849556E002A1FE6 /* CreatePlanarAndGeodeticBuffersView.swift in Sources */,
 				E041ABD7287DB04D0056009B /* SampleInfoView.swift in Sources */,
+				1C43BC822A43781200509BF8 /* SetVisibilityOfSubtypeSublayerView.Model.swift in Sources */,
 				D752D9462A3A6F80003EB25E /* MonitorChangesToMapLoadStatusView.swift in Sources */,
 				E0082217287755AC002AD138 /* View+Sheet.swift in Sources */,
 				00181B462846AD7100654571 /* View+Alert.swift in Sources */,
 				E0FE32E728747778002C6ACA /* BrowseBuildingFloorsView.swift in Sources */,
 				D752D95F2A3BCE06003EB25E /* DisplayMapFromPortalItemView.swift in Sources */,
+				1CAB8D072A3CCBD8002AA649 /* SetVisibilityOfSubtypeFeatureLayerView.Views.swift in Sources */,
```
yo1995, line 1416: It seems `SetVisibilityOfSubtypeFeatureLayerView.Views.swift` was added twice to the project. When the project does a self-clean it will be removed. cc @CalebRas

PR #213, opened 2023-06-27
[Update] Refactor `CreateBuffersAroundPointsView.Model`

### LINKED ISSUE #212: Refactor Create buffers around points sample
Get rid of the suboptimal usage of published properties in the model.

yo1995:
## Description

Close #212. This PR updates `Create buffers around points` added in #185 .

## Linked Issue(s)

- #212 

## How To Test

The sample should behave the same as before.

## To Discuss

In hindsight I should have paid more attention to the code review. This sample was one of Caleb's first samples so I let it go without being too harsh.

Ideally we should have a more careful mindset on how to use the data model object. Try to avoid the case where the model knows about states in the UI, or in other words, UI states should be derived from the model, not reverse.

### DIFF
--- a/Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
+++ b/Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
@@ -39,12 +39,12 @@ struct CreateBuffersAroundPointsView: View {
         MapView(map: model.map, graphicsOverlays: model.graphicsOverlays)
             .onSingleTapGesture { _, mapPoint in
                 // Update tapPoint and bring up input box if point is within bounds.
-                guard model.boundaryContains(mapPoint) else {
+                if model.boundaryContains(mapPoint) {
+                    tapPoint = mapPoint
+                    inputBoxIsPresented = true
+                } else {
                     status = .outOfBoundsTap
-                    return
                 }
-                tapPoint = mapPoint
-                inputBoxIsPresented = true
             }
             .overlay(alignment: .top) {
                 Text(status.label)
@@ -58,8 +58,9 @@ struct CreateBuffersAroundPointsView: View {
                     Toggle("Union", isOn: $shouldUnion)
                         .toggleStyle(.switch)
                         .onChange(of: shouldUnion) { _ in
-                            guard !model.bufferPoints.isEmpty else { return }
-                            model.drawBuffers(shouldUnion: shouldUnion)
+                            if !model.bufferPoints.isEmpty {
+                                model.drawBuffers(shouldUnion: shouldUnion)
+                            }
                         }
                     Button("Clear") {
                         model.clearBufferPoints()
@@ -72,23 +73,22 @@ struct CreateBuffersAroundPointsView: View {
                 TextField("100", text: $radiusInput)
                     .keyboardType(.numberPad)
                 Button("Done") {
-                    defer { radiusInput.removeAll() }
-                    
-                    // Ensure the tapPoint is within the boundary.
+                    // Check to ensure the tapPoint is within the boundary.
                     guard let tapPoint, model.boundaryContains(tapPoint) else {
                         status = .outOfBoundsTap
                         return
                     }
                     
                     // Ensure that the input is valid.
-                    guard let radius = Double(radiusInput),
-                          radius > 0 && radius < 300 else {
+                    if let radius = Double(radiusInput),
+                       radius > 0 && radius < 300 {
+                        model.addBuffer(point: tapPoint, radius: radius)
+                        model.drawBuffers(shouldUnion: shouldUnion)
+                        status = .bufferCreated
+                    } else {
                         status = .invalidInput
-                        return
                     }
-                    model.addBuffer(point: tapPoint, radius: radius)
-                    model.drawBuffers(shouldUnion: shouldUnion)
-                    status = .bufferCreated
+                    radiusInput.removeAll()
                 }
                 Button("Cancel") {
                     radiusInput.removeAll()
--- a/Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
+++ b/Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
@@ -73,22 +73,27 @@ struct CreateBuffersAroundPointsView: View {
                 TextField("100", text: $radiusInput)
                     .keyboardType(.numberPad)
                 Button("Done") {
+                    defer { radiusInput.removeAll() }
+                    guard let tapPoint else {
+                        preconditionFailure("Missing tap point")
+                    }
+
+                    let newStatus: Status
                     // Check to ensure the tapPoint is within the boundary.
-                    guard let tapPoint, model.boundaryContains(tapPoint) else {
-                        status = .outOfBoundsTap
-                        return
-                    }
-                    
-                    // Ensure that the input is valid.
-                    if let radius = Double(radiusInput),
-                       radius > 0 && radius < 300 {
-                        model.addBuffer(point: tapPoint, radius: radius)
-                        model.drawBuffers(shouldUnion: shouldUnion)
-                        status = .bufferCreated
+                    if model.boundaryContains(tapPoint) {
+                        // Ensure that the input is valid.
+                        if let radius = Double(radiusInput),
+                           radius > 0 && radius < 300 {
+                            model.addBuffer(point: tapPoint, radius: radius)
+                            model.drawBuffers(shouldUnion: shouldUnion)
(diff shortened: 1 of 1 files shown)


Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -43,11 +55,11 @@ struct CreateBuffersAroundPointsView: View {
             .toolbar {
                 ToolbarItemGroup(placement: .bottomBar) {
                     // Union toggle switch.
-                    Toggle("Union", isOn: $model.shouldUnion)
+                    Toggle("Union", isOn: $shouldUnion)
                         .toggleStyle(.switch)
-                        .onChange(of: model.shouldUnion) { _ in
+                        .onChange(of: shouldUnion) { _ in
                             if !model.bufferPoints.isEmpty {
```
des12437, line 61: This could be a guard statement for readability.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -43,11 +55,11 @@ struct CreateBuffersAroundPointsView: View {
...
-                                model.drawBuffers()
+                                model.drawBuffers(shouldUnion: shouldUnion)
                             }
                         }
```
philium: The action closure should use the value passed to it. Also, not a big deal, but a guard statement could be used.

```swift
.onChange(of: shouldUnion) { newShouldUnion in
    guard !model.bufferPoints.isEmpty else { return }
    model.drawBuffers(shouldUnion: newShouldUnion)
}
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -57,29 +69,29 @@ struct CreateBuffersAroundPointsView: View {
                     .disabled(model.bufferPoints.isEmpty)
                 }
             }
-            .alert("Buffer Radius", isPresented: $model.inputBoxIsPresented, actions: {
-                TextField("100", text: $model.radiusInput)
+            .alert("Buffer Radius", isPresented: $inputBoxIsPresented, actions: {
+                TextField("100", text: $radiusInput)
```
philium, line 72: Manually converting to/from string and double is tedious and error prone. Thankfully, SwiftUI has our back. Instead of storing the user input, store the buffer radius:

```swift
@State private var bufferRadius: Double = 100
```

And then pass that when creating the `Text` instance:

```swift
TextField("radius in miles", value: $bufferRadius, format: .number)
```

des12437 (approved):

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -57,29 +69,29 @@ struct CreateBuffersAroundPointsView: View {
...
+                        model.addBuffer(point: tapPoint, radius: radius)
+                        model.drawBuffers(shouldUnion: shouldUnion)
                         status = .bufferCreated
                     } else {
                         status = .invalidInput
                     }
-                    model.radiusInput = ""
+                    radiusInput.removeAll()
                 }
                 Button("Cancel") {
```
philium, line 93: Setting a role here allows the return key to activate the Done button.

```suggestion
                Button("Cancel", role: .cancel) {
```

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -213,8 +213,9 @@ private extension CreateBuffersAroundPointsView {
             return true
         }
         
-        /// Draw points and their buffers on the map.
-        func drawBuffers() {
+        /// Draws points and their buffers on the map.
+        /// - Parameter shouldUnion: A Boolean indicating whether the buffers should union.
+        func drawBuffers(shouldUnion: Bool) {
```
philium, line 218: An alternative parameter name is `unioned` (`drawBuffers(unioned:)`.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -241,19 +242,22 @@ private extension CreateBuffersAroundPointsView {
             }
         }
         
-        /// Add a point with its radius to the bufferPoints array.
-        func addBufferPoint() {
+        /// Adds a point with its radius to the buffer points array.
+        /// - Parameters:
+        ///   - point: The center point to create a buffer.
+        ///   - radius: The radius of the buffer.
+        func addBuffer(point: Point, radius: Double) {
```
philium, line 249: Should `point` have an argument label? I'm thinking `at` makes sense (`addBuffer(at:radius:)`).

yo1995: The purpose of the method is to add a `(point, radius)` tuple to an array rather than to add a graphical buffer at the point, so maybe `addBufferPoint(point:radius:)`?

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -57,29 +68,30 @@ struct CreateBuffersAroundPointsView: View {
...
-                    } else {
+                    guard let radius = Double(radiusInput),
+                          radius > 0 && radius < 300 else {
                         status = .invalidInput
+                        return
                     }
-                    model.radiusInput = ""
+                    model.addBuffer(point: tapPoint, radius: radius)
+                    model.drawBuffers(shouldUnion: shouldUnion)
+                    status = .bufferCreated
```
philium, line 91: One other thing. The else clause of guard statements wasn't intended for doing additional work. In my experience, doing so _always_ harms readability. It breaks the mental model of "forward progress is halted". Here, an initial guard statement followed by an if statement would make this closure much more readable:

```swift
guard let tapPoint else {
    preconditionFailure("Missing tap point")
}

let newStatus: Status
if model.boundaryContains(tapPoint) {
    if let radius = Double(radiusInput),
       radius > 0 && radius < 300 {
        model.addBuffer(point: tapPoint, radius: radius)
        model.drawBuffers(shouldUnion: shouldUnion)
        newStatus = .bufferCreated
    } else {
        newStatus = .invalidInput
    }
} else {
    newStatus = .outOfBoundsTap
}
status = newStatus
```

yo1995: I agree, this is a good point and I've had put it in the style guide before.

Would you say that going on we should limit the `else` branch to only 1 line, that can be one of the following

- return/return nil/return sth empty or default
- throw
- fatalError
-

philium: A guard's else clause can also continue a loop. But that is all part of the documentation for guard statements, inherent to how they function. I would just say that a guard's else clause shouldn't perform any work other than what is needed. Its job is to get out of the current scope and it should do so as quickly as possible. By "other work", I mean work with consequences. There are legitimate use cases for having additional code in a guard's else clause, such as logging. But even that, in my experience, is usually just temporary for debugging.

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
yo1995: > If the user stops editing the text in an invalid state, the text field updates the field’s text to the last known valid value. [doc](https://developer.apple.com/documentation/swiftui/textfield/init(value:format:prompt:label:)-99ntf)

This is an interesting behavior. If the default value is 100 and I tap the backspace 3 times, the resulting value is 1. Which might be surprising to a user who expect an invalid input. But it is not a big deal.

des12437 (approved):

Shared/Samples/Create buffers around points/CreateBuffersAroundPointsView.swift
```diff
@@ -213,8 +217,9 @@ private extension CreateBuffersAroundPointsView {
             return true
         }
         
-        /// Draw points and their buffers on the map.
-        func drawBuffers() {
+        /// Draws points and their buffers on the map.
+        /// - Parameter unioned: A Boolean indicating whether the buffers should union.
```
pgruenler, line 221:
```suggestion
        /// - Parameter unioned: A Boolean indicating whether the buffers should be unioned.
```

pgruenler (commented):
Looks good. Just one suggestion.

pgruenler (approved):
Meant to approve with one suggestion.

pgruenler (approved):

PR #214, opened 2023-06-27
[Update] Examine SwiftUI importation

yo1995:
## Description

This PR removes extra `import SwiftUI` for data models. Close the card https://github.com/Esri/arcgis-maps-sdk-swift-samples/projects/1#card-89559142

## Linked Issue(s)

- https://github.com/Esri/arcgis-maps-sdk-swift-samples/projects/1#card-89559142
- https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/203#discussion_r1242904166

## How To Test

All affected samples still compile.

### DIFF
--- a/Shared/Samples/Change map view background/ChangeMapViewBackgroundView.Model.swift
+++ b/Shared/Samples/Change map view background/ChangeMapViewBackgroundView.Model.swift
@@ -12,9 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Foundation
-import SwiftUI
 import ArcGIS
+import SwiftUI
 
 extension ChangeMapViewBackgroundView {
     /// The model used to store the geo model and other expensive objects
--- a/Shared/Samples/Add dynamic entity layer/AddDynamicEntityLayerView.Model.swift
+++ b/Shared/Samples/Add dynamic entity layer/AddDynamicEntityLayerView.Model.swift
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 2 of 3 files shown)


Shared/Samples/Change map view background/ChangeMapViewBackgroundView.Model.swift
```diff
@@ -12,9 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Foundation
-import SwiftUI
 import ArcGIS
+import SwiftUI
```
yo1995, line 16: This model uses `Color` which unfortunately relies on SwiftUI. Nevertheless, since SwiftUI implicitly depend on Foundation, I removed it.

philium: While it does use `Color`, it looks like it just converts all colors to `UIColor`, so the code might be simpler if it used `UIColor` instead. That would still mean importing a UI framework (`UIKit` instead of `SwiftUI`), but that can't really be avoided in this case.

philium (approved):

Shared/Samples/Change map view background/ChangeMapViewBackgroundView.Model.swift
yo1995: Seems like the [ColorPicker](https://developer.apple.com/documentation/swiftui/colorpicker) in the other file/view which consumes these properties only supports `Color` or `CGColor` as initializing parameters, so I'll keep it unchanged for now.

vquach2404 (approved):

PR #220, opened 2023-07-05
[Fix] Sample metadata decoding script works incorrectly with empty directory

yo1995:
## Description

This PR fixes an issue in the decoder script. When Git switches branches, it sometimes leaves an empty directory behind. When the script is run, it cannot find the metadata file at the path and prints a "sample couldn’t be decoded." error.

This PR fixes it by returning `nil` for non-existing metadata.

Thanks @CalebRas for reporting this bug!

## How To Test

Create an empty folder in the `Samples` folder, and the project still builds.

### DIFF
--- a/Scripts/GenerateSampleViewSourceCode.swift
+++ b/Scripts/GenerateSampleViewSourceCode.swift
@@ -84,14 +84,14 @@ private let sampleMetadata: [SampleMetadata] = {
             .filter(\.hasDirectoryPath)
             .compactMap { url in
                 // Gets the metadata JSON under each subdirectory.
+                let metadataFile = url.appendingPathComponent("README.metadata.json")
+                guard fileManager.fileExists(atPath: metadataFile.path) else {
+                    // Sometimes when Git switches branches, it leaves an empty directory
+                    // that causes the decoder to throw an error. Here we skip the directories
+                    // that don't have the metadata JSON file.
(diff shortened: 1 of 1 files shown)


CalebRas (dismissed):
It fixes the issue and looks good to me!

Scripts/GenerateSampleViewSourceCode.swift
```diff
@@ -77,14 +77,22 @@ private let sampleMetadata: [SampleMetadata] = {
...
                 // Gets the metadata JSON under each subdirectory.
                 do {
-                    let data = try Data(contentsOf: url.appendingPathComponent("README.metadata.json"))
+                    let metadataFile = url.appendingPathComponent("README.metadata.json")
+                    guard fileManager.fileExists(atPath: metadataFile.path) else {
+                        // Sometimes when Git switches branches, it leaves an empty directory
+                        // that causes the decoder to throw an error. Here we skip the directories
+                        // that don't have the metadata JSON file.
+                        return nil
+                    }
```
philium, line 94: Could this go before the do…catch statement?

yo1995: Oops, it should!

philium (approved):

CalebRas (approved):

PR #222, opened 2023-07-07
[New] Create load report

des12437:
## Description

This PR implements `Create load report` in `Utility Networks` category.
URL to README: [URL](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/des12437/Create-load-report-sample/Shared/Samples/Create%20load%20report/README.md)

## Linked Issue(s)

- `swift/issues/4070`

## How To Test
- Add and remove phases to be included in the load report.
- Tap "Run" to initiate a downstream trace on the network and create the load report.
- Tap "Reset" to clear the phases.


Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+    public static func == (lhs: CodedValue, rhs: CodedValue) -> Bool {
+        lhs.name == rhs.name
+    }
+}
+
+extension CodedValue: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(name)
+    }
+}
```
yo1995, line 283: I'm not sure what is the best practice, but it seems to me that adding public extension to ArcGIS types would have side effects. We removed the map and graphics overlay's `ObservableObject` conformance in 200.1 and I think it is the same case here.
 
ref https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/223

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+
+extension CreateLoadReportView {
```
yo1995, line 285:
```suggestion
private extension CreateLoadReportView {
```

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+        @Published private(set) var excludedPhases = [CodedValue]()
+        
+        /// A list of phases that are included in the load report.
+        @Published private(set) var includedPhases = [CodedValue]()
+        
+        /// A list of possible phases populated from the network's attributes.
+        private var allPhases = [CodedValue]()
+        
+        /// The phase summaries in the load report.
+        @Published private(set) var summaries = [CodedValue: PhaseSummary]()
```
yo1995, line 47:
```suggestion
       private var summaries = [CodedValue: PhaseSummary]()
```
This doesn't seem to be used in the UI.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+        
+        /// The number formatter for phase summaries.
+        private let numberFormatter: NumberFormatter = {
+            let formatter = NumberFormatter()
+            formatter.numberStyle = .decimal
+            formatter.usesGroupingSeparator = true
+            return formatter
+        }()
```
yo1995, line 55: Please use format style when possible.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+        private let utilityNetwork = UtilityNetwork(url: .utilityNetwork)
+        
+        /// The initial conditional expression.
+        private var initialExpression: UtilityTraceConditionalExpression!
+        
+        /// The trace parameters for creating load reports.
+        private var traceParameters: UtilityTraceParameters!
+        
+        /// The network attributes for the comparison.
+        private var phasesNetworkAttribute: UtilityNetworkAttribute!
```
yo1995, line 34: I added a new section to the contrib guide about using IUO. In general moving on we should only use it can never lead to an unexpected crash.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+            // Create downstream trace parameters with elements and function outputs.
+            let traceParameters = UtilityTraceParameters(traceType: .downstream, startingLocations: [startingLocation])
+            traceParameters.addResultTypes([.elements, .functionOutputs])
+            
+            // The service category for counting total customers.
+            if let serviceCategory = utilityNetwork.definition?.categories.first(where: { $0.name == "ServicePoint" }),
+               // The load attribute for counting total load.
+               let loadAttribute = utilityNetwork.definition?.networkAttributes.first(where: { $0.name == "Service Load" }),
+               // The phase attribute for getting total phase current load.
+               let phasesNetworkAttribute = utilityNetwork.definition?.networkAttributes.first(where: { $0.name == "Phases Current" }) {
```
yo1995, line 101: `utilityNetwork.definition` is conditionally unwrapped three times here, so it may be a good idea to unwrap it before calling the `first`s.

Shared/Samples/Create load report/CreateLoadReportView.Views.swift
```diff
@@ -0,0 +1,63 @@
...
+
+extension CreateLoadReportView {
+    struct LoadReportView: View {
+        /// The view model for the sample.
+        @ObservedObject var model: Model
+        
+        var body: some View {
+            Form {
+                Section("Phases, Total Customers(C), Total Load(L)") {
+                    ForEach(model.includedPhases.sorted { $0.name < $1.name }, id: \.name) { phase in
```
yo1995, line 26: Since the arrays are sorted in `addPhase` and `deletePhase` methods, I think the sort here is not needed.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,299 @@
...
+                let formattedString = String(
+                    format: "C: %@  L: %@",
+                    numberFormatter.string(from: summary.totalCustomers as NSNumber)!,
+                    numberFormatter.string(from: summary.totalLoad as NSNumber)!
+                )
+                return formattedString
+            } else {
+                return nil
+            }
+        }
```
yo1995, line 245: I think `String(format:_)` is not what we should use in SwiftUI.
@philium is using interpolated string the recommended way?
Such as `"C: \(summary.totalCustomers.formatted()), L: \(summary.totalLoad.formatted())"` (with the right format style)

philium: Yes, that works. Or, if you want more control over the format, use string interpolation of the value itself and provide a format style to use. Either way, this method (and anything format-related for display to the user) should go on the view itself:

```swift
func summaryForPhase(_ phase: CodedValue) -> LocalizedStringKey {
    let format: IntegerFormatStyle = .number
    if let summary = model.summaries[phase] {
        return "C: \(summary.totalCustomers, format: format)  L: \(summary.totalLoad, format: format)"
    } else {
        return "N/A"
    }
}
```

yo1995: Thank you, I updated the samples contribution guideline with these localization considerations.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
des12437: Okay is there another way to add the conformance? I added it to create the dictionary.

Screenshot 2023-07-07 at 2 31 22 PM

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,292 @@
...
+        
+        // MARK: Methods
+        
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        func setup() async throws {
+            try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+            try await loadUtilityNetwork()
+        }
+        
+        /// Load the utility network.
```
yo1995, line 71:
```suggestion
        /// Loads the utility network.
```

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
yo1995: Let's wait for Ryan's thoughts. All I can think of are hacky ways to store the coded value with a hashable dict key. 🤕

yo1995 (commented):
Mostly looking good!

rolson (commented):
Using this sample, I don't really see the point of the Reset button. It doesn't seem like it's needed.  Just re-run the report if you want to re-run it.

des12437:
> Using this sample, I don't really see the point of the Reset button. It doesn't seem like it's needed. Just re-run the report if you want to re-run it.

The reset button UI is specified in the [common design](https://devtopia.esri.com/runtime/common-samples/blob/main/designs/CreateLoadReport/implementation-details.md#sample-ui), but I can remove it.

rolson:
You can talk to @yo1995 about if it should be removed but I don't see the point and it could simplify the code a bit, which is important.

yo1995 (commented):
I'm fine with removing the reset button. It is useful for desktop platforms so the common design needs no change.

rolson (commented):
In terms of coded-value hashable problem, you can simply use ObjectIdentifier.  See changes in [this branch](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/ryan/load-report-pr-feedback) for an idea of how to do that in an encapsulated way.

Shared/Samples/Create load report/CreateLoadReportView.swift
```diff
@@ -0,0 +1,66 @@
...
+                } catch {
+                    // Presents an error message if the utility network fails to load.
+                    self.error = error
+                }
+            }
+            .alert(isPresented: $isShowingAlert, presentingError: error)
+            .toolbar {
+                ToolbarItemGroup(placement: .bottomBar) {
+                    Button("Run") {
+                        Task { try await model.run() }
```
rolson, line 44: Any error thrown from here will be discarded.  Nothing will be presented to the user, nothing will be logged. I think I would advise a do/catch and setting the error property in this view.

Shared/Samples/Create load report/CreateLoadReportView.swift
```diff
@@ -0,0 +1,66 @@
...
+        didSet { isShowingAlert = error != nil }
+    }
+    
+    var body: some View {
+        LoadReportView(model: model)
+            .task {
+                do {
+                    try await model.setup()
+                } catch {
+                    // Presents an error message if the utility network fails to load.
```
rolson, line 36:
```suggestion
                    // Presents an error message if setup fails. This could occur if the utility network fails to load.
```

Shared/Samples/Create load report/CreateLoadReportView.swift
```diff
@@ -0,0 +1,66 @@
...
+            .overlay(alignment: .center) {
+                if let status = model.statusText {
+                    ZStack {
+                        Color.clear.background(.ultraThinMaterial)
+                        VStack {
+                            Text(status)
+                            ProgressView()
+                                .progressViewStyle(.circular)
+                        }
+                        .padding(6)
```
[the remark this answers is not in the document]
des12437: Fixed. This suggestion can also be applied to my other sample in review.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,283 @@
...
+        func setup() async throws {
+            try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+            try await loadUtilityNetwork()
+        }
+        
+        /// Loads the utility network.
+        private func loadUtilityNetwork() async throws {
+            statusText = "Loading utility network…"
+            try await utilityNetwork.load()
+            statusText = nil
```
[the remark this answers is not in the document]
des12437: Would the function observe load status changes to set the status text?

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -15,6 +15,28 @@
 import ArcGIS
 import SwiftUI
 
+extension CreateLoadReportView {
+    fileprivate struct PhaseSummaries {
```
rolson, line 19: Can you please document this type and its members? Also you may want to move it to the same extension as `PhaseSummary`.  I just placed it here in my branch to quickly prototype.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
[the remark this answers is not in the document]
rolson: No. It would be something that just took a closure and reset the status at the end of the closure.  However, it might not carry its own weight because it's easy to use a defer statement.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,292 @@
...
+                  let traceConfiguration = getTraceConfiguration() else { return }
+            
+            // Set the default expression.
+            initialExpression = traceConfiguration.traversability?.barriers as? UtilityTraceConditionalExpression
+            
+            // Create downstream trace parameters with elements and function outputs.
+            let traceParameters = UtilityTraceParameters(traceType: .downstream, startingLocations: [startingLocation])
+            traceParameters.addResultTypes([.elements, .functionOutputs])
+            
+            guard let definition = utilityNetwork.definition else { return }
```
rolson, line 107: this function returns early a lot without actually actually failing in any way that the caller of it would know that it actually failed.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,292 @@
...
+                .defaultTraceConfiguration
+        }
+        
+        func run() async throws {
+            statusText = "Creating load report…"
+            defer { statusText = nil }
+            
+            guard let phasesNetworkAttribute,
+                  let initialExpression,
+                  let traceParameters else { return }
```
rolson, line 180: This is an early return and it didn't actually run anything or fail.  How does the caller of this function know about the failure?  Same question was posed above. I think this is more of a question for the samples in general. What do you want to happen in the case where these variables are not there?  You could make them preconditions, but then you would need to make sure the "run" button is not enabled if any of these three variables are nil.

des12437: These early returns now throw an error to be presented to the user.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -91,11 +69,11 @@ extension CreateLoadReportView {
         private func loadUtilityNetwork() async throws {
             statusText = "Loading utility network…"
             try await utilityNetwork.load()
-            statusText = nil
+            defer { statusText = nil }
             
             guard let startingLocation = makeStartingLocation(),
                   // Get the base condition and trace configuration from a default tier.
-                  let traceConfiguration = getTraceConfiguration() else { return }
+                  let traceConfiguration = getTraceConfiguration() else { throw LoadError.cannotLoadNetwork }
```
rolson, line 76: The error is called `cannotLoadNetwork`, but the network did load or the call on line 71 would have thrown an error.

rolson: alternatively you can have the dependent functions be throwing and return non optionals:

```swift
private func makeStartingLocation() throws -> UtilityElement
```

And they can throw something like:

```
enum Error {
    case startingLocationNotFound
    case traceConfigurationNotFound
}
```

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -177,7 +155,7 @@ extension CreateLoadReportView {
             
             guard let phasesNetworkAttribute,
                   let initialExpression,
-                  let traceParameters else { return }
+                  let traceParameters else { throw LoadError.cannotGenerateReport }
```
rolson, line 158: I'm not sure this really helps the user. A few reasons, first it doesn't tell them why.  Second, it allows them to hit the `run` button when the `loadUtilityNetwork()` function actually couldn't find everything it was looking for.  It might be better to fail the setup or `loadUtilityNetwork()` if those variables cannot be filled out.

Shared/Samples/Create load report/CreateLoadReportView.swift
```diff
@@ -0,0 +1,73 @@
...
+    
+    var body: some View {
+        LoadReportView(model: model)
+            .task {
+                do {
+                    try await model.setup()
+                } catch {
+                    // Presents an error message if setup fails. This could occur if the utility network fails to load.
+                    self.error = error
+                    fatalError("Failed to setup sample.")
```
rolson, line 38: you don't want to fatal error here. The error will never be presented to the user.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
rolson: This comment still applies. I will write up a small branch to demonstrate what my line of thinking is here.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -251,29 +252,39 @@ extension CreateLoadReportView {
...
         }
         
+        /// Deletes the provided phase from the included phases.
+        /// - Parameter phase: The phase to delete.
         func deletePhase(_ phase: CodedValue) {
             excludedPhases.append(phase)
             includedPhases = includedPhases.filter { $0.name != phase.name }
             sortPhases()
             summaries.removeSummary(forPhase: phase)
         }
```
des12437: These properties are accessed by the view so cannot be made private.

rolson (approved):

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,375 @@
...
+        /// An error that occurred during setup.
+        private(set) var setupError: Error? {
+            didSet {
+                updateAllowsCreateLoadReport()
+                isShowingSetupError = setupError != nil
+            }
+        }
+        
+        /// A Boolean value indicating whether a setup error is showing.
+        @Published var isShowingSetupError = false
```
yo1995, line 66: Next time you can put the error and associated booleans in the view. In theory, the data model shouldn't know anything about the view. Here with the boolean in the model, the model knows the states of the view.

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
```diff
@@ -0,0 +1,375 @@
...
+            allowsCreateLoadReport = setupError == nil && !includedPhases.isEmpty
+        }
+        
+        /// Performs important tasks including adding credentials, loading and adding operational layers.
+        func setup() async {
+            do {
+                try await ArcGISEnvironment.authenticationManager.arcGISCredentialStore.add(.publicSample)
+                try await setupTraceParameters()
+            } catch {
+                self.setupError = error
```
yo1995, line 91:
```suggestion
                setupError = error
```

yo1995 (approved):

Shared/Samples/Create load report/CreateLoadReportView.Model.swift
rolson: > In theory, the data model shouldn't know anything about the view.

I think there are various ways to do these things. One of those ways is MVVM.  In the case of MVVM, the view model doesn't know how the view presents an error, but it knows if the error should be presented.  This makes the view model testable without a UI.

Opinions vary on this. But I'm of the opinion personally that MVVM works well with SwiftUI and that testable view models are nice.

Also in this case, the model had to know about the setup error so that it could enforce the "allows create load report" boolean. That would mean that the view and model would have duplicated state if the setup error was also on the view.

yo1995: agree

rolson (approved):

yo1995 (approved):

des12437:
Thank you both @rolson @yo1995

PR #221, opened 2023-07-06
[Update] Sample Viewer category front page

CalebRas:
## Description

This PR updates the front page of the sample viewer to organize the samples into categories, similar to the old iOS Sample Viewer. The new common design categories were used. All of the currently implemented samples fall under one of the sample categories. 

## Linked Issue(s)

- `swift/issues/4229`

## How To Test

* Load the sample viewer.
* Tap on a category to navigate to a list of samples in that category. 
* Tap on the search bar to search for a sample.

## Screenshots

|Before|After|
|:-:|:-:|
|Simulator Screen Shot - iPhone 14 Pro - 2023-07-06 at 15 49 35|Simulator Screen Shot - iPhone 14 Pro - 2023-07-06 at 15 43 26|

## To Discuss

* There is a bug where the navigation title at the top of the samples list jumps after tapping on a category and the list first loads.
* There is a bug when you back out of a sample after you entered it through a category. It then loads to the category view instead of the samples list view and it no longer lets you in that specific category.

PR #227, opened 2023-07-20
[Fix] Issue with sidebar navigation

philium:
Fixes issue described [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/221#issuecomment-1641069920), where going back from a sample takes the user all the way back to the Categories view, not the view for the category to which the sample belongs. This also means that when the categories are displayed in a sidebar, tapping on a category opens that category in the sidebar, which is expected behavior for this kind of a navigation setup on iOS.

### DIFF
--- a/Shared/Supporting Files/Views/CategoryGridView.swift
+++ b/Shared/Supporting Files/Views/CategoryGridView.swift
@@ -43,6 +43,7 @@ struct CategoryGridView: View {
                     } label: {
                         CategoryTitleView(category: category)
                     }
+                    .isDetailLink(false)
                     .clipShape(RoundedRectangle(cornerRadius: 15))
                     .contentShape(RoundedRectangle(cornerRadius: 30))
                     .buttonStyle(.plain)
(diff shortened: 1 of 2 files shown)


yo1995:
@CalebRas Can you double check if this is the expected behavior on all iOS 16+ targets? Thanks.

CalebRas:
> @CalebRas Can you double check if this is the expected behavior on all iOS 16+ targets? Thanks.

The bugs I was originally experiencing are both gone. Now, there is a new bug where the navigation tile jumps from inline to large when backing out of a sample.  


https://github.com/Esri/arcgis-maps-sdk-swift-samples/assets/34120440/522ec3bb-fe37-4c3c-9281-99084b5b6465

yo1995:
> Now, there is a new bug where the navigation tile jumps from inline to large when backing out of a sample.

That might be sth we have no control, unfortunately 😥 . The same code behaves correctly on iOS 17 beta, so I guess it is an iOS/SwiftUI bug.


https://github.com/Esri/arcgis-maps-sdk-swift-samples/assets/9660181/801396a1-fda5-4c99-8344-2c2d4e876aea

philium:
> Now, there is a new bug where the navigation tile jumps from inline to large when backing out of a sample.

That bug was reproducible before the category changes went in. I noticed it the other day.

dfeinzimer (approved):
Works nicely across all platforms

PR #228, opened 2023-07-20
[Update] Improve Sample Viewer search

CalebRas:
## Description

This PR updates the sample search bar in the Sample Viewer to include the searching of a sample's description and tags in addition to the name. Currently, the search bar performs a partial-text search on the names, a partial-text search on the descriptions, and a full-text search on the tags, in that order. The search results are divided into three sections, Name, Description, and Tags. Sections do not include samples that were found in a previous section, if any. Also, sections that have no results are hidden. If the search query is empty, a list of all the samples is displayed, sorted by name. Additionally, the query search text is bolded in the sample names and descriptions. 

## Linked Issue(s)

- `swift/issues/4230`

## How To Test

- Tap on the search bar on the main screen
- Enter a search query to see the results

## Screenshots

|Before|After|
|:-:|:-:|
|Simulator Screen Shot - iPhone 14 Pro - 2023-07-20 at 09 39 50|Simulator Screen Shot - iPhone 14 Pro - 2023-07-20 at 11 45 48|

## To Discuss

A couple things to think about:
- Is partial-text search for names, partial-text search for descirptions, and full-text search for tags okay?
- Should sections include results that were found previous section?
- Should sections with no results be hidden?
- Should bold or highlighted text be used? (See https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/228#discussion_r1283779963)

|Bold (Markdown)|Highlighted (`AttributedString`)|
|:-:|:-:|
|Bold|Highlighted|


Shared/Supporting Files/Extensions/CategoryView+Search.swift
```diff
@@ -15,13 +15,51 @@
...
+            sample.name.localizedCaseInsensitiveContains(query)
         }
+        return nameSearchResults
+    }
+    
+    /// Searches the samples using the sample's description and the query.
+    /// - Returns: The samples whose description partially matches the query.
+    func searchSamplesDescriptions() -> [Sample] {
+        // The samples already found by name with query.
+        let previousSearchResults = searchSamplesNames()
```
yo1995, line 35: This makes me think if you can use a state property `Set` to hold the temporary results (or the remainder of the samples to be searched), instead of cascading the results from previously run methods. Otherwise it would result in 1-2-3-4-… n-squared rounds of looping through the `samples` array as we add n criteria. 

> Should sections include results that were found previous section?

I'm not object to including duplicates in each section. Although in iOS, it is often represented by a segmented control-like UI. See Music app as an example.

yo1995 (commented):
Haven't finished the review yet, here are some initial comments.

Shared/Supporting Files/Extensions/CategoryView+Search.swift
CalebRas: I added state property, but I am not sure if it was in the pattern you were thinking off. So, let me know what you think.

Shared/Supporting Files/Extensions/CategoryView+Search.swift
```diff
@@ -14,14 +14,44 @@
...
+    /// - Returns: The samples whose name partially matches the query.
+    func searchSamplesNames() -> [Sample] {
+        // Return all the samples when query is empty.
+        guard !query.isEmpty else { return samples }
+        
+        // Perform a partial text search using the sample's name and the query.
+        let nameSearchResults = samples.filter { sample in
+            sample.name.localizedCaseInsensitiveContains(query)
         }
+        previousSearchResults.formUnion(nameSearchResults.map { $0.name })
```
yo1995, line 28:
```suggestion
        previousSearchResults.formUnion(nameSearchResults.map(\.name))
```
Can use keypath here.

Shared/Supporting Files/Views/CategoryView.swift
```diff
@@ -24,20 +24,53 @@ struct CategoryView: View {
...
+                        }
+                    }
+                    if !tagsSearchResults.isEmpty {
+                        Section(header: Text("Tag Results")) {
+                            SampleListView(samples: tagsSearchResults, query: query)
+                        }
+                    }
+                }
+                .onChange(of: query) { _ in
+                    previousSearchResults = []
```
yo1995, line 66:
```suggestion
                    previousSearchResults.removeAll(keepingCapacity: true)
```
In this specific usecase, I think clearing the set while keeping capacity makes sense, as the set might be refilled immediately.

Shared/Supporting Files/Extensions/CategoryView+Search.swift
```diff
@@ -14,14 +14,44 @@
...
+    /// Searches the samples using the sample's description and the query.
+    /// - Returns: The samples whose description partially matches the query.
+    func searchSamplesDescriptions() -> [Sample] {
+        // Perform a partial text search using the sample's description and
+        // the query for the samples that are not already found.
+        let descriptionSearchResults = samples.filter { sample in
+            sample.description.localizedCaseInsensitiveContains(query) &&
+            !previousSearchResults.contains(sample.name)
+        }
+        previousSearchResults.formUnion(descriptionSearchResults.map { $0.name })
```
yo1995, line 41:
```suggestion
        previousSearchResults.formUnion(descriptionSearchResults.map(\.name))
```

Shared/Supporting Files/Views/CategoryView.swift
```diff
@@ -24,20 +24,53 @@ struct CategoryView: View {
...
+                    nameSearchResults = searchSamplesNames()
+                    descriptionSearchResults = searchSamplesDescriptions()
+                    tagsSearchResults = searchSamplesTags()
```
yo1995, line 69: I might be playing devil's advocate here, but here are a few thoughts popped up:

- Try to make the search methods "[pure function](https://en.wikipedia.org/wiki/Pure_function)" so they don't have side effects. Sth like

```swift
    func searchNames(in samples: [Sample], with query: String) -> [Sample] {
        samples.filter { sample in
            sample.name.localizedCaseInsensitiveContains(query)
        }
    }
```

- How about the other logics? It may be better to just move them to the same file of the view. This also allows us to make the `previousSearchResults` a local variable, no need to be a state.

```swift
    private func searchSamples() async {
        // The names of the samples already found in a previous section.
        var previousSearchResults: Set<String> = []
        
        nameSearchResults = searchNames(in: samples, with: query)
        previousSearchResults.formUnion(nameSearchResults.map(\.name))
        
        descriptionSearchResults = searchDescriptions(in: samples, with: query)
            .filter { !previousSearchResults.contains($0.name) }
        previousSearchResults.formUnion(descriptionSearchResults.map(\.name))
        
        tagsSearchResults = searchTags(in: samples, with: query)
            .filter { !previousSearchResults.contains($0.name) }
    }
```

- Here though, I do hesitate on whether or not marking the helper method `searchSamples` as `async`. It is currently completely synchronous, and even with all 200-ish samples included in a future version, the time complexity would only be around O(3 * 200 * n) where n denotes the average length of a sample name / description / [tags]. That is a very short amount of time given the iOS 15+ device's speed. Still, it may be considered a "time consuming" task depend on how we view it. Let's don't add `async` for now.

yo1995 (commented):
I'll take a look at the bold text method after these suggestions are addressed.

Shared/Supporting Files/Views/SampleListView.swift
```diff
@@ -58,5 +70,34 @@ private extension SampleListView {
                 .animation(.easeOut(duration: 0.2), value: isShowingDescription)
             }
         }
+        
+        /// Bolds the first occurrence of substring within a given string using markdown.
+        /// - Parameters:
+        ///   - text: The `String` containing the substring.
+        ///   - substring: The substring to bold.
+        /// - Returns: The `String` with the bolded substring.
+        func boldSubstring(_ text: String, substring: String) -> String {
```
yo1995, line 79: I propose 2 options. Can you grab screenshots for both and add them to the screenshots in the description (that capture all name, description, and tags), so the next reviewers can vote on which path to go? 

Personally I see the bold text more than highlight, but I slightly lean towards highlight myself.

```swift
        // 1
        private func boldSubstring(_ text: String, substring: String) -> String {
            if let range = text.range(of: substring.trimmingCharacters(in: .whitespacesAndNewlines), options: .caseInsensitive) {
                var boldedText = text
                boldedText.replaceSubrange(range, with: "**" + text[range] + "**")
                return boldedText
            }
            return text
        }
        // 2
        private func attributedString(_ text: String, substring: String) -> AttributedString {
            var attributedString = AttributedString(text)
            if let range = attributedString.range(of: substring, options: .caseInsensitive) {
                attributedString[range].backgroundColor = .yellow.opacity(0.25)
            }
            return attributedString
        }
```

1. There is that issue with markdown syntax that contains leading/trailing whitespace. By trimming them off it works. Also the code can be much simplified.
2. We can also use the new [`AttributedString`](https://developer.apple.com/documentation/foundation/attributedstring) that Apple recommends to use starting from iOS 15. It works seamlessly with SwiftUI `Text` and is highly customizable.

- Mails uses yellow highlight; Maps, Contacts, System level search (Siri Suggestions) use bold text.
- I'll leave the method naming to Phil R.

CalebRas: I am also in favor of using the `AttributedString` since Apple recommends using it and there are more of examples of it in their apps. It also shows up much better in the description than the bold text does. I will wait for others input, though.

philium: If you're going to stick with bolding (which may be fine, I don't know; I haven't looked at the rest of the PR), then unless I'm missing something, the implementation could be greatly simplified. Part of that simplification is moving the method to an extension of `String`:

```swift
private extension String {
    func boldingFirstOccurrence(of substring: String) -> String {
        guard let range = localizedStandardRange(of: substring) else {
            return self
        }
        return replacingCharacters(in: range, with: "**\(substring)**")
    }
}
```

yo1995: I agree, we can add the method as an extension to `String`.

yo1995 (commented):
Please make the bold method an private extension. I'll add Phil to review as well.

Shared/Supporting Files/Views/CategoryView.swift
philium: According to [the documentation](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/), the `async` keyword is used "to indicate that a function or method is asynchronous." An asynchronous function/method is defined as "a special kind of function or method that can be suspended while it’s partway through execution." Also, "inside an asynchronous method, the flow of execution is suspended only when you call another asynchronous method — suspension is never implicit or preemptive — which means every possible suspension point is marked with `await`." Therefore a function/method should not be marked `async` if it does not await anything (ie. does not contain an `await` keyword), regardless of how long the function/method takes to execute.

Shared/Supporting Files/Views/CategoryGridView.swift
```diff
@@ -38,8 +38,10 @@ struct CategoryGridView: View {
             LazyVGrid(columns: columns) {
                 ForEach(categories, id: \.self) { category in
                     NavigationLink {
-                        SampleListView(samples: samples.filter { $0.category == category })
-                            .navigationTitle(category)
+                        List {
```
philium, line 41: Why was `List` moved here? Now the name `SampleListView` is really confusing because it isn't actually a list. The list and navigation title really should be a part of that view and a better name for that view would be `CategoryView`, since it is a view that displays a category.

Shared/Supporting Files/Views/CategoryView.swift
```diff
@@ -24,20 +24,47 @@ struct CategoryView: View {
...
     @Binding private(set) var query: String
     
+    /// The samples to display in the name section of the search list.
+    @State private var nameSearchResults: [Sample] = []
+    
+    /// The samples to display in the description section of the search list.
+    @State private var descriptionSearchResults: [Sample] = []
+    
+    /// The samples to display in the tags section of the search list.
+    @State private var tagsSearchResults: [Sample] = []
```
philium, line 34: Since these are all related, you might consider combining them:

```swift
struct SearchResult {
    var nameMatches: [Sample]
    var descriptionMatches: [Sample]
    var tagMatches: [Sample]
}

@State private var searchResult: SearchResult?
```

Shared/Supporting Files/Views/SampleListView.swift
```diff
@@ -40,9 +51,10 @@ private extension SampleListView {
             } label: {
                 HStack {
                     VStack(alignment: .leading, spacing: 5) {
-                        Text(sample.name)
+                        Text(.init(sample.name.boldingFirstOccurrence(of: boldedText)))
+                        
                         if isShowingDescription {
-                            Text(sample.description)
+                            Text(.init(sample.description.boldingFirstOccurrence(of: boldedText)))
```
philium, line 57:
```suggestion
                            Text(sample.description.boldingFirstOccurrence(of: boldedText))
```

Shared/Supporting Files/Views/SampleListView.swift
```diff
@@ -61,3 +73,18 @@ private extension SampleListView {
         }
     }
 }
+
+private extension String {
+    /// Bolds the first occurrence of substring within the string using markdown.
+    /// - Parameters:
+    ///   - substring: The substring to bold.
```
philium, line 80:
```suggestion
    /// - Parameter substring: The substring to bold.
```

Shared/Supporting Files/Views/CategoryView.swift
yo1995: Thanks for the comment Phil. If I want to wrap a long-running synchronous call in Swift, is the correct way to use an actor or go back to [Thread](https://developer.apple.com/documentation/foundation/thread)?

yo1995:
I'll assign this PR to myself and work on the feedback.

Shared/Supporting Files/Views/CategoryGridView.swift
yo1995: I think Caleb moved it here to cope with the problem that the search result list need to  have different sections, while the category list doesn't need section. By changing "SampleListView" into a `ForEach`, it can deal with both of the cases. Do you have an alternative? This UI is seen in the Shortcuts and Mails app.

Another idea would be instead of putting the result lists into sections, they can be in separate lists switched by a segmented control. This UI is seen in Files and Music app.

philium: I think that the category view (with the list of all the samples in a given category) and the search results view should be two separate views. There's not a whole lot of overlap there. But as a single view, it seems like it could have a list but then consumers could change the list style as needed (`listStyle(_:)`).

yo1995: Thanks. I've made the following changes

- Remove `SampleListView` and moved the `SampleRow` view into a separate file
- Created `SearchResultView` to display the search results
- Inlined the sample list under a category grid item, and set its list style to "sidebar" (in CategoryGridView.swift)

yo1995:
Ready for another look.

Shared/Supporting Files/Views/SampleRow.swift
```diff
@@ -0,0 +1,77 @@
...
+                }
+                Spacer()
+                Button {
+                    isShowingDescription.toggle()
+                } label: {
+                    Image(systemName: isShowingDescription ? "info.circle.fill" : "info.circle")
+                }
+                .buttonStyle(.borderless)
+            }
+            .animation(.easeOut(duration: 0.2), value: isShowingDescription)
```
philium, line 60: I would argue that this is the actual row and that the creation of the navigation link could (and should) be moved up to the parent. Then the view itself would just need (besides the one state property) a title and description property, both attributed strings. That would allow it to be tested in a preview. A "convenience" initializer could be added that takes a sample:

```swift
init(sample: Sample, query: String) {
    self.init(
        title: sample.description.boldingFirstOccurrence(of: query)
        description: sample.description.boldingFirstOccurrence(of: query)
    )
}
```

Shared/Supporting Files/Views/SampleRow.swift
```diff
@@ -0,0 +1,77 @@
...
+                            .font(.caption)
+                            .foregroundColor(.secondary)
+                            .transition(.move(edge: .top).combined(with: .opacity))
+                    }
+                }
+                Spacer()
+                Button {
+                    isShowingDescription.toggle()
+                } label: {
+                    Image(systemName: isShowingDescription ? "info.circle.fill" : "info.circle")
```
philium, line 56:
```suggestion
                    Image(systemName: "info.circle")
                        .symbolVariant(isShowingDescription ? .fill : .none)
```

Shared/Supporting Files/Views/SampleRow.swift
```diff
@@ -0,0 +1,77 @@
...
+                    
+                    if isShowingDescription {
+                        Text(sample.description.boldingFirstOccurrence(of: boldedText))
+                            .font(.caption)
+                            .foregroundColor(.secondary)
+                            .transition(.move(edge: .top).combined(with: .opacity))
+                    }
+                }
+                Spacer()
+                Button {
```
philium, line 53: FYI: I just noticed that this button is not visible when the row is selected. Is that a known issue?

Shared/Supporting Files/Views/SearchResultView.swift
```diff
@@ -0,0 +1,121 @@
...
+    /// The search query in the search bar.
+    let query: String
+    
+    /// The search result to display in the various sections.
+    @State private var searchResult = SearchResult(nameMatches: [], descriptionMatches: [], tagMatches: [])
+    
+    var body: some View {
+        List {
+            if !searchResult.nameMatches.isEmpty {
+                Section(header: Text("Name Results")) {
```
philium, line 30: This `Section` initializer is deprecated.

```suggestion
                Section("Name Results") {
```

Shared/Supporting Files/Views/SearchResultView.swift
```diff
@@ -0,0 +1,121 @@
...
+                    ForEach(searchResult.nameMatches, id: \.name) { sample in
+                        SampleRow(sample: sample, boldedText: query)
+                    }
+                }
+            }
+            if !searchResult.descriptionMatches.isEmpty {
+                Section(header: Text("Description Results")) {
```
philium, line 37: This `Section` initializer is deprecated.

```suggestion
                Section("Description Results") {
```

Shared/Supporting Files/Views/SearchResultView.swift
```diff
@@ -0,0 +1,121 @@
...
+                    ForEach(searchResult.descriptionMatches, id: \.name) { sample in
+                        SampleRow(sample: sample, boldedText: query)
+                    }
+                }
+            }
+            if !searchResult.tagMatches.isEmpty {
+                Section(header: Text("Tags Results")) {
```
philium, line 44: This `Section` initializer is deprecated.

```suggestion
                Section("Tags Results") {
```

Although consider using "Tag Results" as it reads better and the result matches a single tag.

```suggestion
                Section("Tag Results") {
```

Shared/Supporting Files/Views/SearchResultView.swift
```diff
@@ -0,0 +1,121 @@
...
+
+struct SearchResultView: View {
+    /// All samples retrieved from the Samples directory.
+    let samples: [Sample]
+    
+    /// The search query in the search bar.
+    let query: String
+    
+    /// The search result to display in the various sections.
+    @State private var searchResult = SearchResult(nameMatches: [], descriptionMatches: [], tagMatches: [])
```
philium, line 25: This doesn't need to be a state property because it is never updated by the view itself. Instead, the view should be initialized with this:

```swift
let searchResult: SearchResult
```

You can either create that outside of this view or the way it is now with an initializer:

```swift
extension SearchResultView {
    init(samples: [Sample], query: String) {
        let nameMatches: [Sample]
        let descriptionMatches: [Sample]
        let tagMatches: [Sample]
        
        if query.isEmpty {
            // Show all samples in the name section when query is empty.
            nameMatches = samples
            descriptionMatches = []
            tagMatches = []
        } else {
            // The names of the samples already found in a previous section.
            var previousSearchResults: Set<String> = []
            
            // Partially match a query to a sample's name.
            nameMatches = samples.filter { $0.name.localizedCaseInsensitiveContains(query) }
            previousSearchResults.formUnion(nameMatches.map(\.name))
            
            // Partially match a query to a sample's description.
            descriptionMatches = samples.filter { $0.description.localizedCaseInsensitiveContains(query) }
                .filter { !previousSearchResults.contains($0.name) }
            previousSearchResults.formUnion(descriptionMatches.map(\.name))
            
            // Match a query to one of the sample's tags.
            tagMatches = samples.filter { sample in
                sample.tags.contains { tag in
                    tag.localizedCaseInsensitiveCompare(query) == .orderedSame
                }
            }
            .filter { !previousSearchResults.contains($0.name) }
        }
        
        let searchResult = SearchResult(
            nameMatches: nameMatches,
            descriptionMatches: descriptionMatches,
            tagMatches: tagMatches
        )
        self.init(searchResult: searchResult)
    }
}
```

Then the `onAppear` and `onChange` modifiers aren't needed.

Shared/Supporting Files/Views/CategoryGridView.swift
```diff
@@ -38,8 +38,11 @@ struct CategoryGridView: View {
             LazyVGrid(columns: columns) {
                 ForEach(categories, id: \.self) { category in
                     NavigationLink {
-                        SampleListView(samples: samples.filter { $0.category == category })
-                            .navigationTitle(category)
+                        List(samples.filter { $0.category == category }, id: \.name) { sample in
+                            SampleRow(sample: sample)
+                        }
+                        .listStyle(.sidebar)
+                        .navigationTitle(category)
```
philium, line 45: Methinks this should be a view named `CategoryView`:

```swift
struct Category {
    let name: String
    let samples: [Sample]
}

struct CategoryView: View {
    let category: Category

    var body: some View {
        List(category.samples, id: \.name) { sample in
            SampleRow(sample: sample)
        }
        .listStyle(.sidebar)
        .navigationTitle(category.name)
    }
}
```

Shared/Supporting Files/Views/SampleRow.swift
yo1995: #170 tried to fix that, but it was not merged due to a navigation bug related to [List with selection](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/170#pullrequestreview-1388986734).

#226 could have get rid of the info button, but the newly proposed disclosure group UI was rejected by a few reviewers.

Shared/Supporting Files/Views/ContentView.swift
```diff
@@ -39,8 +39,8 @@ struct ContentView: View {
     }
     
     var sidebar: some View {
-        CategoryView(samples: samples, query: $query)
-            .searchable(text: $query, prompt: "Search By Sample Name")
+        CategoryView(samples: samples, query: query)
+            .searchable(text: $query, prompt: "Search")
```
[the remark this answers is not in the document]
yo1995: When I put the `searchable` modifier on `Sidebar`, the search text field padding looks more natural than put it within the sidebar. I'll leave it for further comments.

Shared/Supporting Files/Views/CategoryGridView.swift
yo1995: I named it `CategoryList` for now.

Shared/Supporting Files/Views/SampleRow.swift
yo1995: For some reason, the compiler doesn't allow me to use the synthesized initializer outside the `SampleRow` file. 🤔

Shared/Supporting Files/Views/CategoriesView.swift
```diff
@@ -49,12 +49,29 @@ struct CategoryGridView: View {
...
+        /// The samples in a category.
+        let samples: [Sample]
+        
+        var body: some View {
+            List(samples, id: \.name) { sample in
+                NavigationLink {
+                    SampleDetailView(sample: sample)
+                        .id(sample.name)
+                } label: {
+                    SampleRow(sample: sample, query: "")
```
yo1995, line 68: I couldn't use the synthesized initializer for some reason. Maybe I need to look up the Swift doc 🤔 , but I vaguely recall it retains the default initializer even if a "convenience" one is added to a struct, and per https://www.swiftbysundell.com/tips/struct-convenience-initializers/

philium: You'll notice in the code in the linked article, they are adding an initializer in an extension. Adding custom initializers in the struct declaration itself will suppress generation of the memberwise initializer.

Shared/Supporting Files/Views/ContentView.swift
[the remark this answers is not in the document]
philium: How is the padding different? Having the modifier on the category view or inside the category view should be functionally equivalent. I'm not seeing any visual difference either:

Before | After
------ | -----
Screenshot 2023-08-10 at 8 35 19 AM | Screenshot 2023-08-10 at 8 36 19 AM

Shared/Supporting Files/Views/SampleRow.swift
yo1995: Making them non-private solved the problem 🤦  Thank you.

Shared/Supporting Files/Views/ContentView.swift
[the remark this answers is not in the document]
yo1995: |Before|After|
|:-:|:-:|
|Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2023-08-11 at 14 29 13|Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2023-08-11 at 14 23 37|

Maybe this only happens on iPad? Notice the small additional padding of the navigation title and search bar, that makes them not aligned with the categories view.

Besides, I cannot quite get the `searchable` modifier working when I put it on the `Group` of conditional `CategoriesView`/`SamplesSearchView`. When the search bar becomes the first responder, it doesn't switch to the `SamplesSearchView`.

The [code example in the doc](https://developer.apple.com/documentation/swiftui/environmentvalues/issearching) does put the modifier within its parent view.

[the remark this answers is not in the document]
philium: Fascinating! So be it.

philium (approved):

Shared/Supporting Files/Views/SamplesSearchView.swift
```diff
@@ -0,0 +1,138 @@
...
+                        NavigationLink {
+                            SampleDetailView(sample: sample)
+                                .id(sample.name)
+                        } label: {
+                            SampleRow(sample: sample, query: query)
+                        }
+                    }
+                }
+            }
+        }
```
dfeinzimer, line 62:
```suggestion
        }
        .listStyle(.insetGrouped)
```

This would remove the disclosure indicators on iOS, unless we want users to be able to collapse the categories.

dfeinzimer (approved):
Just the one suggestion

Shared/Supporting Files/Views/SamplesSearchView.swift
yo1995: From this [blog post](https://sarunw.com/posts/swiftui-list-style/#sidebar) it seems to me that the most semantically suitable style here should be iOS: `insetGrouped` and iPadOS/macOS: `sidebar`. Apple didn't do that though. So I don't think it is a big deal.

I actually like to have the disclosure button, and by using the default "automatic" list style, maybe it is more future proof. 🙂

Shared/Supporting Files/Views/SampleRow.swift
yo1995: This will be fixed in another PR.

yo1995:
Thanks for all your suggestions!

yo1995:
When I was closing the opened tabs, I found the [`searchScopes(_:scopes:)`](https://developer.apple.com/documentation/swiftui/view/searchscopes(_:scopes:)) API which is useful for further improving the experience. Will log a card for iOS 16+ improvements.

PR #230, opened 2023-07-27
[Fix] Center category grid text

CalebRas:
## Description

This PR implements a fix to the Sample Viewer by centering category title text on the category grid tiles. Initially, text was aligned left when there were multiple lines.

## How To Test

- Load the Sample Viewer on iPad and verify the text is centered on the category grid titles.
- Verify the app works the same otherwise. 

## Screenshots

|Before|After|
|:-:|:-:|
|iPad Screen Shot, Before|iPad Screen Shot, After|


dfeinzimer (approved):
Looks good on all platforms

PR #236, opened 2023-08-07
[Fix] Clicking anywhere on the cell is not working in Analyze network with subnetwork trace sample

des12437:
## Description
List items can now be selected by clicking anywhere in the cell. Before only the cell text was selectable. This PR also has a bug fix to reset previous textfield input. 

## Linked Issue(s)

- `swift/issues/4402`

## How To Test
Add condition by selecting the attribute, comparison, and value. Make sure that list item can be selected by clicking anywhere in the cell. 

## To Discuss

I tried to use a Button to avoid use of `.onTapGesture` but the UI had a few issues including incorrect accent color and line separator style, so I decided to use the original implementation:


saratchandrakarumuri (approved):
Now the selection is working as expected

PR #243, opened 2023-08-09
[New] Add Custom Dynamic Entity Data Source

zkline101:
## Description

This PR implements `Add custom dynamic entity data source` in `Layers` category.
URL to README: 

## Linked Issue(s)

- `swift/issues/4383`

## How To Test
- Dynamic entities should start to appear on the map
- Tap on a dynamic entity and all the attributes for that entity should appear in a callout

## Screenshots
add-custom-dynamic-entity-data-source


Samples.xcodeproj/project.pbxproj
```diff
@@ -1561,7 +1605,6 @@
 				79B7B80A2A1BF8EC00F57C27 /* CreateAndSaveKMLView.swift in Sources */,
 				7573E81C29D6134C00BEED9C /* TraceUtilityNetworkView.Enums.swift in Sources */,
 				7573E81A29D6134C00BEED9C /* TraceUtilityNetworkView.Model.swift in Sources */,
-				0005580C28185C0600224BC6 /* SampleList.swift in Sources */,
```
yo1995, line 1564: This is fixed in #245

Samples.xcodeproj/project.pbxproj
```diff
@@ -795,6 +809,7 @@
 				00E5401427F3CCA200CF66D5 /* Products */,
 				003D7C332821EBCC009DDFD2 /* Scripts */,
 				00E5400B27F3CCA100CF66D5 /* Shared */,
+				79D84D092A815AC700F45262 /* Frameworks */,
```
yo1995, line 812:
```suggestion
```
This might be added because the daily build was used.

Samples.xcodeproj/project.pbxproj
```diff
@@ -970,6 +993,23 @@
 			path = "Create and save KML file";
 			sourceTree = "<group>";
 		};
+		79D84D092A815AC700F45262 /* Frameworks */ = {
+			isa = PBXGroup;
+			children = (
+			);
+			name = Frameworks;
+			sourceTree = "<group>";
+		};
```
yo1995, line 1002:
```suggestion
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+    /// The placement of the callout.
+    @State private var calloutPlacement: CalloutPlacement?
+    
+    /// A dynamic entity observation that was selected on the screen.
+    @State private var selectedObservation: DynamicEntityObservation?
+    
+    var body: some View {
+        MapViewReader { proxy in
+            MapView(map: model.map)
+                .onSingleTapGesture { screenPoint, _  in
```
yo1995, line 34:
```suggestion
                .onSingleTapGesture { screenPoint, _ in
```

Shared/Samples/Add custom dynamic entity data source/README.md
```diff
@@ -0,0 +1,40 @@
...
+* CustomDynamicEntityFeedEvent
+* DynamicEntity
+* DynamicEntityDataSource
+* DynamicEntityLayer
+* LabelDefinition
+* TrackDisplayProperties
+
+## About the data
+
+This sample uses a .json file containing observations of marine vessels in the Pacific North West.
```
yo1995, line 36:
```suggestion
This sample uses a .json file containing observations of marine vessels in the Pacific North West.
This sample uses a [.json file containing observations of marine vessels in the Pacific North West](https://www.arcgis.com/home/item.html?id=a8a942c228af4fac96baa78ad60f511f) hosted on ArcGIS Online.
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+                    calloutPlacement = nil
+                }
+                .callout(placement: $calloutPlacement.animation(.default.speed(2))) { _ in
+                    let attributes = selectedObservation!.attributes
+                    VStack(alignment: .leading) {
+                        // Display all the attributes in the callout.
+                        ForEach(Array(attributes.sorted(by: { $0.key < $1.key })), id: \.key) { item in
+                            Text("\(item.key): \(item.value as? String ?? "")")
+                        }
+                    }
```
yo1995, line 47: No need to display all the attributes, the implementation details showed "VesselName", "CallSign", "COG", "SOG".

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+                }
+                .task(id: tappedScreenPoint) {
+                    guard let tappedScreenPoint = tappedScreenPoint,
```
yo1995, line 50:
```suggestion
                    guard let tappedScreenPoint,
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
```diff
@@ -0,0 +1,140 @@
...
+private struct VesselFeed: CustomDynamicEntityFeed {
+    let events: AsyncPublisher<AnyPublisher<CustomDynamicEntityFeedEvent, Never>>
+    
+    /// Creates a feed that is receiving the events that are being passed in.
+    ///
+    /// The events will be emitted with a delay so it resembles live data.
+    /// - Parameter events: The events that will be emitted by the feed
+    init<S>(events: S) where S: Sequence, S.Element == CustomDynamicEntityFeedEvent {
+        self.events = events.publisher
+            .delay(for: 1, scheduler: RunLoop.main)
```
yo1995, line 126: I'll ask other platform to see how they emit the changes. Since there are a few thousands of entries, maybe we can make the delay smaller.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
```diff
@@ -0,0 +1,140 @@
...
+                        
+                        // We successfully decoded the vessel JSON so we should
+                        // add that vessel as a new observation.
+                        return .newObservation(
+                            geometry: Point(x: geometry.x, y: geometry.y, spatialReference: .wgs84),
+                            attributes: decodable.attributes.makeDictionary()
+                        )
+                    }
+                
+                return VesselFeed(events: feedEvents)
```
yo1995, line 76: Initially I thought it would be better reading the JSON file [line by line](https://stackoverflow.com/a/75882427/14369688), but it seems the `feedEvents` is a static sequence rather than an async sequence, which means all the events are in memory before the data source starts. Is it how this API supposed to work?

Shared/Samples/Add custom dynamic entity data source/Vessel.swift
```diff
@@ -0,0 +1,64 @@
...
+            case callSign = "CallSign"
+            case globalID = "globalid"
+        }
+        
+        private let mmsi: String
+        private let sog: Double
+        private let cog: Double
+        private let vesselName: String
+        private let callSign: String
+        private let globalID: String
```
yo1995, line 42: If the suggestion above is applied, the `globalID` field can be omitted.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// This schema is derived from the first row in the custom data source.
+    static var vesselFields: [Field] = [
+        Field(type: .text, name: "MMSI", alias: "MMSI", length: 256),
+        Field(type: .float64, name: "SOG", alias: "SOG", length: 8),
+        Field(type: .float64, name: "COG", alias: "COG", length: 8),
+        Field(type: .text, name: "VesselName", alias: "VesselName", length: 256),
+        Field(type: .text, name: "CallSign", alias: "CallSign", length: 256),
+        Field(type: .text, name: "globalid", alias: "globalid", length: 256)
+    ]
+}
```
philium, line 114: This property should be computed. Note: Explicit `return` used to silence linter warning.

```suggestion
private extension Array<Field> {
    /// An array of fields that match the attributes of each observation in the data source.
    ///
    /// This schema is derived from the first row in the custom data source.
    static var vesselFields: Self {
        return [
            Field(type: .text, name: "MMSI", alias: "MMSI", length: 256),
            Field(type: .float64, name: "SOG", alias: "SOG", length: 8),
            Field(type: .float64, name: "COG", alias: "COG", length: 8),
            Field(type: .text, name: "VesselName", alias: "VesselName", length: 256),
            Field(type: .text, name: "CallSign", alias: "CallSign", length: 256),
            Field(type: .text, name: "globalid", alias: "globalid", length: 256)
        ]
    }
}
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
```diff
@@ -0,0 +1,140 @@
...
+            .values
+    }
+}
+
+private extension URL {
+    /// The URL to the selected vessels JSON data.
+    static var selectedVesselsDataSource: URL {
+        Bundle.main.url(
+            forResource: "AIS_MarineCadastre_SelectedVessels_CustomDataSource",
+            withExtension: "json"
```
philium, line 137: This really is a misnomer as the file is not actually valid JSON syntax. It really is just a text file that contains JSON snippets. But I'm guessing there's nothing that we can do about that?

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+    @State private var selectedObservation: DynamicEntityObservation?
+    
+    var body: some View {
+        MapViewReader { proxy in
+            MapView(map: model.map)
+                .onSingleTapGesture { screenPoint, _  in
+                    tappedScreenPoint = screenPoint
+                    
+                    // Hides the callout.
+                    calloutPlacement = nil
```
philium, line 38: Consider doing this when the identify operation doesn't return any results. The way it is now, the animation is clearly off:

Simulator Screen Recording - iPhone 14 Pro - 2023-08-11 at 16 05 39

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+                }
+                .callout(placement: $calloutPlacement.animation(.default.speed(2))) { _ in
+                    let attributes = selectedObservation!.attributes
+                    VStack(alignment: .leading) {
+                        // Display all the attributes in the callout.
+                        ForEach(Array(attributes.sorted(by: { $0.key < $1.key })), id: \.key) { item in
```
philium, line 44:
```suggestion
                        ForEach(attributes.sorted(by: { $0.key < $1.key }), id: \.key) { item in
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+                            Text("\(item.key): \(item.value as? String ?? "")")
```
philium, line 45: This would allow the attribute values to be either a string or a double, ie. no need to use `String.init(format:)` when creating the dictionary.

```suggestion
                            Text("\(item.key): \(String(describing: item.value))")
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+    @StateObject private var model = Model()
+    
+    /// The point on the screen the user tapped.
+    @State private var tappedScreenPoint: CGPoint?
+    
+    /// The placement of the callout.
+    @State private var calloutPlacement: CalloutPlacement?
+    
+    /// A dynamic entity observation that was selected on the screen.
+    @State private var selectedObservation: DynamicEntityObservation?
```
philium, line 29: This doesn't actually need to be stored. Just get it from the callout placement itself:

```swift
.callout(placement: $calloutPlacement.animation(.default.speed(2))) { placement in
    let attributes = (placement.geoElement as! DynamicEntityObservation).attributes
    ...
}
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
yo1995: William said they used 10ms delay.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,64 @@
...
+                    guard let tappedScreenPoint = tappedScreenPoint,
+                          let identifyResult = try? await proxy.identify(
+                            on: model.dynamicEntityLayer,
+                            screenPoint: tappedScreenPoint,
+                            tolerance: 2
+                          ),
+                          let firstObservation = identifyResult.geoElements.first as? DynamicEntityObservation else { return }
+                    
+                    // Set the callout placement to the observation that was tapped on.
+                    calloutPlacement = .geoElement(firstObservation)
```
philium, line 59: `firstObservation` isn't needed. Just map over the first geo element:

```swift
calloutPlacement = identifyResult.geoElements.first.map { .geoElement($0) }
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
philium: Something I just thought of. Instead of turning the vessel's attributes into a dictionary, consider just storing the vessel attributes as a value in the dictionary:

```swift
let vesselAttributes = (placement.geoElement as! DynamicEntityObservation).attributes["vesselAttributes"] as! Vessel.Attributes
```

Then the attributes could just be used here directly, simplifying things significantly.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
philium: I too was thinking about this. It can be done with `FileHandle` or with `URL`, but I couldn't find a way to get a publisher from an async sequence. We could still go down that route using [Throttle](https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncAlgorithms.docc/Guides/Throttle.md), but it would require adding another dependency. I don't really see a problem with that.

zkline101: Hm, I see your point. I was thinking since we already have the data available we can just load it beforehand but maybe that is not the best to show users since they will probably be receiving live data.

zkline101: > I don't really see a problem with that.

Is it worth adding a dependency, just for this one sample?

philium: I think so. Async sequences are the new kid on the block while Combine is aging quickly. If we anticipate that users will need to do something similar (read a large file line-by-line asynchronously), it would be great to demonstrate for them the best way to do that. Using a publisher definitely isn't, since it requires reading the whole file into memory first.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
yo1995: I can coordinate with other platforms to fix the JSON. Is it OK if I simply add a `[]` to the whole file to make it an array?

philium: It would also need commas at the end of every line (except for the last one), right? Although, that could inhibit reading and parsing the vessels in line by line, at least without some massaging.

yo1995: I agree, we should keep the current line by line structure. It's so called [JSON Lines](https://jsonlines.org/) ([SO thread](https://stackoverflow.com/questions/12451431/loading-and-parsing-a-json-file-with-multiple-json-objects)) and has a recommended file extension `.jsonl`. Do you want me to change it to that?

zkline101: Interesting, I haven't heard of that before. If that is the better way to represent this file then I am ok with the file extension change.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Model.swift
yo1995: After some reading I didn't find a good native way to "map" an async sequence to another, as pointed out by this user on SO: https://stackoverflow.com/a/73426716/14369688

So I borrowed his idea to roll our own feed with a `Task` block. But I haven't investigate the lifetime of this approach. It seems to work just fine. Bottom line, we can add a feed cancel/disconnect method if we need? 🤔 

What about sth like this?

```swift
            // call site of the custom feed struct
            let customDataSource = CustomDynamicEntityDataSource(info: info) {
                VesselFeed(inputLog: URL.selectedVesselsDataSource.lines)
            }

private struct VesselFeed: CustomDynamicEntityFeed {
    let inputLog: AsyncLineSequence<URL.AsyncBytes>
    
    var events: AsyncStream<CustomDynamicEntityFeedEvent> {
        let decoder = JSONDecoder()
        
        return AsyncStream<CustomDynamicEntityFeedEvent> { continuation in
            Task {
                for try await line in inputLog {
                    try await Task.sleep(nanoseconds: 10_000_000)
                    
                    let decodable = try decoder.decode(Vessel.self, from: line.data(using: .utf8)!)
                    
                    // The geometry that was decoded from the JSON.
                    let geometry = decodable.geometry
                    
                    // We successfully decoded the vessel JSON so we should
                    // add that vessel as a new observation.
                    continuation.yield(
                        .newObservation(
                            geometry: Point(x: geometry.x, y: geometry.y, spatialReference: .wgs84),
                            attributes: decodable.attributes.makeDictionary()
                        )
                    )
                }
                continuation.finish()
            }
        }
    }
}
```

zkline101: I just used `Task.sleep` in the for await loop when reading the file lines, which delays the observations being sent out which seems like a good middle ground here.

zkline101: Ah, that is exactly what you did too! I just saw your message.

Samples.xcodeproj/project.pbxproj
```diff
@@ -742,13 +755,14 @@
...
-				1C965C4629DBA879002F8536 /* 681d6f7694644709a7c830ec57a2d72b */,
-				00D4EF8128638BF100B9CC30 /* cb1b20748a9f4d128dad8a87244e3e37 */,
+				792222DB2A81AA5D00619FFE /* a8a942c228af4fac96baa78ad60f511f */,
 				00C94A0228B53DCC004E42D9 /* 7c4c679ab06a4df19dc497f577f111bd */,
+				D701D7242A37C7E4006FF0C8 /* 07d62a792ab6496d9b772a24efea45d0 */,
 				00D4EF8328638BF100B9CC30 /* 15a7cbd3af1e47cfa5d2c6b93dc44fc2 */,
 				00D4EF8E28638BF100B9CC30 /* 68ec42517cdd439e81b036210483e8e7 */,
+				1C965C4629DBA879002F8536 /* 681d6f7694644709a7c830ec57a2d72b */,
+				00D4EF8128638BF100B9CC30 /* cb1b20748a9f4d128dad8a87244e3e37 */,
+				D77572AC2A295DC100F490CD /* d1453556d91e46dea191c20c398b82cd */,
```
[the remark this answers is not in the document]
zkline101: Note sure, I might have clicked on sort by name in this folder.

Samples.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
zkline101: Probably from using one of the daily swift packages.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
zkline101: > Instead of turning the vessel's attributes into a dictionary, consider just storing the vessel attributes as a value in the dictionary:


Nice idea but we need to turn the attributes into a dictionary either way because we need to pass that to `CustomDynamicEntityFeedEvent.newObservation(geometry:attributes)`

Shared/Samples/Add custom dynamic entity data source/Vessel.swift
zkline101: Ok, I removed globalID.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+    /// A marine vessel that can be decoded from the vessel JSON.
+    struct Vessel {
+        /// The location of the vessel.
+        let geometry: (x: Double, y: Double)
+        /// The attributes of the vessel.
+        let attributes: [String: Any]
+    }
+}
+extension AddCustomDynamicEntityDataSourceView.Vessel: Decodable {
+    private enum CodingKeys: CodingKey {
```
yo1995, line 25: To my understanding an enum's name should be singular, option set name can be plural?

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
philium: Not if you store the vessel's attributes (a value of `Vessel.Attributes`) in the dictionary like I suggested.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+            let attributes = try container.decode(Attributes.self, forKey: .attributes)
+            return [
+                Attributes.CodingKeys.mmsi.rawValue: attributes.mmsi,
+                Attributes.CodingKeys.sog.rawValue: attributes.sog,
+                Attributes.CodingKeys.cog.rawValue: attributes.cog,
+                Attributes.CodingKeys.vesselName.rawValue: attributes.vesselName,
+                Attributes.CodingKeys.callSign.rawValue: attributes.callSign
+            ]
+        }()
+        self.init(geometry: (geometry.x, geometry.y), attributes: attributes)
```
yo1995, line 69: Maybe get it flattened?
```suggestion
        let attributes = try container.decode(Attributes.self, forKey: .attributes)
        
        let attributesDictionary: [String: Any] = [
            Attributes.CodingKeys.mmsi.rawValue: attributes.mmsi,
            Attributes.CodingKeys.sog.rawValue: attributes.sog,
            Attributes.CodingKeys.cog.rawValue: attributes.cog,
            Attributes.CodingKeys.vesselName.rawValue: attributes.vesselName,
            Attributes.CodingKeys.callSign.rawValue: attributes.callSign
        ]
        
        self.init(geometry: (geometry.x, geometry.y), attributes: attributesDictionary)
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct AddCustomDynamicEntityDataSourceView: View {
+    /// A map with an ArcGIS oceans basemap style.
+    @State var map: Map = {
```
yo1995, line 20:
```suggestion
    @State private var map: Map = {
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+        map.initialViewpoint = Viewpoint(
+            latitude: 47.984,
+            longitude: -123.657,
+            scale: 3e6
+        )
+        return map
+    }()
+    
+    /// The dynamic entity layer that is displaying our custom data.
+    @State var dynamicEntityLayer: DynamicEntityLayer
```
yo1995, line 31: This can be a computed property, and the state variable initialization below can be removed.
```suggestion
    private var dynamicEntityLayer: DynamicEntityLayer {
        map.operationalLayers.first as! DynamicEntityLayer
    }
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+    
+    /// The point on the screen the user tapped.
+    @State private var tappedScreenPoint: CGPoint?
+    
+    /// The placement of the callout.
+    @State private var calloutPlacement: CalloutPlacement?
+    
+    init() {
```
yo1995, line 39: Extract the logic of creating a custom dynamic entity layer from the initializer to a separate static method (`private static func makeDynamicEntityLayer() -> DynamicEntityLayer`), and/or add step by step comments, may be beneficial for a user.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+                    
+                    // The geometry that was decoded from the JSON.
+                    let geometry = decodable.geometry
+                    
+                    // We successfully decoded the vessel JSON so we should
+                    // add that vessel as a new observation.
+                    continuation.yield(.newObservation(
+                        geometry: Point(x: geometry.x, y: geometry.y, spatialReference: .wgs84),
+                        attributes: decodable.attributes
+                    ))
```
yo1995, line 130: Some naming suggestions:

- decodable > vessel
- geometry > location

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+    }
+}
+
+private extension Array<Field> {
+    /// An array of fields that match the attributes of each observation in the data source.
+    ///
+    /// This schema is derived from the first row in the custom data source.
+    static var vesselFields: Self {
```
yo1995, line 145: Maybe move it to where the data source info is created, to make it more obvious as one of the steps to create the data source.

Shared/Samples/Add custom dynamic entity data source/README.md
```diff
@@ -0,0 +1,40 @@
...
+* CustomDynamicEntityFeedEvent
+* DynamicEntity
+* DynamicEntityDataSource
+* DynamicEntityLayer
+* LabelDefinition
+* TrackDisplayProperties
+
+## About the data
+
+This sample uses a [JSON file containing observations of marine vessels in the Pacific North West](https://www.arcgis.com/home/item.html?id=a8a942c228af4fac96baa78ad60f511f) hosted on ArcGIS Online.
```
yo1995, line 36:
```suggestion
This sample uses a [JSON Lines file containing observations of marine vessels in the Pacific North West](https://www.arcgis.com/home/item.html?id=a8a942c228af4fac96baa78ad60f511f) hosted on ArcGIS Online.
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
zkline101: The name comes directly from [apple doc](https://developer.apple.com/documentation/foundation/archives_and_serialization/encoding_and_decoding_custom_types):

> Codable types can declare a special nested enumeration named CodingKeys that conforms to the [CodingKey](https://developer.apple.com/documentation/swift/codingkey) protocol.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+                case vesselName = "VesselName"
+                case callSign = "CallSign"
+            }
+            
+            let mmsi: String
+            let sog: Double
+            let cog: Double
+            let vesselName: String
+            let callSign: String
+        }
```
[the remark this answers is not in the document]
zkline101: I think I'm confused about the question. They should get redeclared for each entry of a Vessel because each line in JSON that we decode is a separate vessel.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
philium: This enum is different from other enums in that all cases will be used. There may not be a corresponding value for them in the decoded/encoded data, but these are all the coding keys, so a plural makes sense whereas it doesn't for most other enums.

yo1995: Dang, I forgot the details of doing custom coding!

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+extension AddCustomDynamicEntityDataSourceView {
+    /// A marine vessel that can be decoded from the vessel JSON.
+    struct Vessel {
+        /// The location of the vessel.
+        let geometry: (x: Double, y: Double)
```
philium, line 19: Why was this changed to a tuple?

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+                        attributes: decodable.attributes
+                    ))
+                }
+                
+                continuation.finish()
+            } catch {
+                continuation.finish(throwing: error)
+            }
+        }
+    }
```
philium, line 138: This is much more complicated than it needs to be.

```suggestion
    let events = URL.selectedVesselsDataSource.lines.map { line in
        // Delay observations to simulate live data.
        try await Task.sleep(nanoseconds: 10_000_000)
        
        let decoder = JSONDecoder()
        let vessel = try decoder.decode(
            AddCustomDynamicEntityDataSourceView.Vessel.self,
            from: line.data(using: .utf8)!
        )
        
        // The geometry that was decoded from the JSON.
        let geometry = vessel.geometry
        
        // We successfully decoded the vessel JSON so we should
        // add that vessel as a new observation.
        return CustomDynamicEntityFeedEvent.newObservation(
            geometry: Point(x: geometry.x, y: geometry.y, spatialReference: .wgs84),
            attributes: vessel.attributes
        )
    }
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+                            tolerance: 2
+                          ) else {
+                        // Hides the callout.
+                        calloutPlacement = nil
+                        
+                        return
+                    }
+                    
+                    // Set the callout placement to the observation that was tapped on.
+                    calloutPlacement = identifyResult.geoElements.first.map { .geoElement($0) }
```
philium, line 98: Never use a guard statement when an if…else will do. Else clauses of guard statements shouldn't do any work. Since forward progress is being made in this case, we should use an if…else and enforce that `calloutPlacement` is always updated.

```suggestion
                    let newCalloutPlacement: CalloutPlacement?
                    if let tappedScreenPoint,
                       let identifyResult = try? await proxy.identify(
                        on: dynamicEntityLayer,
                        screenPoint: tappedScreenPoint,
                        tolerance: 2
                       ) {
                        // Set the callout placement to the observation that was tapped on.
                        newCalloutPlacement = identifyResult.geoElements.first.map { .geoElement($0) }
                    } else {
                        // Hides the callout.
                        newCalloutPlacement = nil
                    }
                    calloutPlacement = newCalloutPlacement
```

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+/// The vessel feed that is emitting custom dynamic entity events.
+private struct VesselFeed: CustomDynamicEntityFeed {
+    let events: AsyncThrowingStream<CustomDynamicEntityFeedEvent, Error> = .init { continuation in
+        Task.detached {
+            do {
+                let fileHandle = try FileHandle(forReadingFrom: .selectedVesselsDataSource)
+                let decoder = JSONDecoder()
```
philium, line 110: FYI: Apple has said before the coders are not thread safe and therefore shouldn't be re-used in asynchronous contexts.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
zkline101: I did that so the decodable `Geometry` type could remain an implementation detail.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
zkline101: In this code, we are re-creating the decoder each time the map closure is called, that doesn't seem necessary. Thoughts?

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
zkline101: What's the benefit?

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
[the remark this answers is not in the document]
zkline101: Oh, I see what you mean now. You mean we are declaring these structs each time a line is decoded. hmm

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
zkline101: I like this implementation of the events async sequence much more though.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
philium: It wasn't an implementation detail before because it was actually being used outside of the implementation. So there is no reason that I can see to make it a private implementation detail. I only suggested that for `Attributes` because the individual attributes themselves weren't being used outside of the implementation. But as I got further in my review, I realized that converting the attributes to a dictionary isn't even necessary. I probably should have gone back and revised my earlier comments.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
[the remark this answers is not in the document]
philium: There's nothing wrong with that.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
philium: > we are re-creating the decoder each time the map closure is called

That is intentional and appropriate, as I said in my comment [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/243#discussion_r1296174957). See [this](https://forums.swift.org/t/is-it-ok-to-create-jsondecoder-only-once-in-withthrowingtaskgroup/57260/3) forum post for more information.

zkline101: Great info, thank you.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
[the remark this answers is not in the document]
yo1995: I agree, they happen at compile time so shouldn't take extra time at run time. Just a styling choice.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+extension AddCustomDynamicEntityDataSourceView.Vessel: Decodable {
+    /// A geometry that gives the location of the vessel.
+    struct Geometry: Decodable {
+        /// The x coordinate of the geometry.
+        let x: Double
+        /// The y coordinate of the geometry.
+        let y: Double
+    }
```
philium, line 31: This should go up in the `Vessel` declaration itself, not in the extension that adds `Decodable` conformance.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    init(from decoder: Decoder) throws {
+        /// The attributes that define meta data for the vessel.
+        struct Attributes: Decodable {
+            enum CodingKeys: String, CodingKey { // swiftlint:disable:this nesting
+                case mmsi = "MMSI"
+                case sog = "SOG"
+                case cog = "COG"
+                case vesselName = "VesselName"
+                case callSign = "CallSign"
```
philium, line 46: Since this is a _really_ private type there's not a whole lot of benefit to following the API design guidelines regarding naming. You could just use the JSON key names as the vessel properties names, removing the need for this enum.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
philium: Yes, doing things in the initializer of the view is definitely not the way to go…using a model would almost be preferred. Not that I'm suggesting that, but the way it is now, if the parent's body were to be re-computed, then this initializer would be called and each of these objects would be re-created. The map should be created in a function (`makeMap()`) that is called to initialize the `map` property and then the layer could just be retrieved from the operational layers using a computed property. Or, if you really want to, go back to using a model.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,148 @@
...
+                }
+        }
+    }
+}
+
+/// The vessel feed that is emitting custom dynamic entity events.
+private struct VesselFeed: CustomDynamicEntityFeed {
+    typealias Events = AsyncThrowingMapSequence<AsyncLineSequence<URL.AsyncBytes>, CustomDynamicEntityFeedEvent>
+    
+    let events = URL.selectedVesselsDataSource.lines.map { line in
```
philium, line 118: `Events` isn't needed as it is inferred by the compiler from the type of `events`.

```suggestion
private struct VesselFeed: CustomDynamicEntityFeed {
    let events = URL.selectedVesselsDataSource.lines.map { line in
```

zkline101: Yeah, I tried that and I was getting a compile error saying `VesselFeed` did not conform to `CustomDynamicEntityFeed`.

philium: Which version of Xcode are you using? It works fine in Xcode 14.1.

zkline101: Welp, just kidding I tried again and the compile error went away.

zkline101: I was using 14.3.1 and the compile error would not go away but it did now so all good.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
zkline101: True but I don't see a huge benefit to removing the enum. I would prefer to keep the camel case conventions of variable names intact.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
zkline101: > Not that I'm suggesting that, but the way it is now, if the parent's body were to be re-computed, then this initializer would be called and each of these objects would be re-created. 

Good point.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
zkline101: Based on this [comment](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/243#discussion_r1296506907), it makes since to remove the init so I'll implement the suggestion over creating a model object.

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
philium: I was trying hard to get the name of the variable to match the name of the parameter. In hindsight, I could have used `decodedAttributes` for the `Attributes` value and `attributes` for the dictionary value.

philium (approved):

yo1995 (approved):

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.swift
```diff
@@ -0,0 +1,150 @@
...
+        
+        map.addOperationalLayer(dynamicEntityLayer)
+        
+        return map
+    }
+}
+
+/// The vessel feed that is emitting custom dynamic entity events.
+private struct VesselFeed: CustomDynamicEntityFeed {
+    let events = URL.selectedVesselsDataSource.lines.map { line in
```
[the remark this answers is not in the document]
yo1995: Maybe is this https://github.com/apple/swift/blob/main/stdlib/public/Concurrency/AsyncThrowingMapSequence.swift#L94C41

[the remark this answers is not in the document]
zkline101: Yep! There it is line 94: https://github.com/apple/swift/blob/c38b9b1e08f01c05089de09c42ca8a9b7d1209bb/stdlib/public/Concurrency/AsyncThrowingMapSequence.swift#L94C41 !

Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
```diff
@@ -0,0 +1,71 @@
...
+            let y: Double
+        }
+        
+        /// The location of the vessel.
+        let geometry: Geometry
+        /// The attributes of the vessel.
+        let attributes: [String: Any]
+    }
+}
+extension AddCustomDynamicEntityDataSourceView.Vessel: Decodable {
```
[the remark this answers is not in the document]
zkline101: Whoops good catch, fixed!

yo1995 (approved):

philium (approved):

zkline101:
Thank you both for the great reviews!

PR #245, opened 2023-08-10
[Fix] Xcode project file

philium:
Xcode automatically makes this change whenever the project file is saved.


yo1995:
Thanks. I planned to fix this as part of #228, OK to fix it separately.

philium:
> OK to fix it separately.

I would argue that it is best to fix it separately and as soon as it is discovered. Since it is a bug in the main development branch, it will be propagated to any feature branch created until it is fixed. For example, it showed up in #243 too. Addressing it in a feature PR means that the fix won't get merged until the feature itself is approved and merged, which could take some time.

PR #247, opened 2023-08-15
[Update] Floating Panel usage

dfeinzimer:
## Description

This PR updates usage of the Floating Panel across three samples in response to changes in [Toolkit #404](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/404)

## How To Test

Test the 3 changed samples on Catalyst, iPhone and iPad.

### DIFF
--- a/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
+++ b/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
@@ -60,20 +60,14 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                 .task(id: identifyScreenPoint) {
                     guard let identifyScreenPoint,
                           let layer,
-                          let identifyResult = await Result(awaiting: {
-                              try await proxy.identify(
-                                on: layer,
-                                screenPoint: identifyScreenPoint,
-                                tolerance: 3
-                              )
-                          })
-                        .cancellationToNil()
-                    else {
-                        return
-                    }
-                    
+                          let identifyResult = try? await proxy.identify(
+                            on: layer,
+                            screenPoint: identifyScreenPoint,
+                            tolerance: 3
+                          )
+                    else { return }
                     self.identifyScreenPoint = nil
-                    self.popup = try? identifyResult.get().popups.first
+                    self.popup = identifyResult.popups.first
                     self.showsPopup = self.popup != nil
                 }
                 .floatingPanel(
--- a/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
+++ b/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
@@ -66,7 +66,7 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                             tolerance: 3
                           )
                     else { return }
-                    self.identifyScreenPoint = nil
+                    defer { self.identifyScreenPoint = nil }
                     self.popup = identifyResult.popups.first
                     self.showsPopup = self.popup != nil
                 }
--- a/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
+++ b/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
@@ -60,20 +60,15 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                 .task(id: identifyScreenPoint) {
                     guard let identifyScreenPoint,
                           let layer,
-                          let identifyResult = await Result(awaiting: {
-                              try await proxy.identify(
-                                on: layer,
-                                screenPoint: identifyScreenPoint,
-                                tolerance: 3
-                              )
-                          })
-                        .cancellationToNil()
+                          let identifyResult = try? await proxy.identify(
+                            on: layer,
+                            screenPoint: identifyScreenPoint,
+                            tolerance: 3
+                          )
                     else {
                         return
                     }
-                    
-                    self.identifyScreenPoint = nil
-                    self.popup = try? identifyResult.get().popups.first
+                    self.popup = identifyResult.popups.first
                     self.showsPopup = self.popup != nil
                 }
                 .floatingPanel(
--- a/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
+++ b/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
@@ -66,7 +66,6 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                             tolerance: 3
                           )
                     else { return }
-                    defer { self.identifyScreenPoint = nil }
                     self.popup = identifyResult.popups.first
                     self.showsPopup = self.popup != nil
                 }
@@ -74,7 +73,7 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                     selectedDetent: .constant(.half),
                     horizontalAlignment: .leading,
                     isPresented: $showsPopup
-                ) {
+                ) { [popup] in
                     PopupView(popup: popup!, isPresented: $showsPopup)
                         .showCloseButton(true)
                         .padding()
--- a/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
+++ b/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
@@ -60,15 +60,20 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                 .task(id: identifyScreenPoint) {
                     guard let identifyScreenPoint,
                           let layer,
-                          let identifyResult = try? await proxy.identify(
-                            on: layer,
-                            screenPoint: identifyScreenPoint,
-                            tolerance: 3
-                          )
+                          let identifyResult = await Result(awaiting: {
+                              try await proxy.identify(
+                                on: layer,
+                                screenPoint: identifyScreenPoint,
+                                tolerance: 3
+                              )
+                          })
+                        .cancellationToNil()
                     else {
                         return
                     }
-                    self.popup = identifyResult.popups.first
+                    
+                    self.identifyScreenPoint = nil
+                    self.popup = try? identifyResult.get().popups.first
                     self.showsPopup = self.popup != nil
                 }
                 .floatingPanel(
--- a/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
+++ b/Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
@@ -60,28 +60,30 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
                 .task(id: identifyScreenPoint) {
                     guard let identifyScreenPoint,
                           let layer,
-                          let identifyResult = try? await proxy.identify(
-                            on: layer,
-                            screenPoint: identifyScreenPoint,
-                            tolerance: 3
-                          ),
-                          let firstPopup = identifyResult.popups.first
-                    else { return }
-                    self.popup = firstPopup
-                    self.showsPopup = true
+                          let identifyResult = await Result(awaiting: {
+                              try await proxy.identify(
+                                on: layer,
+                                screenPoint: identifyScreenPoint,
+                                tolerance: 3
+                              )
+                          })
+                        .cancellationToNil()
+                    else {
+                        return
+                    }
+                    
+                    self.identifyScreenPoint = nil
+                    self.popup = try? identifyResult.get().popups.first
+                    self.showsPopup = self.popup != nil
                 }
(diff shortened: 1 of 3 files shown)


dfeinzimer:
@yo1995 will this need a 2nd reviewer?

yo1995:
I'll add Zach to take a look

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
```diff
@@ -60,28 +60,30 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
...
-                    self.popup = firstPopup
-                    self.showsPopup = true
+                          let identifyResult = await Result(awaiting: {
+                              try await proxy.identify(
+                                on: layer,
+                                screenPoint: identifyScreenPoint,
+                                tolerance: 3
+                              )
+                          })
+                        .cancellationToNil()
```
yo1995, line 70: Why wrap the result in another result?

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
```diff
@@ -60,28 +60,30 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
...
+                    else {
+                        return
+                    }
+                    
+                    self.identifyScreenPoint = nil
```
yo1995, line 75: Why explicitly reset the identify point?

yo1995 (approved):
Changes work for me.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
```diff
@@ -60,28 +60,30 @@ struct DisplayPointsUsingClusteringFeatureReductionView: View {
...
                     isPresented: $showsPopup
                 ) {
-                    Group {
-                        if let popup = popup {
-                            PopupView(popup: popup, isPresented: $showsPopup)
-                                .showCloseButton(true)
-                        }
-                    }
-                    .padding()
+                    PopupView(popup: popup!, isPresented: $showsPopup)
```
zkline101, line 84: The app crashes for me here because we are force unwrapping a `nil` popup.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
zkline101: It looks like it is for when the result is a failure and the error is a cancellation error then the result should be `nil`. But I am also not sure what the use case is for that.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
dfeinzimer: Did you get this working? Should be okay on the Toolkit's `v.next` now.

zkline101: Yep, all good.

zkline101 (approved):
Approving but I would like to see the answers to Ting's questions.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
dfeinzimer: I added these Result wrappers because the FloatingPanel changes were working fine in the Toolkit's Popup example but not in these samples. `popup` was still `nil` at the time of the force unwrap. 

The missing Result wrappers was the main difference I spotted between the two implementations.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
dfeinzimer: This is also done in the Popup's example code. It's not 100% necessary so I'm happy to remove it if you prefer.

yo1995: No need to change.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
yo1995: That's kind of strange. I don't expect the popup state property to be changed until the first identify operation runs. 🤔

zkline101: > That's kind of strange. I don't expect the popup state property to be changed until the first identify operation runs.

Agreed.

dfeinzimer: No, that crashes for me. Does it work on your side?

yo1995: 🤔

dfeinzimer: Did some more experimentation, appears the `self.identifyScreenPoint = nil` prevents the crash.

zkline101 (dismissed):

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
yo1995: Me too confused by where the problem comes from. Maybe there is some magic with SwiftUI that caused the `PopupView` to be computed too early when the popup isn't non `nil` yet.

yo1995 (dismissed):

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
dfeinzimer: @philium I'm curious if you see anything we might be missing here. Removing the following line in this branch will cause a crash upon selecting a feature.

https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/9d2e3b7d6080595e04da06777420852ee7714525/Shared/Samples/Display%20points%20using%20clustering%20feature%20reduction/DisplayPointsUsingClusteringFeatureReductionView.swift#L69

philium: Basically what is happening is that the floating panel's content is stale—an older version of the content closure is being evaluated. Why that is, I don't know. I do have a fix, but unfortunately it is in the toolkit, so you'll have to go with your workaround for now.

philium: Never mind. I thought I had a fix, but apparently not.

philium: For now, why not just conditionally unwrap the popup?

```swift
if let popup {
    // Create popup view.
}
```

It seems like that would be more straightforward than setting `identifyScreenPoint` to `nil`.

dfeinzimer: @njarecha identified the issue during workshop this afternoon.

`identifyScreenPoint` which is of type `CGPoint` is being used to start the task. `CGPoint` doesn't conform to Equatable. 

The documentation of the task modifier's `id` parameter states:

> The value to observe for changes. The value must conform to the Equatable protocol.

Because of this, we need to manually reset `identifyScreenPoint` to avoid undefined behavior. Nimesh also suggested this reset should be `defer`red.

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
dfeinzimer: Final explanation is [above](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/247#discussion_r1297741574).

Shared/Samples/Display points using clustering feature reduction/DisplayPointsUsingClusteringFeatureReductionView.swift
yo1995: That's quite surprising as well, as one of the [deprecated doc](https://developer.apple.com/documentation/coregraphics/1456179-cgpointequaltopoint) for CGPoint says

> The [CGPoint](https://developer.apple.com/documentation/corefoundation/cgpoint) type adopts the Equatable protocol

Anyway, great to learn the root cause and potentially something we can put into the onTapGesture's discussion. Thanks!

philium: > `CGPoint` doesn't conform to Equatable.

Though it doesn't show up in the doc as an explicit conformance, `CGPoint` most definitely does conform to `Equatable`, as seen in [this doc](https://developer.apple.com/documentation/coregraphics/1456179-cgpointequaltopoint):

image

Not to mention that if it didn't, the compiler would throw an error because that constraint of the `id` parameter is enforced by the compiler.

> should be `defer`red.

A [`defer` statement](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Defer-Statement) ensures that whatever is inside the block is executed "just before transferring program control outside of the scope that the `defer` statement appears in," whichever way that is done. In this case, that happens as soon as the end of the `action` closure of the task modifier is reached. In other words, there's no difference between using a defer statement and moving the contents of the defer statement to end of the closure. And since the order of setting the state properties doesn't matter, using `defer` here is essentially pointless.

dfeinzimer: Hm, this is interesting, so do you think this isn't truly the root cause? This line still makes the difference between crashing or not. Is there possibly a semantic difference between "adopts" and "conforms"? 

I often check the "Conforms To" section in the doc when looking for conformance so I'm a bit disappointed that it doesn't show here if it does conform to it.

Screenshot 2023-08-17 at 14 24 50

dfeinzimer: Note that I've also tried swapping out the Floating Panel for a sheet here and a sheet suffers the same crash if the point isn't reset.

philium: I'll do a little more investigating. I have another idea.

dfeinzimer: I was able to write up a pretty compact reproducer.

<details>

```swift
import SwiftUI

struct ContentView: View {
    @State private var point: CGPoint?
    
    @State private var isPresented = false
    
    @State private var content: String?
    
    var body: some View {
        Text("Hello, world!")
            .onTapGesture {
                point = CGPoint(
                    x: Double(Int.random(in: 100...400)),
                    y: Double(Int.random(in: 500...900))
                )
            }
            .task(id: point) {
                guard point != nil else { return }
                point = nil                // <---- Will crash without this line commented
                content = "Some message"
                isPresented = true
            }
            .sheet(isPresented: $isPresented) {
                Text(content!)
            }
            .padding()
    }
}
```

</details>

njarecha: Found the explanation in this [thread](https://stackoverflow.com/questions/74554151/task-in-swiftui-runs-but-view-that-is-inside-a-sheet-is-not-updated).

If you change the `floatingPanel`'s closure  to capture the `popup` then it works as expected without setting `identifyScreenPoint = nil`.

```
                .floatingPanel(
                    selectedDetent: .constant(.half),
                    horizontalAlignment: .leading,
                    isPresented: $showsPopup
                ) { [popup] in
                    PopupView(popup: popup!, isPresented: $showsPopup)
                        .showCloseButton(true)
                        .padding()
                }
```

philium: It would be great if there was a version of the `floatingPanel` modifier that took an optional item as a data source, presenting the panel when the item is non-nil and providing it as a non-optional to the content closure, similar to [`sheet(item:onDismiss:content:)`](https://developer.apple.com/documentation/swiftui/view/sheet(item:ondismiss:content:)).

dfeinzimer: > It would be great if there was a version of the floatingPanel modifier that took an optional item as a data source, presenting the panel when the item is non-nil and providing it as a non-optional to the content closure, similar to [sheet(item:onDismiss:content:)](https://developer.apple.com/documentation/swiftui/view/sheet(item:ondismiss:content:)).

I've been thinking about that this week. I'll create an issue for it.

dfeinzimer:
Updated to Nimesh's proposed fix, ready for re-review

zkline101 (approved):

yo1995 (approved):

PR #251, opened 2023-08-16
[Fix] Temporary directory creation

yo1995:
## Description

This PR fixes a bug in temporary directory creation in iOS 17 beta on device.

There is a [File system changes introduced in iOS 17](https://developer.apple.com/forums/thread/733642) that puts the app bundle to a different volume on disk than its document/temporary and other data container directories. This made the old approach of getting the URL to the app-specific temporary directory no longer valid.

The [url(for:in:appropriateFor:create:)](https://developer.apple.com/documentation/foundation/filemanager/1407693-url) API specified that…

- `appropriateFor url`: This parameter is ignored unless the directory parameter contains the value [FileManager.SearchPathDirectory.itemReplacementDirectory](https://developer.apple.com/documentation/foundation/filemanager/searchpathdirectory/itemreplacementdirectory) and the domain parameter contains the value [userDomainMask](https://developer.apple.com/documentation/foundation/filemanager/searchpathdomainmask/1408037-userdomainmask).
- `shouldCreate create`: When creating a temporary directory, this parameter is ignored and the directory is always created.

Therefore, when creating the `itemReplacementDirectory` using this method, both parameters are trivial except the `appropriateFor` URL's volume. As suggested by Eskimo, until we move to iOS 16+ when we have access to the newer APIs, we can use the app's document directory to get the correct volume where the temp folder should be created.

## Linked Issue(s)

Please see the details in the Apple developer forum response by Eskimo.

- `swift/issues/4364`
- https://developer.apple.com/forums/thread/735726
- https://developer.apple.com/forums/thread/733642

## How To Test

1. Build and run the sample on device that has installed iOS 17 beta 4 or iOS 17 public beta, or above.
2. Run any of these 4 samples involved, such as `Generate offline map`.
3. Prior to this fix, it would crash with "NSCocoaErrorDomain Code=513 You don’t have permission to save the file"
4. After this fix, no crash will happen.

### DIFF
--- a/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
+++ b/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
@@ -60,7 +60,7 @@ extension DownloadPreplannedMapAreaView {
         
         init() {
             // Creates temp directory.
-            temporaryDirectory = FileManager.makeTemporaryDirectory()
+            temporaryDirectory = FileManager.createTemporaryDirectory()
             
             // Initializes the online map and offline map task.
             onlineMap = Map(item: portalItem)
@@ -299,7 +299,7 @@ private extension OfflineMapModel {
 
 private extension FileManager {
     /// Creates a temporary directory and returns the URL of the created directory.
-    static func makeTemporaryDirectory() -> URL {
+    static func createTemporaryDirectory() -> URL {
         // swiftlint:disable:next force_try
         try! FileManager.default.url(
             for: .itemReplacementDirectory,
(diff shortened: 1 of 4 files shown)


Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
```diff
@@ -299,12 +299,12 @@ private extension OfflineMapModel {
 
 private extension FileManager {
     /// Creates a temporary directory and returns the URL of the created directory.
-    static func createTemporaryDirectory() -> URL {
+    static func makeTemporaryDirectory() -> URL {
```
philium, line 302: We went with `create` because this method creates something (a directory on disk) rather than makes something (a value of some type). The API design guideline doesn't apply here because this isn't a factory method. The old name is more appropriate.

saratchandrakarumuri (approved):
Tested on both iOS16 and 17 devices and the samples with this workflow are now working as expected

PR #252, opened 2023-08-16
[Fix] accessory button and help text color is not clearly visible

des12437:
## Description

Fixes a bug in iOS 16+ where the help text and accessory button are not clearly visible when the list item selection color is the accent color (purple/pink) instead of grey. This PR adds opacity to the help text so that the system sets the text color. The accessory button is now a label with an icon and tap gesture rather than a Button since the system will set the color automatically.

## Linked Issue(s)

- `swift/issues/3878`

## How To Test
- Test on iPhone and iPad on iOS 16 and iOS 17

## Screenshots
Before:

After:

### DIFF
--- a/Shared/Supporting Files/Views/SampleRow.swift
+++ b/Shared/Supporting Files/Views/SampleRow.swift
@@ -40,6 +40,7 @@ struct SampleRow: View {
             Label {} icon: {
                 Image(systemName: "info.circle")
                     .symbolVariant(isShowingDescription ? .fill : .none)
+                    .imageScale(.medium)
             }
             .onTapGesture {
                 isShowingDescription.toggle()
--- a/Shared/Supporting Files/Views/SampleRow.swift
+++ b/Shared/Supporting Files/Views/SampleRow.swift
@@ -32,18 +32,18 @@ struct SampleRow: View {
                 if isShowingDescription {
                     Text(description)
                         .font(.caption)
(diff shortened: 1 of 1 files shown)


saratchandrakarumuri (dismissed):
Tested on both iOS 15 and 16 platforms and it is now rendering as expected on both the platforms

Shared/Supporting Files/Views/SampleRow.swift
```diff
@@ -32,18 +32,18 @@ struct SampleRow: View {
...
                         .transition(.move(edge: .top).combined(with: .opacity))
                 }
             }
             Spacer()
-            Button {
-                isShowingDescription.toggle()
-            } label: {
+            Label {} icon: {
                 Image(systemName: "info.circle")
                     .symbolVariant(isShowingDescription ? .fill : .none)
```
yo1995, line 42:
```suggestion
                    .symbolVariant(isShowingDescription ? .fill : .none)
                    .imageScale(.medium)
```

yo1995 (commented):
Just one suggestion, otherwise looks good.

yo1995 (approved):

yo1995:
@saratchandrakarumuri in case you missed the email notification 😉

saratchandrakarumuri (approved):
Tested again with iOS 16 and iOS 17. This is now rendering as expected

PR #253, opened 2023-08-17
[Update] Adjust info view control size

yo1995:
## Description

This PR adjusts the info view's control position and size.

- Add an ideal width to the segmented picker so it's width doesn't differ among iOS versions
- Add spacer to bottom bar to make the code picker centered

## Linked Issue(s)

See `swift/issues/4377`.

## How To Test

Build and run on iOS 16/17 iPad.

## To Discuss

- Ideally, we don't apply the fixed width frame modifier to the segmented picker. But that would cause the width problem described in issue 4377.
- Another idea is to use `fixedSize()` modifier to make the segmented control tight. While it is commonly seen on iPhone, it is generally wider on iPad and Mac.
- Please share your thoughts on which looks better:
   - segmented picker: tight / fixed width (current) / fill the top bar (original, with bug on iOS 17)
   - code picker: centered (current) / aligned to the right / system default (original)

### DIFF
--- a/Shared/Supporting Files/Views/SampleInfoView.swift
+++ b/Shared/Supporting Files/Views/SampleInfoView.swift
@@ -43,7 +43,9 @@ struct SampleInfoView: View {
                     }
                 }
                 .pickerStyle(.segmented)
-                .frame(idealWidth: 320)
+                // Set the ideal width to a big value so the segmented control
+                // takes up the entire empty space of the toolbar on iOS 17.
+                .frame(idealWidth: UIScreen.main.bounds.width)
             }
             
             ToolbarItem(placement: .confirmationAction) {
--- a/Shared/Supporting Files/Views/SampleInfoView.swift
+++ b/Shared/Supporting Files/Views/SampleInfoView.swift
@@ -43,6 +43,7 @@ struct SampleInfoView: View {
                     }
                 }
                 .pickerStyle(.segmented)
+                .frame(idealWidth: 320)
             }
             
             ToolbarItem(placement: .confirmationAction) {
@@ -51,13 +52,15 @@ struct SampleInfoView: View {
                 }
             }
             
-            ToolbarItem(placement: .bottomBar) {
+            ToolbarItemGroup(placement: .bottomBar) {
                 if informationMode == .code {
+                    Spacer()
                     Picker("Source Code File Picker", selection: $selectedSnippetIndex) {
                         ForEach(sample.snippetURLs.indices, id: \.self) { index in
                             Text(sample.snippets[index].dropLast(6))
                         }
                     }
+                    Spacer()
                 }
             }
         }


des12437:
The wider segmented picker looks better now for iOS 17. Is there a way to match the segmented picker width on iOS 16?

As for the code picker alignment, I prefer the centered alignment change made in this PR over the system default.

Before/After on iPad iOS 17:
       

iPad iOS 16:

yo1995:
> The wider segmented picker looks better now for iOS 17. Is there a way to match the segmented picker width on iOS 16?

For some reason, `.frame(maxWidth: .infinity)` doesn't work here. We can use a big enough number such as `.frame(idealWidth: UIScreen.main.bounds.width)`, but I don't like this solution.

Do you have other ideas? 🤔  To me the segmented control not taking up the entire width is more like an iOS 17 bug, rather than a new behavior.

des12437:
> For some reason, `.frame(maxWidth: .infinity)` doesn't work here. We can use a big enough number such as `.frame(idealWidth: UIScreen.main.bounds.width)`, but I don't like this solution.
> 
> Do you have other ideas? 🤔 To me the segmented control not taking up the entire width is more like an iOS 17 bug, rather than a new behavior.

Not at the moment, I tried a few `.frame()` modifier variants which did not work. The segmented picker looks good using `.frame(idealWidth: UIScreen.main.bounds.width)` but if there is good reason to not use that then the original solution is good.

dfeinzimer (approved):
I think these changes look nice and work across iOS 16 and 17 on the three platforms. The centered code picker looks good too.

des12437 (dismissed):

saratchandrakarumuri (commented):
It now looks similar on both iOS 15 and 16. But on iOS 17 beta, 

- The name of the code file is not using the accent color and just rendering in black.
- Width of the tabs is still different on iOS 17 as it takes only part of the width where as in iOS 16 it spreads to whole width

2023-08-21 at 2 49 PM

yo1995:
With the changes in dbbed15 using a big value in the frame modifier, it now looks like this.

Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2023-08-22 at 13 39 00

saratchandrakarumuri:
@yo1995 I now see that the display width is matching on iOS 17 with rest of the platforms. The name of the code file is still rendering in black color in contrast to the accent colors on other platforms. Is that something out of control or expected?

yo1995:
> The name of the code file is still rendering in black color in contrast to the accent colors on other platforms. Is that something out of control or expected?

That is unexpected. I'm not seeing it on iOS 17 beta Sim, iPad Pro 11 inch 4th gen though. When you build the `v.next` branch, do you also see the button as black?

saratchandrakarumuri:
@yo1995 Yaa... I am seeing it in black... I just noticed that in your screen shot it is rendering in accent color

dg0yal:
Let's get this moving again

des12437:
The info view control size changes look good, however, the bottom bar code picker does not render on iOS 17. Can you verify this? The issue does not seem related to the changes in this PR.

yo1995:
> The info view control size changes look good, however, the bottom bar code picker does not render on iOS 17. Can you verify this? The issue does not seem related to the changes in this PR.

image

I'm using Xcode 15 beta 7

des12437:
> image
> 
> I'm using Xcode 15 beta 7

Okay thanks. There must be an issue with my simulator configuration.

des12437 (approved):

dfeinzimer (approved):
Looks good. Better top padding in iOS 17 than 16:

iOS 16:
Simulator Screenshot - iPhone 14 Pro - 2023-09-07 at 15 34 04
iOS 17:
Simulator Screenshot - iPhone 14 Pro - 2023-09-07 at 15 37 02

saratchandrakarumuri (approved):
All the changes are now looking with all iOS versions

saratchandrakarumuri (approved):
All the changes are now looking with all iOS versions

yo1995:
Thank you!

Issue #80, opened 2022-11-18
Fix sample info view presentation auto dismiss when custom sheet modifier is in place


yo1995:
See #77 

I'm leaning towards fixing it when we move up to **iOS 16** by leveraging the new [popover behavior](https://developer.apple.com/documentation/swiftui/view/presentationdetents(_:)) (https://www.hackingwithswift.com/quick-start/swiftui/how-to-show-a-popover-view) that switch between popover and sheet automatically, so we don't need to hack into the `UIViewRepresentable` and navigation controller. 

Thanks for Chris' update in https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/47

Meanwhile, for the remainder time of iOS 15, I want to remove the switching mechanism between screen orientations to alleviate the complexity of conditions in the updateUIView method

https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/b9d5c704d76f862f3404e2cccaf1f67a81d1afc6/Shared/Supporting%20Files/Extensions/View%2BSheet.swift#L282

yo1995:
Also, in #134 the dismiss logic needs to be fixed. See https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/42/files#diff-de5737b47911aa7de83a8916aeab221a5d7c5f73823ba283aafd21360deae875R304

yo1995:
Phil suggested using iOS 16 native popover modifier https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/262 and we would like to add it to the extension now, like in #47 , before we eventually move to iOS 16 min later next year.

yo1995:
This is fixed in #306 with a few minor issues

PR #258, opened 2023-08-23
[Update] Download preplanned map area feedback

des12437:
## Description

This PR addresses feedback from Nick on the initial testing for 200.1.0 sample viewer TestFlight app for the `Download  preplanned map area` sample:
- > Oooh. A legit bug! I dismissed the preplanned area picker in the preplanned sample then navigated back from the preplanned sample to the samples list and the preplanned area picker reappeared over the sample list! I think you have to download a preplanned area to see this.
Might be related to the sheet issue as well: https://github.com/Esri/arcgis-maps-sdk-swift-samples/issues/80

  This PR fixes this bug caused by a sheet selection bug. The sheet can be closed with a "Done" button which calls 'dismiss()' but does not set the selection value to false. The "Done" button now sets the selection value, but to do so the value had to be moved from to view to the model which is not ideal.

- > This map ("Seager Park Area") looks empty by default on an iPhone. I have to zoom in to notice it has content. Not sure what to do about that though…
  >I spotted the same when I review this sample. The best way IMO is to set the initial zoom scale on these preplanned map area to a close enough value so the symbols appear on all platform regardless desktop or mobile.
  
  This issue is fixed by expanding the visible area's envelope when an offline map is selected. Now all layer content is visible for each offline map.

- > Would it be better if, when you download an area, it is automatically selected? Currently you have to tap to download, wait for it to download, and then tap it again to select it.

  I did not implement this suggestion based on discussion in the swift issue deciding to not make this change.

## Linked Issue(s)

- `swift/issues/3931`

## How To Test
- View each offline map after downloading and verify that all layer content is visible.

## Screenshots
Before/After layer visibility

## To Discuss

- Should the selection sheet bug fix be removed from this PR in favor of waiting for the expected fix in https://github.com/Esri/arcgis-maps-sdk-swift-samples/issues/80 once iOS 16 is the minimum deployment target?

### DIFF
--- a/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
+++ b/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
@@ -58,6 +58,12 @@ extension DownloadPreplannedMapAreaView {
         /// A Boolean value indicating if the offline content can be deleted.
         @Published private(set) var canRemoveDownloadedMaps = false
         
+        /// A Boolean value indicating whether to select a map.
+        @Published var isShowingSelectMapView = false
+        
+        /// A Boolean value indicating if the selected offline map just changed.
+        @Published var offlineMapDidChange = false
+        
         init() {
             // Creates temp directory.
             temporaryDirectory = FileManager.createTemporaryDirectory()
@@ -108,6 +114,7 @@ extension DownloadPreplannedMapAreaView {
                 } else if case .success(let mmpk) = info.result {
                     // If we have already downloaded, then open the map in the mmpk.
                     offlineMap = mmpk.maps.first
+                    offlineMapDidChange = true
                 } else {
                     // If we have a failure, then keep the online map selected.
                     selectedMap = oldValue
--- a/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.swift
+++ b/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.swift
@@ -16,9 +16,6 @@ import ArcGIS
 import SwiftUI
 
 struct DownloadPreplannedMapAreaView: View {
-    /// A Boolean value indicating whether to select a map.
-    @State private var isShowingSelectMapView = false
-    
     /// A Boolean value indicating whether to show delete alert.
     @State private var isShowingDeleteAlert = false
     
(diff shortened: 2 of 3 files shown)


Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
```diff
@@ -58,6 +58,12 @@ extension DownloadPreplannedMapAreaView {
         /// A Boolean value indicating if the offline content can be deleted.
         @Published private(set) var canRemoveDownloadedMaps = false
         
+        /// A Boolean value indicating whether to select a map.
+        @Published var isShowingSelectMapView = false
```
yo1995, line 62: > The "Done" button now sets the selection value, but to do so the value had to be moved from to view to the model which is not ideal.

I don't think this problem should be fixed in the view level. It would be better addressed at the root, in `View+Sheet.swift`

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.swift
```diff
@@ -16,53 +16,65 @@ import ArcGIS
...
                     
-                    Button("Select Map") {
-                        isShowingSelectMapView.toggle()
-                    }
-                    .sheet(isPresented: $isShowingSelectMapView, detents: [.medium]) {
-                        MapPicker()
-                            .environmentObject(model)
-                    }
+                    let builder = EnvelopeBuilder(envelope: visibleArea.extent)
+                    builder.expand(by: 0.5)
```
yo1995, line 33: While this works, I think what Nick originally proposed is more appealing to me. We can set the initial viewpoint to a closer scale before the map is displayed. Maybe a private extension like Mark has done in the toolkit: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/286

des12437: I applied this suggestion to set the initial viewpoint so that the map scale is zoomed in for each map, including the map initially presented rather than just the offline maps:

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
des12437: I fixed the dismiss sheet bug by adding back the logic from [this](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/42/files#diff-de5737b47911aa7de83a8916aeab221a5d7c5f73823ba283aafd21360deae875R304) PR. Do you know why that logic was removed? It's necessary to set the sheet's `isSelected` property to `false` when dismissing the sheet.

yo1995: I'm not sure. Maybe it fell through the cracks and wasn't merged when the sample was refactored by Ryan.

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.swift
```diff
@@ -26,7 +26,7 @@ struct DownloadPreplannedMapAreaView: View {
     @StateObject private var model = Model()
     
     var body: some View {
-        MapView(map: model.currentMap)
+        MapView(map: model.currentMap,viewpoint: model.currentMap.initialViewpoint?.zoomedIn())
```
yo1995, line 29:
```suggestion
        MapView(map: model.currentMap, viewpoint: model.currentMap.initialViewpoint?.zoomedIn())
```

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.swift
```diff
@@ -73,3 +73,12 @@ struct DownloadPreplannedMapAreaView: View {
             .background(.thinMaterial, ignoresSafeAreaEdges: .horizontal)
     }
 }
+
+private extension Viewpoint {
```
yo1995, line 77: Probably keep this consistent with `private extension Envelope` in other samples.

des12437: If the viewpoint is made a static method/property then `Viewpoint.targetGeometry` is not available. 
Screenshot 2023-09-06 at 1 38 36 PM

yo1995: OK. Add a docstring to this method then.

yo1995 (approved):

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.swift
des12437: I added the docstring and renamed the method.

yo1995 (approved):

nixta:
If I open the on-demand sample then the preplanned sample, I don't see any controls below the map.

Oddly, switching between samples got the controls to show up once, but now I can't get them to show up at all.

IMG_F20F6EA14C4B-1

des12437:
> If I open the on-demand sample then the preplanned sample, I don't see any controls below the map.
> 
> Oddly, switching between samples got the controls to show up once, but now I can't get them to show up at all.

I tried to reproduce this on iOS 16 and 17 but could not. What iOS version and simulator did you see this issue on? Also which on-demand sample did you open first?

nixta:
It's really inconsistent. I got unlucky and saw it first time I ran the app, but after that it took quite some effort to see it again. Just search for "Offline" then switch between Generate and Preplanned. Here's a video. I switched back and forth a lot before I could repro for the video, and you'll notice that I can just exit the sample and go right back in and it works fine again.

https://github.com/Esri/arcgis-maps-sdk-swift-samples/assets/25374/161203c7-cf21-4fa1-a332-1b0d92cbf76b

Sorry I can't get a better repro!

nixta (approved):
Aside from the weirdness of the bottom UX not appearing on the odd occasion, this looks much better. Thank you!

I'll let someone else make the call about whether that should be a show-stopper. Given how hard it's been for me to repro, I'd say probably not. We could open another issue for that.

des12437:
> Aside from the weirdness of the bottom UX not appearing on the odd occasion, this looks much better. Thank you!
> 
> I'll let someone else make the call about whether that should be a show-stopper. Given how hard it's been for me to repro, I'd say probably not. We could open another issue for that.

I will log a separate issue for the UI bug, thank you for finding that!

PR #248, opened 2023-08-15
[Fix] Add status message and fix UI for `Trace utility network`

yo1995:
## Description

This PR

- Add a status message for potentially failed/nil method
- Fix the UI on iOS 17 for bottom bar. It is defaulted to the left edge without spacers.

## Linked Issue(s)

- `swift/issues/4401`

## How To Test

Run the sample on iOS 17 beta, UI looks good.

PR #259, opened 2023-08-23
[Fix] "false" utility network load problem

yo1995:
## Description

This PR fixes an issue in the `Trace utility network` sample.

In the original PR, when the sample opens, a computed property `private var network: UtilityNetwork? { map.utilityNetworks.first }` tries to get the first utility network from the web map. However, because the web map is not explicitly loaded, `network` can become `nil` occasionally, depending on the load timing of the map.

This leads to the problem in 4401 where the `network?.makeElement(arcGISFeature: feature)` call silently fails. A status message was added in #248 to show the error.

After discussing with Jen, both the web map and the network need to be loaded. Only loading the map is not enough; only loading the network doesn't guarantee the map to be present either. Once the map is loaded, we can safely assume the utility network's description is available.

## Linked Issue(s)

- `swift/issues/4401`
- https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/130#discussion_r1147978503

## How To Test

- On `v.next`, sometimes you cannot add element to the utility network
- With the fixes on this branch, the sample is correct

### DIFF
--- a/Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
+++ b/Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
@@ -22,7 +22,7 @@ extension TraceUtilityNetworkView {
         
         /// The domain network for this sample.
         private var electricDistribution: UtilityDomainNetwork? {
-            network?.definition?.domainNetwork(named: "ElectricDistribution")
+            network.definition?.domainNetwork(named: "ElectricDistribution")
         }
         
         /// The URLs of the relevant feature layers for this sample.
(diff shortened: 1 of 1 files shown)


dfeinzimer (approved):

dg0yal:
let's get this moving

saratchandrakarumuri (approved):
Tested this multiple times with the fix and identify is working as expected on UN.

saratchandrakarumuri (approved):
Tested this multiple times with the fix and identify is working as expected on UN.

yo1995:
Thank you!

PR #262, opened 2023-08-31
Add dynamic entity layer settings presentation refactor

philium:
@des12437 I played around with things for a bit and this seems to work on iPhone (portrait and landscape), iPad, and macOS. Please take a look and tell me what you think. Note that the `popover` modifier is adaptive, meaning that it will use a sheet in compact width environments.


des12437 (approved):
Looks good on all devices. This approach is preferable over checking the size classes.

PR #272, opened 2023-09-21
[New] Style symbols from mobile style file

CalebRas:
## Description

This PR implements `Style symbols from mobile style file` in `Visualization` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-StyleSymbolsFromMobileStyleFile/Shared/Samples/Style%20symbols%20from%20mobile%20style%20file)

## Linked Issue(s)

- `swift/issues/4072`

## How To Test

- Tap "Symbol" create an emoji face by selecting a symbol for each category and adjusting the color/size.
- Tap on the map to add the emoji to map.
- Tap "Clear" to remove all the emojis from the map.

## Screenshots

style-symbols-from-mobile-style-file-1 style-symbols-from-mobile-style-file-2


Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
```diff
@@ -0,0 +1,121 @@
...
+                // Update the symbols when the display scale changes.
+                symbols = await getSymbols(displayScale: displayScale)
+                await model.updateCurrentSymbol(displayScale: displayScale)
+            }
+        }
+        
+        /// Gets the symbols from the symbol style.
+        /// - Parameter displayScale: The display scale of the environment for creating symbol swatches.
+        /// - Returns:  An `Array` of `SymbolDetails` which contain a symbol and its associated information.
+        private func getSymbols(displayScale: Double) async -> [SymbolDetails] {
```
rolson, line 80: if any of the `try?` calls fail in this function then the error information is discarded. You might consider making this function `throws` and doing something with the error. That would also simplify a lot of the code in this function, alleviating the need for some guard statements.

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                }
+            }
+            .task(id: model.symbolOptionSelections) {
+                // Update the current symbol when an option selection changes.
+                await model.updateCurrentSymbol(displayScale: displayScale)
+            }
+            .task(id: displayScale) {
+                // Update the symbols when the display scale changes.
+                symbols = await getSymbols(displayScale: displayScale)
```
rolson, line 72: when does the display scale ever change?

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
rolson, line 80: You could think about renaming this
```suggestion
        private func symbols(forDisplayScale: Double) async -> [SymbolDetails] {
```

rolson, line 80: I would consider moving this function to the model. Also storing the symbols on the model could simplify some things. You have already a model, then have state for the symbols. Why not just put the symbols on the model

CalebRas: I originally had the function throw, but I chose to go with `try?` so that if an error is throw in the task group, the function returns the symbols it could load. That way, you still get a list of symbols even if you have a faulty key, which I found is pretty easy to mess up. Also, I tried to mimic [a similar pattern](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/main/Shared/Samples/Display%20content%20of%20utility%20network%20container/DisplayContentOfUtilityNetworkContainerView.Model.swift#L120-L135) I found in other samples. If you think that it would be better to have it throw I can definitely do that here and in `updateCurrentSymbol(displayScale:)`, though.

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
CalebRas: The display scale changes when the app is moved to another screen which has a different resolution, which really can only happen on Mac Catalyst and even then, it is unlikely. I have been including it in the style samples since Ting said it was better for accessibility, and I found the pattern in the [Show utility associations](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/main/Shared/Samples/Show%20utility%20associations/ShowUtilityAssociationsView.swift#L75-L78) and [Display content of utility network container](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/main/Shared/Samples/Display%20content%20of%20utility%20network%20container/DisplayContentOfUtilityNetworkContainerView.swift#L70-L73) samples. Let me know if you want me to remove as it won't make much of a difference.

rolson: Good info, thank you! I would leave it then.

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
CalebRas: I didn't put the symbols on the model since they are only accessed by one view, which is also why I have that function as part of the view and not the model. I was under the impression that it made it more encapsulated for lack of a better word. It may easy to understand/cleaner if it was part of the model, though. Let me know if you think I should move everything to model.

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.swift
```diff
@@ -0,0 +1,190 @@
...
+            .toolbar {
+                ToolbarItemGroup(placement: .bottomBar) {
+                    Button("Symbol") {
+                        isShowingSymbolOptionsSheet = true
+                    }
+                    .sheet(isPresented: $isShowingSymbolOptionsSheet, detents: [.large]) {
+                        symbolOptionsList
+                    }
+                    
+                    Spacer()
```
dfeinzimer, line 44: Not a big deal but would probably give equal spacing on either side of the spacer.

```suggestion
                    
                    Spacer()
                    
```

or

```suggestion
                    Spacer()
```

dfeinzimer (approved):

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
rolson: I don't feel strongly about it. In general I don't think it's great to eat errors and I wouldn't want our users copying code that does that. But that's more of a samples app philosophical discussion that is probably out of scope of this PR.

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.swift
```diff
@@ -0,0 +1,189 @@
...
+        }
+        
+        init(symbol: Symbol, image: UIImage, name: String = "", key: String = "", category: String = "") {
+            self.symbol = symbol
+            self.image = image
+            self.name = name
+            self.key = key
+            self.category = category
+        }
+    }
```
rolson, line 151: You can have the compiler help you here:

```swift
    /// A symbol and its associated information.
    struct SymbolDetails {
        /// The symbol from the symbol style.
        let symbol: Symbol
        /// The image swatch of the symbol.
        let image: UIImage
        /// The name of the symbol as found in the symbol style.
        var name: String = ""
        /// The key of the symbol as found in the symbol style.
        var key: String = ""
        /// The category of the symbol as found in the symbol style.
        var category: String = ""
    }
```

And then add your extension for label:

```swift

extension StyleSymbolsFromMobileStyleFileView.SymbolDetails {
    /// The human-readable label of the symbol name.
    var label: String {
        let splitName = name.replacingOccurrences(of: "-", with: " ").split(separator: " ")
        return splitName.last?.capitalized ?? name
    }
}
```

This has a few benefits:
- using the compiler to synthesize the init
- the struct definition is just the data, the label is not data per se but a function of the data so an extension is appropriate

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.swift
```diff
@@ -0,0 +1,189 @@
...
+        var size: Double
+        
+        init(eyes: String, hat: String, mouth: String, color: Color, size: Double) {
+            categoryKeys[.eyes] = eyes
+            categoryKeys[.hat] = hat
+            categoryKeys[.mouth] = mouth
+            self.color = color
+            self.size = size
+        }
+    }
```
rolson, line 169: categoryKeys does not need to be initialized with `[:]`. Also, can just be assigned once:

```swift
    struct SymbolOptions: Equatable {
        /// A dictionary containing the key of a symbol for each symbol category.
        var categoryKeys: [SymbolCategory: String]
        /// The color of the symbol.
        var color: Color
        /// The size of the symbol.
        var size: Double
        
        init(eyes: String, hat: String, mouth: String, color: Color, size: Double) {
            categoryKeys = [
                .eyes: eyes,
                .hat: hat,
                .mouth: mouth
            ]
            self.color = color
            self.size = size
        }
    }
```

rolson (commented):
Looks good. Couple minor suggestions

Shared/Samples/Style symbols from mobile style file/StyleSymbolsFromMobileStyleFileView.SymbolOptionsListView.swift
CalebRas: @rolson did you still think it would be better to move these things to the model?

rolson: Since the error handling was added, it does make sense to move more to the model so that code that calls these things doesn't have to do/catch and set an error on the model.

For the error handling, the methods don't have to be throwing per se, but could just set an error on the model.

CalebRas: I moved everything to the model and added an `updateDisplayScale(displayScale:)` method that prevents everything from reloading when the display scale doesn't actually change. I also removed the error checking from `updateCurrentSymbol()` since it would throw a cancellation error if you changed the symbols too quickly. Thanks for all your feedback!

dfeinzimer (dismissed):

rolson (approved):
Very nice!

dfeinzimer (approved):

PR #279, opened 2023-09-28
[New] Display dimensions

CalebRas:
## Description

This PR implements `Display dimensions` in `Layers` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-DisplayDimensions/Shared/Samples/Display%20dimensions)

## Linked Issue(s)

- `swift/issues/4586`

## How To Test

- Toggle "Dimension Layer" to change the layer's visibility.
- Toggle "Definition Expression: Dimensions >= 450m" to change the layer's definition expression. 

## Screenshots

display-dimensions


CalebRas:
I am not super happy with the design/implementation of the settings, but I couldn't find anything better since it you can't use toggles within a `Menu` and add having a half sheet for just two toggles does really make sense to me. So, I would appreciate any feedback you guys have on that!

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
```diff
@@ -0,0 +1,121 @@
...
+            }
+            .alert(isPresented: $isShowingErrorAlert, presentingError: error)
+    }
+    
+    /// The button that brings up the settings view.
+    @ViewBuilder private var settingsButton: some View {
+        let button = Button("Settings") {
+            isShowingSettings = true
+        }
+            .disabled(dimensionLayer == nil)
```
njarecha, line 72:
```suggestion
        }
        .disabled(dimensionLayer == nil)
```

njarecha (changes requested):
Looks good

njarecha:
@CalebRas 
> I am not super happy with the design/implementation of the settings, but I couldn't find anything better since it you can't use toggles within a Menu and add having a half sheet for just two toggles does really make sense to me. So, I would appreciate any feedback you guys have on that!

The settings is same as [old sample ](https://github.com/Esri/arcgis-runtime-samples-ios/tree/main/arcgis-ios-sdk-samples/Layers/Display%20dimensions). The design does not have the `settings button`. I'm ok with directly putting the two toggles directly in the overlay at bottom.

CalebRas:
I went with the overlay design that Nimesh suggested since the implementation is much simpler, has better backwards compatibility, and doesn't distract from the topic of the sample like the "Settings" button implementation did.

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
```diff
@@ -0,0 +1,105 @@
...
+                    
+                    // Set the dimension layer using the one on the map.
+                    if let layer = map.operationalLayers.first(where: { $0 is DimensionLayer }) as? DimensionLayer {
+                        dimensionLayer = layer
+                    } else {
+                        throw SetupError.noDimensionLayer
+                    }
+                } catch {
+                    self.error = error
+                }
```
rolson, line 58: If you wanted to clean up the view code some, you could move all this or just the section between the do/catch into an async function on the view.

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
```diff
@@ -0,0 +1,105 @@
...
+                .padding(.bottom)
+            }
+            .alert(isPresented: $isShowingErrorAlert, presentingError: error)
+    }
+}
+
+private extension DisplayDimensionsView {
+    /// The errors for the sample that can be thrown during setup.
+    private enum SetupError: String, LocalizedError {
+        case noMap = "The MMPK doesn't contain a map."
```
rolson, line 87: I would prefer not to use contractions in formal text:

```suggestion
        case noMap = "The MMPK does not contain a map."
```

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
```diff
@@ -0,0 +1,105 @@
...
+        case noDimensionLayer = "The map doesn't contain a dimension layer."
```
rolson, line 88:
```suggestion
        case noDimensionLayer = "The map does not contain a dimension layer."
```

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
```diff
@@ -0,0 +1,105 @@
...
+                    Spacer()
+                    Group {
+                        Toggle("Dimension Layer", isOn: Binding(
+                            get: { dimensionLayer?.isVisible ?? false },
+                            set: { dimensionLayer?.isVisible = $0 }
+                        ))
+                        Toggle("Definition Expression:\nDimensions >= 450m", isOn: Binding(
+                            get: { dimensionLayer?.definitionExpression == "DIMLENGTH >= 450" },
+                            set: { dimensionLayer?.definitionExpression = $0 ? "DIMLENGTH >= 450" : "" }
+                        ))
```
rolson, line 71: This is an interesting approach that I wouldn't have considered. I think it I like it better than what I would've done with State variables:

```swift

    @State private var dimensionLayerIsVisible = true
    @State private var definitionExpressionIsApplied = false
    
    var body: some View {
        MapView(map: map)
            .task {
                do {
                    // Load the local mobile map package using a URL.
                    let mapPackage = MobileMapPackage(fileURL: .edinburghPylonDimensions)
                    try await mapPackage.load()
                    
                    // Set the map to the first map in the mobile map package.
                    if let map = mapPackage.maps.first {
                        // Set the minScale to maintain dimension readability.
                        map.minScale = 4e4
                        self.map = map
                    } else {
                        throw SetupError.noMap
                    }
                    
                    // Set the dimension layer using the one on the map.
                    if let layer = map.operationalLayers.first(where: { $0 is DimensionLayer }) as? DimensionLayer {
                        dimensionLayer = layer
                    } else {
                        throw SetupError.noDimensionLayer
                    }
                } catch {
                    self.error = error
                }
            }
            .onChange(of: dimensionLayerIsVisible) {
                dimensionLayer?.isVisible = $0
            }
            .onChange(of: definitionExpressionIsApplied) { definitionExpressionIsApplied in
                if definitionExpressionIsApplied {
                    dimensionLayer?.definitionExpression = "DIMLENGTH >= 450"
                } else {
                    dimensionLayer?.definitionExpression = ""
                }
            }
            .overlay {
                VStack {
                    Spacer()
                    Group {
                        if dimensionLayer != nil {
                            Toggle("Dimension Layer", isOn: $dimensionLayerIsVisible)
                            Toggle("Definition Expression:\nDimensions >= 450m", isOn: $definitionExpressionIsApplied)
                        }
                    }
                    .padding(8)
                    .background(Color.white)
                    .cornerRadius(10)
                }
                .padding()
                .padding(.bottom)
            }
            .alert(isPresented: $isShowingErrorAlert, presentingError: error)
    }
```

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                    .padding(8)
+                    .background(Color.white)
+                    .cornerRadius(10)
+                }
+                .padding()
+                .padding(.bottom)
```
rolson, line 78: consider using `if let` so you don't show toggles if no dimension layer:

```suggestion                if let dimensionLayer {
                    VStack {
                        Spacer()
                        Group {
                            Toggle("Dimension Layer", isOn: Binding(
                                get: { dimensionLayer.isVisible },
                                set: { dimensionLayer.isVisible = $0 }
                            ))
                            Toggle("Definition Expression:\nDimensions >= 450m", isOn: Binding(
                                get: { dimensionLayer.definitionExpression == "DIMLENGTH >= 450" },
                                set: { dimensionLayer.definitionExpression = $0 ? "DIMLENGTH >= 450" : "" }
                            ))
                        }
                        .padding(8)
                        .background(Color.white)
                        .cornerRadius(10)
                    }
                    .padding()
                    .padding(.bottom)
                }
```

rolson (commented):
looks good. couple minor suggestions

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
CalebRas: Thanks!

Shared/Samples/Display dimensions/DisplayDimensionsView.swift
CalebRas: I think it would be more appropriate to disable the toggles so that they cannot be accessed while the dimension layer is loading/doesn't load. That way, the user can still see the intend functionality instead of it just being a map or an empty view.

rolson (approved):

njarecha (approved):
Looks good, just one suggestion.

njarecha (approved):

rolson (approved):

PR #281, opened 2023-10-06
[New] Identify raster cell

CalebRas:
## Description

This PR implements `Identify raster cell` in `Layers` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-IdentifyRasterCell/Shared/Samples/Identify%20raster%20cell)

## Linked Issue(s)

- `swift/issues/4596`

## How To Test

- Tap on the raster layer to see the popup for the raster cell at the tap location.
- Tap and hold to see map magnifier and the popup update dynamically as you drag.

## Screenshots

identify-raster-cell


mhdostal (dismissed):

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,252 @@
...
+            let xCoordinate = "X: \(extent.xMin.formatted(.rounded))"
+            let yCoordinate = "Y: \(extent.yMin.formatted(.rounded))"
+            
+            // Update the callout text.
+            calloutText = "\(attributes)\n\n\(xCoordinate)\n\(yCoordinate)"
+        }
+    }
+    
+    /// An enumeration used to throw an error customized with a string.
+    enum CustomError: LocalizedError {
```
zkline101, line 193: This is no longer needed because we made `rasterLayer` a non-optional constant.

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,252 @@
...
+                    .onEnded { value in
+                        guard let onEnded else { return }
+                        
+                        if value.first != nil {
+                            onEnded(Self.gestureScreenPoint, Self.gestureMapPoint)
+                        }
+                    }
+            )
+    }
+}
```
zkline101, line 238: Nice idea but could we just use [onDragGesture(shouldBegin:onChanged:onEnded:onCancelled:)](https://developers.arcgis.com/swift/api-reference/documentation/arcgis/mapview/ondraggesture(shouldbegin:onchanged:onended:oncancelled:)) ?

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,252 @@
...
+
+private extension FormatStyle where Self == FloatingPointFormatStyle<Double> {
+    /// The format style for rounding a decimal to three places.
+    static var rounded: Self {
```
zkline101, line 242: `rounded` is quite broad maybe we could say `roundedToThreeDecimalPlaces`. And this doesn't need to be in an extension because it is used in one area so it could just be a local variable there.

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,252 @@
...
+            if let location = proxy.location(fromScreenPoint: screenPoint) {
+                calloutPlacement = .location(location, offset: offset)
+            }
+        }
+        
+        /// Updates the text shown in the callout using the attributes and coordinates of a given raster cell.
+        /// - Parameter cell: The raster cell to create the text from.
+        private func updateCalloutText(using cell: RasterCell) {
+            // Create the attributes text using the attributes of the raster cell.
+            let attributes = cell.attributes.map { "\($0.key): \($0.value)" }.sorted(by: >).joined(separator: "\n")
```
zkline101, line 177: This line is pretty hard to read, we can clean that up:
```suggestion
            let attributes = cell.attributes
                .map { "\($0.key): \($0.value)" }
                .sorted(by: >)
                .joined(separator: "\n")
```

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,252 @@
...
+                return nil
+            }
+        }
+        
+        /// Updates the location of the callout placement to a given screen point.
+        /// - Parameters:
+        ///   - screenPoint: The screen point at which to place the callout.
+        ///   - proxy: The proxy used to convert the screen point to a map point.
+        ///   - offsetted: A Boolean value that indicates whether to offset the callout.
+        private func updateCalloutPlacement(to screenPoint: CGPoint, using proxy: MapViewProxy, offsetted: Bool) {
```
zkline101, line 159:
```suggestion
        private func updateCalloutPlacement(to screenPoint: CGPoint, using proxy: MapViewProxy, shouldUseOffset: Bool) {
```

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+        
+        /// Creates a callout displaying the data of a raster cell at a given screen point.
+        /// - Parameters:
+        ///   - screenPoint: The screen point of the raster cell at which to place the callout.
+        ///   - mapViewProxy: The proxy used to handle the screen point.
+        ///   - withOffset: A Boolean value that indicates whether to offset the callout's placement.
+        func callout(at screenPoint: CGPoint, using proxy: MapViewProxy, withOffset: Bool = false) {
+            Task { [weak self] in
```
zkline101, line 116: This can also be done using another [task view modifier](https://developer.apple.com/documentation/swiftui/view/task(id:priority:_:)) on the main view instead of kicking off a task here.

So when any of the gestures occur you could save that screen point in a @State variable and then when that value changes you call this method in that task view modifier linked above.

zkline101 (changes requested):

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
CalebRas: I don't think so. If the user long pressed but didn't drag, the callout placement would still be offset when the magnifier disappears and the gesture ends because the drag gesture would never be called.

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,226 @@
...
+            }
+            .gesture(
+                LongPressGesture()
+                    .simultaneously(with: DragGesture())
+                    .onEnded { value in
+                        guard let onEnded else { return }
+                        
+                        if value.first != nil {
+                            onEnded(Self.gestureScreenPoint, Self.gestureMapPoint)
+                        }
```
zkline101, line 215: Add doc for this code please.

Shared/Samples/Identify raster cell/IdentifyRasterCellView.swift
```diff
@@ -0,0 +1,200 @@
...
+            let xCoordinate = "X: \(extent.xMin.formatted(roundedStyle))"
+            let yCoordinate = "Y: \(extent.yMin.formatted(roundedStyle))"
+            
+            // Update the callout text.
+            calloutText = "\(attributes)\n\n\(xCoordinate)\n\(yCoordinate)"
+        }
+    }
+    
+    // A generic struct made up of two equatable types.
+    struct EquatablePair<T: Equatable, U: Equatable>: Equatable {
```
zkline101, line 160: I was contemplating if we really need this but I guess it is fine because the other solution would be to assign two task view modifiers to the view. One for if `calloutShouldOffset` changes and the other one for `tapScreenPoint` changes. But I'm not sure that is much better.

CalebRas: I didn't go with the two task modifiers so that the callout isn't created twice at the start of the long press. Besides that being less efficient/more redundant, doing so isn't async safe and can occasionally cause it to crash when creating two at once. Optimally, I wanted would want it to just watch the `tapScreenPoint`, but I couldn't figure out a way to force that variable to update in the gesture `onEnded` since the point doesn't actually change then. I guess I could make another state variable just for removing the offset and do two `task` modifiers with that if you think that is a good idea.

zkline101: I think what you have will be fine, thanks.

zkline101:
When I drag across the map, I get this pretty consistently where I haven't before. Can you look into this?
Simulator Screenshot - iPhone 15 Pro Max - 2023-10-13 at 12 10 17

CalebRas:
> When I drag across the map, I get this pretty consistently where I haven't before. Can you look into this? 

I am not able to reproduce it but that last push should fix it.

mhdostal (approved):

mhdostal:
> When I drag across the map, I get this pretty consistently where I haven't before. Can you look into this?

I could not reproduce this with the latest code.

zkline101:
Welp, I cannot repro it anymore 🤷

zkline101 (approved):

PR #283, opened 2023-10-09
[New] Identify graphics

CalebRas:
## Description

This PR implements `Identify graphics` in `Visualization` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-IdentifyGraphics/Shared/Samples/Identify%20graphics)

## Linked Issue(s)

- `swift/issues/4603`

## How To Test

- Tap on the graphic to see the alert.

## Screenshots

identify-graphics


Shared/Samples/Identify graphics/IdentifyGraphicsView.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    /// A Boolean value indicating whether to show an error alert.
+    @State private var isShowingErrorAlert = false
+    
+    /// The error shown in the error alert.
+    @State private var error: Error? {
+        didSet { isShowingErrorAlert = error != nil }
+    }
+    
+    var body: some View {
```
njarecha, line 61: This looks good but I would change this to following for doing identify operation in a `task` rather than in a closure of single tap gesture. Also, content of the result alert.

```swift
{
    /// A map with a topographic basemap.
    @State private var map = Map(basemapStyle: .arcGISTopographic)
    
    /// A graphics overlay containing a yellow polygon graphic.
    @State private var graphicsOverlay = {
        let graphicsOverlay = GraphicsOverlay()
        
        // Create a polygon from a list of points.
        let polygon = Polygon(points: [
            Point(x: -20e5, y: 20e5),
            Point(x: 20e5, y: 20e5),
            Point(x: 20e5, y: -20e5),
            Point(x: -20e5, y: -20e5)
        ])
        
        // Create a graphic of the polygon and add it to the overlay.
        let polygonGraphic = Graphic(geometry: polygon)
        graphicsOverlay.addGraphic(polygonGraphic)
        
        // Create a renderer using a simple fill symbol.
        let polygonSymbol = SimpleFillSymbol(style: .solid, color: .yellow)
        graphicsOverlay.renderer = SimpleRenderer(symbol: polygonSymbol)
        
        return graphicsOverlay
    }()
    
    /// A Boolean value indicating whether the graphic alert is showing.
    @State private var isShowingIdentifyResultAlert = false
    
    /// The text shown in the graphic alert that shows the alert when changed.
    @State private var identifyResultMessage = "" {
        didSet { isShowingIdentifyResultAlert = identifyResultMessage.isNotEmpty }
    }
    
    /// A Boolean value indicating whether to show an error alert.
    @State private var isShowingErrorAlert = false
    
    /// The error shown in the error alert.
    @State private var error: Error? {
        didSet { isShowingErrorAlert = error != nil }
    }
    
    /// Screen point for the identify operation.
    @State private var identifyScreenPoint: CGPoint?
    
    var body: some View {
        MapViewReader { proxy in
            MapView(map: map, graphicsOverlays: [graphicsOverlay])
                .onSingleTapGesture { screenPoint, _ in
                    identifyScreenPoint = screenPoint
                }
                .task(id: identifyScreenPoint) {
                    // Identify the screen point on the graphics overlay using the proxy.
                    // Use `identifyGraphicsOverlays` instead if you need to identify on
                    // all the graphics overlay present in the map view.
                    guard let identifyScreenPoint = identifyScreenPoint,
                          let identifyResult = await Result(awaiting: {
                              try await proxy.identify(
                                on: graphicsOverlay,
                                screenPoint: identifyScreenPoint,
                                tolerance: 12
                              )
                          }).cancellationToNil()
                    else {
                        return
                    }
                    
                    do {
                        // Display an alert if a graphic was found at the screen point.
                        let graphics = try identifyResult.get().graphics
                        if graphics.isNotEmpty {
                            self.identifyResultMessage = "\(graphics.count) graphic\(graphics.count > 1 ? "s" : "") found."
                        }
                    } catch {
                        // Show an error alert for an error throw while identifying.
                        self.error = error
                    }
                    
                    self.identifyScreenPoint = nil
                }
                .alert(
                    "Identify Result",
                    isPresented: $isShowingIdentifyResultAlert,
                    presenting: identifyResultMessage,
                    actions: { _ in },
                    message: { message in
                        Text(message)
                    }
                )
                .alert(
                    isPresented: $isShowingErrorAlert,
                    presentingError: error
                )
        }
    }
}
```

CalebRas: I moved the identify operation to a `task` modifier, as you suggested, but made made it a little more readable since this is supposed to be a pretty simple sample. I also updated the alert but kept the message the same. Even though yours makes more sense, I think it would be better to stick to the design here since all the other sample implementations do it similarly. Let me know what you think.

njarecha: Ok

njarecha (dismissed):
Looks good

Shared/Samples/Identify graphics/README.metadata.json
```diff
@@ -0,0 +1,25 @@
...
+    ],
+    "keywords": [
+        "graphics",
+        "identify",
+        "Graphic",
+        "GraphicsOverlay",
+        "MapView"
+    ],
+    "redirect_from": [],
+    "relevant_apis": [
```
pgruenler, line 16: Identify?

Shared/Samples/Identify graphics/IdentifyGraphicsView.swift
```diff
@@ -0,0 +1,112 @@
...
+                        let identifyResult = try await mapViewProxy.identify(
+                            on: graphicsOverlay,
+                            screenPoint: tapScreenPoint,
+                            tolerance: 12
+                        )
+                        
+                        // Display an alert if any graphics were found at the screen point.
+                        let graphics = identifyResult.graphics
+                        if graphics.isNotEmpty {
+                            let postfix = graphics.count > 1 ? "s" : ""
```
pgruenler, line 86: Is this needed? Can the count ever be more than one?

pgruenler (commented):
Looks good. I had a couple questions.

Shared/Samples/Identify graphics/IdentifyGraphicsView.swift
CalebRas: In this use case, no, because there is only one graphic on the map. If the user tapped where multiple graphics were overlapping, then yes. I think it would be good to leave it to show that use case and because design/other samples do it that way. (This sample would have been better with multiple graphics, but I am trying to follow the design as closely as possible.)

pgruenler (approved):

Shared/Samples/Identify graphics/IdentifyGraphicsView.swift
philium: > I think it would be good to leave it to show that use case

It is good to want to show users how to do things, but we also want to show users how to do things the right way. While the samples app isn't localized, the user's app might be, so we want to make sure that we are modeling best practices when it comes to localization. When localizing strings, concatenating them is a big no-no. While "graphic" + "s" = "graphics" may work in English, that won't work in many other languages. If users were to copy and paste this code into their own localized app (turning both strings into localized strings), it would introduce a bug. Try to keep that in mind when displaying anything localizable to the user.

Shared/Samples/Identify graphics/README.md
```diff
@@ -0,0 +1,31 @@
...
+1. Create a `GraphicsOverlay` and add it to the `MapView`.
+2. Add a `Graphic` along with a `SimpleFillSymbol` to the graphics overlay.
+3. Use the `MapView.onSingleTapGesture(perform:)` to get the screen point where a user tapped.
+4. Identify the graphic on the map view with `MapViewProxy.identify(on:screenPoint:tolerance:returnPopupsOnly:maximumResults:)`.
+
+## Relevant API
+
+* Graphic
+* GraphicsOverlay
+* IdentifyLayerResult
```
njarecha, line 26: This sample is using the `IdentifyGraphicsOverlayResult`

njarecha (approved):

Shared/Samples/Identify graphics/IdentifyGraphicsView.swift
```diff
@@ -0,0 +1,110 @@
...
+                        // all the graphics overlay present in the map view.
+                        let identifyResult = try await mapViewProxy.identify(
+                            on: graphicsOverlay,
+                            screenPoint: tapScreenPoint,
+                            tolerance: 12
+                        )
+                        
+                        // Display an alert if any graphics were found at the screen point.
+                        if identifyResult.graphics.isNotEmpty {
+                            identifyResultMessage = "Tapped on \(identifyResult.graphics.count) graphic(s)."
```
pgruenler, line 85: With the change to "graphic(s)" the screenshot doesn't match anymore. You could fix that with a new screenshot or a code change. I'm okay with either.

njarecha (approved):

pgruenler (approved):

PR #284, opened 2023-10-11
[New] Create mobile geodatabase

CalebRas:
## Description

This PR implements `Create mobile geodatabase` in `Edit and Manage Data` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-CreateMobileGeodatabase/Shared/Samples/Create%20mobile%20geodatabase)

## Linked Issue(s)

- `swift/issues/4607`

## How To Test

- Tap on the map to add feature points.
- Tap "View Table" to see the features in the feature table.
- Tap the share button to export the ".geodatabase" file, e.g. tap "Save to Flies"-> "Save".

## Screenshots

create-mobile-geodatabase-1 create-mobile-geodatabase-2


Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
```diff
@@ -0,0 +1,215 @@
...
+                try await featureTable.add(feature)
+                
+                // Update the feature count.
+                featureCount = featureTable.numberOfFeatures
+            } catch {
+                self.error = error
+            }
+        }
+        
+        /// Updates the list of features by queiring the feature table.
```
dfeinzimer, line 149:
```suggestion
        /// Updates the list of features by querying the feature table.
```

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
```diff
@@ -0,0 +1,215 @@
...
+                spatialReference: .wgs84,
+                geometryType: Point.self
+            )
+            
+            // Create and add the field descriptions for the table.
+            description.addFieldDescriptions([
+                // `FieldType.OID` is the primary key of the SQLite table.
+                FieldDescription(name: "oid", fieldType: .oid),
+                // `FieldType.DATE` is the date column used to store a calendar date.
+                FieldDescription(name: "collection_timestamp", fieldType: .date)
```
dfeinzimer, line 57: OID and oid seem to match but DATE and collection_timestamp do not. Should one change to match the other?

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
```diff
@@ -0,0 +1,215 @@
...
+            description.hasZ = false
+            
+            return description
+        }()
+        
+        /// The list of features in the feature table.
+        @Published private(set) var features: [FeatureItem] = []
+        
+        /// The count of features in the feature table.
+        @Published private(set) var featureCount = 0
```
dfeinzimer, line 72: Can `featureCount` be removed and callers can instead just derive this information from `features`?

`Model.updateFeatures()` will need to run on every new feature creation though.

dfeinzimer:
> I implemented the share button with an `UIActivityViewController` since [ShareLink](https://developer.apple.com/documentation/SwiftUI/ShareLink) is only available to iOS 16+

What about conditionally using a `ShareLink` with `if #available(iOS 16, *) ...`? I suspect 16.0 will be our minimum supported fairly soon but I could be mistaken.

> and there doesn't seem to be a good way to tell if the file had been shared to reset the geodatabase using that View

Should a succesfull share reset the GDB? Is that apart of the sample design? That might be unexpected from the user's perspective, even though this is just a sample. Maybe better to provide an explicit reset button?

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
CalebRas: To my knowledge [FieldType](https://developers.arcgis.com/swift/api-reference/documentation/arcgis/fieldtype) doesn't have a `timestamp` property, so we are using the `date` property as a timestamp here. The old iOS version did [the same thing](https://github.com/Esri/arcgis-runtime-samples-ios/blob/main/arcgis-ios-sdk-samples/Edit%20data/Create%20mobile%20geodatabase/CreateMobileGeodatabaseViewController.swift#L138)  as well as all the other implemenations.

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
CalebRas: Yes, we could do that. All the other samples use a [string instead of a count](url). We could also just add features to the list as they are created in `Model.addFeature(mapPoint:)`. The downside to that is we don't get to showcase querying the feature table. Then again we have a [dedicated sample](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/main/Shared/Samples/Query%20feature%20table) for that. Do you think we should get rid of both the `featureCount` and `Model.queryFeatures()`?

CalebRas:
> Should a successful share reset the GDB? Is that apart of the sample design? That might be unexpected from the user's perspective, even though this is just a sample. Maybe better to provide an explicit reset button?

Sort of. In the design, the button says "Create and share", and it seemed to imply that it removes all the features from all the map as they are made into a geodatabase to be shared. The old iOS implementation created a geodatabase first and adds features to it as it went but kept the remove when shared functionality. I wouldn't be opposed to adding a reset button or maybe an alert asking the user if they want to keep the file after sharing it. Let me know what you think.

dfeinzimer:
I just reviewed the design and it doesn't seem to definitive either way to me. For purposes of a sample I think what you've done is fine. 👍

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
dfeinzimer: If there's a dedicated sample for querying a feature table and it isn't vital to the functionality being sampled, yes, I'd probably opt to just keep the `features` list updated as they're made.

Shared/Samples/Create mobile geodatabase/README.md
```diff
@@ -0,0 +1,46 @@
+# Create mobile geodatabase
+
+Create and share a mobile geodatabase.
+
+![Create mobile geodatabase](create-mobile-geodatabase-1.png)
+![View geodatabase table](create-mobile-geodatabase-2.png)
+
+## Use case
+
+A mobile geodatabase is a collection of various types of GIS datasets contained in a single file (`.geodatabase`) on disk that can store, query, and manage spatial and nonspatial data. Mobile geodatabases are stored in a SQLite database and can contain up to 2 TB of portable data. Users can create, edit and share mobile geodatabases across ArcGIS Pro, ArcGIS Runtime, or any SQL software. These mobile geodatabases support both viewing and editing and enable new offline editing workflows that don't require a feature service.
```
mhdostal, line 10: I think `ArcGIS Runtime` should change.

Shared/Samples/Create mobile geodatabase/README.md
```diff
@@ -0,0 +1,46 @@
...
+* FeatureLayer
+* FeatureTable
+* FieldDescription
+* Geodatabase
+* GeodatabaseFeatureTable
+* TableDescription
+
+## Additional information
+
+Learn more about mobile geodatabases and how to utilize them on the [ArcGIS Pro documentation](https://pro.arcgis.com/en/pro-app/latest/help/data/geodatabases/manage-mobile-gdb/mobile-geodatabases.htm) page. The following mobile geodatabase behaviors are supported in ArcGIS Runtime: annotation, attachments, attribute rules, contingent values, dimensions, domains, feature-linked annotation, subtypes, utility network and relationship classes.
```
mhdostal, line 40: ~"Runtime"~

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.swift
```diff
@@ -0,0 +1,137 @@
...
+    
+    /// The list of features in the feature table.
+    var tableList: some View {
+        NavigationView {
+            List {
+                Section("OID and Collection Timestamp") {
+                    ForEach(model.features, id: \.self) { feature in
+                        HStack {
+                            Text(String(feature.oid))
+                            Spacer()
```
mhdostal, line 98: My personal preference would be to have the variable white space at the end. This way the text is closer to the number and the leading edges of the date/time strings are aligned. (You can see below, when using `Spacer()`, #14 is offset because of the "1" in the time) But it's just personal preference, not a big deal.

```suggestion
                                .padding(.trailing, 8)
```
image

mhdostal (changes requested):

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.swift
mhdostal: With `padding`:

dfeinzimer (approved):

PR #288, opened 2023-10-26
[New] Geocode offline

CalebRas:
## Description

This PR implements `Geocode offline` in `Search and Query` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-GeocodeOffline/Shared/Samples/Geocode%20offline)

## Linked Issue(s)

- `swift/issues/4643`

## How To Test

- **Note:** As of creation, this requires `swift-toolkit-daily` to compile.
- Tap a location on the map to reverse geocode.
- Enter an address in the search bar and press "Enter" to geocode.
- Tap the arrow button in the toolbar and select an address from the list to geocode.

## Screenshots

geocode-offline

## To Discuss

- I implemented the example addresses using a toolbar button instead of with the native suggestions (`.searchable(text:placement:prompt:suggestions:)`) because there is a bug in iOS 17 where the suggestions don't disappear when one is selected or when "Enter" is pressed, making them useless. Related Apple Developer Forum post [here](https://developer.apple.com/forums/thread/731700).


Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+    
+        /// The value used to offset the callout when the map magnifier is showing.
+        private let magnifierOffset = {
+            // Get the magnifier image from ArcGIS package bundle.
+            let bundle = Bundle(identifier: "com.esri.ArcGIS") ?? Bundle(identifier: "ArcGIS")
+            let image = UIImage(named: "Magnifier", in: bundle, with: nil) ?? UIImage()
+            
+            // Determine the amount by which we will need to offset the callout along y-axis.
+            return CGPoint(x: .zero, y: -image.size.height)
+        }()
```
yo1995, line 63: While this might improve user experience, I hesitate to include it as it provides too much unneeded details, even if it is in the old implementation.
Besides, this offset is currently (0, 0) as the image isn't correctly accessed from the local Swift package for some reason. I tried using image literal `#imageLiteral(resourceName: "Magnifier")` and it doesn't work either.

So my suggestion is to get rid of the offset altogether.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+                )
+                
+                if let result = geocodeResults.first,
+                   let displayLocation = result.displayLocation {
+                    // If a result is found, place a marker at the result's location.
+                    let resultText = result.attributes["Match_addr"] as? String ?? ""
+                    updateMarker(to: displayLocation, withText: resultText)
+                    
+                    // Zoom to the extent of the result's location.
+                    viewpoint = Viewpoint(boundingGeometry: displayLocation.extent)
```
yo1995, line 121: Can you refactor this method a bit so that the viewpoint is moved to the view itself?

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+        private let locatorTask = LocatorTask(name: "SanDiego_StreetAddress", bundle: .main)
+        
+        /// The viewpoint of the map view, initially centered on San Diego, CA, USA.
+        @Published var viewpoint = Viewpoint(
+            center: Point(x: -13_042_250, y: 3_857_970, spatialReference: .webMercator),
+            scale: 2e4
+        )
+        
+        /// The placement of the callout on the map.
+        @Published var calloutPlacement: CalloutPlacement?
```
yo1995, line 75: It would be ideal to keep the `calloutPlacement` a property of the view.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+        
+        /// A Boolean value indicating whether the callout placement should be offsetted, e.g. when the map magnifier is showing.
+        @Published var calloutShouldOffset = false
```
yo1995, line 78: If the suggestion above about offset is applied, this can be removed.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+                of: mapPoint
+            ) as? Point else { return }
+            
+            // Create reverse geocode parameters.
+            let reverseGeocodeParameters = ReverseGeocodeParameters()
+            reverseGeocodeParameters.addResultAttributeName("*")
+            reverseGeocodeParameters.maxResults = 1
+            
+            // Perform reverse geocode using the locator task with the point and parameters.
+            let geocodeResults = try? await locatorTask.reverseGeocode(
```
yo1995, line 147: It may be good to keep the error handling consistent for the geocode/reverseGeocode methods. In the method above it throws whereas here the error is omitted.

I'm fine with omitting both as it is unlikely to have null inputs in our case and there isn't network connection involved for offline.

CalebRas: Sounds good. The reason I omitted this one it is and occasionally throw a cancellation error when dragging.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+            
+            // Create a tiled layer using the local tile package.
+            let tileCache = TileCache(fileURL: .streetMap)
+            let tiledLayer = ArcGISTiledLayer(tileCache: tileCache)
+            
+            // Add the layer to the map.
+            map.addOperationalLayer(tiledLayer)
+            
+            return map
+        }()
```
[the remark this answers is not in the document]
CalebRas: Yea, that makes more sense.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,206 @@
...
+        /// The viewpoint of the map view, initially centered on San Diego, CA, USA.
+        @Published var viewpoint = Viewpoint(
+            center: Point(x: -13_042_250, y: 3_857_970, spatialReference: .webMercator),
+            scale: 2e4
+        )
+        
+        /// The placement of the callout on the map.
+        @Published var calloutPlacement: CalloutPlacement?
+        
+        /// A Boolean value indicating whether the callout placement should be offsetted, e.g. when the map magnifier is showing.
```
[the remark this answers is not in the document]
CalebRas: I believe so, but if you have a better one, I am open to suggestions.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+                   let displayLocation = result.displayLocation {
+                    // If a result is found, place a marker at the result's location.
+                    let resultText = result.attributes["Match_addr"] as? String ?? ""
+                    updateMarker(to: displayLocation, withText: resultText)
+                    
+                    // Zoom to the extent of the result's location.
+                    viewpoint = Viewpoint(boundingGeometry: displayLocation.extent)
+                } else {
+                    // If no result was found, inform the user with an alert.
+                    throw "No results found"
```
yo1995, line 110: I checked the .NET sample and [JS example](https://developers.arcgis.com/javascript/latest/tutorials/search-for-an-address/)

.NET sample displays an alert when there is no result. In our case I'm in favor of creating a separate alert only for the no result message, instead of making it part of the error alert.

image

When there is no result, JS display a no result message upon submit button clicked. As this sample is not showcasing the suggestion API, I think it's an overkill to write a UI to do this.

Shared/Samples/Geocode offline/GeocodeOfflineView.swift
```diff
@@ -0,0 +1,158 @@
...
+    private let exampleAddresses = [
+        "910 N Harbor Dr, San Diego, CA 92101",
+        "2920 Zoo Dr, San Diego, CA 92101",
+        "111 W Harbor Dr, San Diego, CA 92101",
+        "868 4th Ave, San Diego, CA 92101",
+        "750 A St, San Diego, CA 92101"
+    ]
+    
+    var body: some View {
+        GeocodeMapView(model: model, viewpoint: $viewpoint)
```
rolson, line 47: you can also pass the model through the environment instead of explicitly in the init.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
yo1995: It might not be easy to do, but still feel like the callout placement should be in the view rather than the model.

1. Move the `calloutPlacement`, `calloutShouldOffset` and `calloutText` into the view
2. Move `updateCalloutPlacement` to the view
3. Since the change of `searchText` affects `calloutPlacement`, it can trigger the child view to update
4. The `geocodeSearch` and `reverseGeocode` methods can both return a `GeocodeResult?`, and let the view unpack the result to get the attributes and `displayLocation`

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,170 @@
...
+        }()
+        
+        /// The locator task used to preform the geocode operations, loaded from a local file.
+        private let locatorTask = LocatorTask(name: "SanDiego_StreetAddress", bundle: .main)
+        
+        /// The placement of the callout on the map.
+        @Published var calloutPlacement: CalloutPlacement?
+        
+        /// A Boolean value indicating whether the callout placement should be offsetted, e.g. when the map magnifier is showing.
+        @Published var calloutShouldOffset = false
```
rolson, line 58: no biggie either way but `calloutIsOffset` should also work.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -0,0 +1,170 @@
...
+        // MARK: Methods
+        
+        /// Geocodes a given address and adds a marker with the corresponding address at the result's location.
+        /// - Parameter address: The given text address to geocode.
+        /// - Returns: The extent of the result's display location.
+        func geocodeSearch(address: String) async -> Envelope? {
+            guard let locatorTask else { return nil }
+            
+            // Create geocode parameters.
+            let geocodeParameters = GeocodeParameters()
```
yo1995, line 76: I'm in favor of keeping the GeocodeParameters construction as local variables, even if they don't change between searches and can be held as properties. Reasons: construction is rather low cost, and it shows a clearer workflow that a geocode/reverseGeocode requires parameters to be specifically configured. When a user copy this snippets, they can easily modify it to suit their usecase.

rolson (approved):
Looks good. Minor comments. I think ideally there would be some error reporting if (reverse) geocode fails. But no big deal either way. Nice sample

yo1995 (commented):
My suggestion would still be moving the callout placement to the view, but OK if the other reviewers agree to keep it as-is. Other than that the changes look good.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
CalebRas: Letting the the views unpack the `GeocodeResult?` would require both the views to have access to the callout variable since the `geocodeSearch` is called from `GeocodeOfflineView` and the `reverseGeocode` is called from `GeocodeMapView`. Are you suggesting to put the callout variables on the `GeocodeOfflineView` and then pass them all to the `GeocodeMapView`?  Thoughts @rolson?

rolson (approved):

yo1995 (approved):

Samples.xcodeproj/project.pbxproj
```diff
@@ -211,6 +212,8 @@
 		D75362D32A1E8C8800D83028 /* ApplyUniqueValueRendererView.swift in Copy Source Code Files */ = {isa = PBXBuildFile; fileRef = D75362D12A1E886700D83028 /* ApplyUniqueValueRendererView.swift */; };
 		D754E3232A1D66820006C5F1 /* StylePointWithPictureMarkerSymbolsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = D754E3222A1D66820006C5F1 /* StylePointWithPictureMarkerSymbolsView.swift */; };
 		D754E3242A1D66C20006C5F1 /* StylePointWithPictureMarkerSymbolsView.swift in Copy Source Code Files */ = {isa = PBXBuildFile; fileRef = D754E3222A1D66820006C5F1 /* StylePointWithPictureMarkerSymbolsView.swift */; };
+		D7553CDB2AE2DFEC00DC2A70 /* GeocodeOfflineView.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7553CD82AE2DFEC00DC2A70 /* GeocodeOfflineView.swift */; };
+		D7553CDD2AE2E00E00DC2A70 /* GeocodeOfflineView.swift in Copy Source Code Files */ = {isa = PBXBuildFile; fileRef = D7553CD82AE2DFEC00DC2A70 /* GeocodeOfflineView.swift */; };
```
yo1995, line 216: It seems GeocodeOfflineView.Model is not added to the copy source code files phase. Would you mind make a quick PR to fix that? Thanks.

PR #290, opened 2023-11-01
[Fix] Sample Viewer navigation bug

CalebRas:
## Description

This PR implements a fix to bugs that would come from nested `NavigationLink`s (e.g. using one in the top view of a sample) on iOS 16+ on both iPad and iPhone.

## Linked Issue(s)

- `swift/issues/NA` (quick fix)

## How To Test

- Ensure the navigating through the Sample Viewer is same.

### DIFF
--- a/Shared/Supporting Files/Views/ContentView.swift
+++ b/Shared/Supporting Files/Views/ContentView.swift
@@ -28,7 +28,9 @@ struct ContentView: View {
                     sidebar
                 }
             } detail: {
-                detail
+                NavigationStack {
+                    detail
+                }
             }
         } else {
             NavigationView {


yo1995 (approved):
From this doc: https://developer.apple.com/documentation/swiftui/navigationsplitview

> You can also embed a [NavigationStack](https://developer.apple.com/documentation/swiftui/navigationstack) in a column. Tapping or clicking a [NavigationLink](https://developer.apple.com/documentation/swiftui/navigationlink) that appears in an earlier column sets the view that the stack displays over its root view. Activating a link in the same column adds a view to the stack.

I think this change makes sense. When a sample is selected from the sidebar ("an earlier column"), the detail view shows the sample as the root view. When a navigation link is tapped in the sample itself, the child view is pushed onto the nav stack of the detail view.

I've skimmed through all samples and there wasn't one that shows a list as the root view, so this wasn't a problem before.

I'm a bit wary because I haven't found a similar usage on GitHub anywhere. Also since there are many kinds of sidebar-detail relationship that a split view supports, the doc isn't very clear for our specific usecase, where both the sidebar and detail view needs stack-based navigation.

Nonetheless, this change seems to work so let's try it out.

des12437 (approved):

PR #292, opened 2023-11-02
[Update] Add changes for callout in `Add dynamic entity layer`

yo1995:
## Description

This PR updates `Add dynamic entity layer` in `Layers` category.
[URL to README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Ting/UpdateDynamicEntitySample/Shared/Samples/Add%20dynamic%20entity%20layer/README.md)

## Linked Issue(s)

- `swift/issues/4688`
- `common-samples/pull/3882`

## How To Test

1. Run the sample, all existing features work
2. Tap on a vehicle. A callout should display
3. When the vehicle moves, the callout should follow along
4. When the observations are purged, the callout should be dismissed

## To Discuss

- The default values for speed and heading are `nan`. Very occasionally it causes the callout subview to give a warning in the console
> Bound preference SizePreferenceKey tried to update multiple times per frame.

If the default values are set to zero I don't see this warning. Shall we make them zero?

- If I don't reset the placement when observations are purged, the callout would stay in the old place when new observations are generated. Thus I passed a binding to the settings view.


Shared/Samples/Add dynamic entity layer/AddDynamicEntityLayerView.VehicleCallout.swift
```diff
@@ -0,0 +1,106 @@
...
+
+private extension Result<ArcadeEvaluationResult, Error> {
+    /// The evaluation as a string. If the evaluation results in an error, `nil`,
+    /// or a type other than a string, then an empty string is returned.
+    var stringValue: String {
+        switch self {
+        case .success(let evalutationResult):
+            if let resultString = evalutationResult.result(as: .string) as? String {
+                return resultString
+            }
```
CalebRas, line 100:
```suggestion
        switch self {
        case .success(let evaluationResult):
            if let resultString = evaluationResult.result(as: .string) as? String {
                return resultString
            }
```

CalebRas (commented):
Looks good! Just one comment.

CalebRas (approved):

njarecha (approved):
Looks good

PR #268, opened 2023-09-15
[New] Group layers together

CalebRas:
## Description

This PR implements `Group layers together` in `Layers` category.
URL to README: [ReadME](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-GroupLayersTogether/Shared/Samples/Group%20layers%20together)

## Linked Issue(s)

- `swift/issues/4075`

## How To Test

- Tap the "Layers" button to bring up the layers sheet.
- For project area group layers, toggle the switch next to a layer's name to change its visibility.
- For the buildings group layers, tap the cell with the layer's name to change its visibility.
- Toggle the switch next to a group layer's name to change the visibility of the whole group.

## Screenshots

group-layers-together

PR #294, opened 2023-11-08
[New] Find route around barriers

CalebRas:
## Description

This PR implements `Find route around barriers` in `Routing and Logistics` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-FindRouteAroundBarriers/Shared/Samples/Find%20route%20around%20barriers)

## Linked Issue(s)

- `swift/issues/4646`

## How To Test

- Select "Stops" and tap on the map to add stops to the route.
- Select "Barriers" and tap on the map tap on the map to add areas that can't be crossed by the route.
- Tap "Route" to find the route and display it.
- Tap the settings button to toggle preferences like find the best sequence or preserve the first or last stop.
- Tap the directions button to view a list of the directions and tap on a direction to see the route for the direction.
- Tap the trash button to remove all features associated with the current feature selection.

## Screenshots
find-route-around-barriers-1 find-route-around-barriers-2


Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+        var routeParameters: RouteParameters?
+        
+        /// The resulting route from a routing operation with the route task.
+        private(set) var route: Route?
+        
+        /// The text with the time and distance of the current route.
+        @Published private(set) var routeInfoText = ""
+        
+        /// The count of stops currently on the map.
+        @Published private(set) var stopsCount: Int = 0
```
zkline101, line 76: Nitpick: No need to specify the type
```suggestion
        @Published private(set) var stopsCount = 0
```

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+            )
+            textSymbol.offsetY = stopSymbol.offsetY
+            
+            // Create a graphic with the marker symbol and text symbol.
+            let compositeSymbol = CompositeSymbol(symbols: [stopSymbol, textSymbol])
+            let stopGraphic = Graphic(geometry: point, symbol: compositeSymbol)
+            
+            // Add the new graphic to the stop graphics overlay.
+            stopGraphicsOverlay.addGraphic(stopGraphic)
+            stopsCount += 1
```
zkline101, line 105: Why do you need a stop count? Can't you just check how many elements are in `stopGraphicsOverlay` ?

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+                barrierGraphicsOverlay.removeAllGraphics()
+            }
+        }
+        
+        /// Routes using the route parameters and the current stops and barriers on the map.
+        func route() async throws {
+            guard let routeParameters, stopsCount >= 2 else { return }
+            
+            // Update the route parameters' stops using the stop graphics.
+            var stops = [Stop]()
```
zkline101, line 143: Do you need this and the `barriers` array? Can't you just append the stop or barrier to the parameters after you create an instance?

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+            // Format the route's distance.
+            let distanceText = route.totalLength.formatted()
+            
+            routeInfoText = "\(timeText) (\(distanceText))"
+        }
+    }
+    
+    /// An enumeration representing different features that can be added to the map by the user in this sample.
+    enum AddableFeature {
+        case stop, barrier
```
zkline101, line 196: Each case should have some doc as well.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,237 @@
...
+                        Spacer()
+                        
+                        SheetButton(title: "Directions") {
+                            List {
+                                ForEach(
+                                    Array((model.route?.directionManeuvers ?? []).enumerated()),
+                                    id: \.offset
+                                ) { (_, direction) in
+                                    Button {
+                                        Task {
```
zkline101, line 97: Similar comment to the one above.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+        /// The graphic for the direction route.
+        let directionGraphic = Graphic(
+            symbol: SimpleLineSymbol(style: .dashDot, color: .orange, width: 5)
+        )
+        
+        /// The route task for routing.
+        let routeTask = RouteTask(url: .sanDiegoNetworkAnalysis)
+        
+        /// The route parameters for routing with the route task.
+        var routeParameters: RouteParameters?
```
zkline101, line 67: Instead of making this optional and having to unwrap it in multiple places we can probably make the default an empty parameters using `RouteParameters()`

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+                let stop = Stop(point: point)
+                stops.append(stop)
+            }
+            routeParameters.clearStops()
+            routeParameters.setStops(stops)
+            
+            // Update the route parameters' barriers using the barrier graphics.
+            var barriers = [PolygonBarrier]()
+            for barrierGraphic in barrierGraphicsOverlay.graphics {
+                guard let polygon = barrierGraphic.geometry as? Polygon else { continue }
```
zkline101, line 155: I'm getting a build error here:
```suggestion
                guard let polygon = barrierGraphic.geometry as? ArcGIS.Polygon else { continue }
```

zkline101 (commented):
A few other notes:
- I noticed weird behavior where I hit the delete button, and the barriers were deleted but the route and stops were not. Can you reproduce this?
- The delete button should probably only be active if there is something to delete

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
CalebRas: No. The "Route" button's `disabled` modifier won't update just by watching `stopGraphicsOverlay.graphics.count` and requires some type of State variable to work correctly.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
CalebRas: To my knowledge, there isn't a way to append stops to the route parameter since set methods overwrite the previous values I think. I removed clear methods since, if that's the case, they are not necessary.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,237 @@
...
+
+struct FindRouteAroundBarriersView: View {
+    /// The view model for the sample.
+    @StateObject private var model = Model()
+    
+    /// The feature type to be added or removed from the map.
+    @State private var featureSelection: AddableFeature = .stop
+    
+    /// A Boolean value indicating whether a routing operation is in progress.
+    @State private var routing = false
```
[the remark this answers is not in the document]
CalebRas: I modified the name a bit to match [this feedback](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/288#discussion_r1376791534) from Ryan.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,237 @@
...
+    
+    /// A Boolean value indicating whether routing will find the best sequence.
+    @State private var findsBestSequence = false
```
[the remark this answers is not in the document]
CalebRas: Same comment as above.

CalebRas:
> I noticed weird behavior where I hit the delete button, and the barriers were deleted but the route and stops were not. Can you reproduce this?

Yes. This is intended behavior since it is how it was [done in the old iOS sample](https://github.com/Esri/arcgis-runtime-samples-ios/blob/main/arcgis-ios-sdk-samples/Route%20and%20directions/Route%20around%20barriers/RouteAroundBarriersViewController.swift#L200-L207). The thought process is that the user would be able to test with different stops while keeping the barriers or vice versa. I did disable the button based on the current feature selection so that functionality is more implied.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,246 @@
...
+                        Button {
+                            model.reset(features: featuresSelection)
+                        } label: {
+                            Image(systemName: "trash")
+                        }
+                        .disabled(
+                            featuresSelection == .stops ? model.stopsCount == 0 :
+                                model.barriersCount == 0
+                        )
+                    }
```
rolson, line 168: Not necessary, but if you wanted you could break this out into a separate view. You could pass context through the environment

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Model.swift
```diff
@@ -0,0 +1,212 @@
...
+        /// Routes using the route parameters and the current stops and barriers on the map.
+        func route() async throws {
+            // Update the route parameters' stops using the stop graphics.
+            var stops = [Stop]()
+            for stopGraphic in stopGraphicsOverlay.graphics {
+                guard let point = stopGraphic.geometry as? Point else { continue }
+                let stop = Stop(point: point)
+                stops.append(stop)
+            }
+            routeParameters.setStops(stops)
```
rolson, line 152: I've been advised to try to avoid imperative for loops in Swift and to prefer, where possible, to use functional style programming constructs:

```suggestion
            routeParameters.setStops(
                stopGraphicsOverlay.graphics
                    .compactMap { $0.geometry as? Point }
                    .map(Stop.init(point:))
            )
```

Same feedback applies elsewhere.

rolson (commented):
Great sample!

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,242 @@
...
+                self.error = error
+            }
+        }
+        .alert(isPresented: $errorAlertIsShowing, presentingError: error)
+    }
+}
+
+private extension FindRouteAroundBarriersView {
+    /// A button with a given label that brings up a sheet containing given content.
+    struct SheetButton<Content: View, Label: View>: View {
```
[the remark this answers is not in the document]
CalebRas: Possibly. I am planning on updating the custom sheet modifier in the Samples during Sprint X.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,242 @@
...
+                    .toolbar {
+                        ToolbarItem(placement: .confirmationAction) {
+                            Button("Done") {
+                                sheetIsShowing = false
+                            }
+                        }
+                    }
+            }
+            .navigationViewStyle(.stack)
+            .frame(idealWidth: 320, idealHeight: 428)
```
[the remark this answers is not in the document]
CalebRas: This line was included accidentally.

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
```diff
@@ -0,0 +1,242 @@
...
+            if #available(iOS 16, *) {
+                button
+                    .popover(isPresented: $sheetIsShowing, arrowEdge: .bottom) {
+                        sheetContent
+                            .presentationDetents([.fraction(0.5)])
+#if targetEnvironment(macCatalyst)
+                            .frame(minWidth: 300, minHeight: 270)
+#else
+                            .frame(minWidth: 320, minHeight: 390)
+#endif
```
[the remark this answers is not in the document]
CalebRas: We have been doing this in the samples as setting the frame helps [improve visibility](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/268#discussion_r1329016967) on iPhone (especially when there is a lot of content in the popover, such as with directions sheet,)  and it allows the popover content to be viewable on Mac Catalyst. Without it, the popover is tiny. Correct me if I am wrong @philium.

rolson (approved):

zkline101 (approved):
Looks good

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.swift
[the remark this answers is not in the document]
philium: No, that sounds right. I had tried various different things to get the popover to size its height to fit, but was unsuccessful.

PR #296, opened 2023-11-09
[Setup] Add Privacy Manifest

zkline101:
## Description

These are the things that need to be declared in the privacy manifest if we collect any of this info:
* [Describing data use](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests)
- [Required Use APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

I only listed in the privacy manifest that we use course location for a qualified reason. I didn't see any other items we need to list, please confirm this as a part of reviewing this PR.

Here is the privacy report for the Samples:
[Samples-PrivacyReport 2023-11-08 15-52-45.pdf](https://github.com/Esri/arcgis-maps-sdk-swift-samples/files/13302445/Samples-PrivacyReport.2023-11-08.15-52-45.pdf)

## Linked Issue(s)

- `swift/issues/4177`

### DIFF
--- a/Shared/PrivacyInfo.xcprivacy
+++ b/Shared/PrivacyInfo.xcprivacy
@@ -7,20 +7,7 @@
 	<key>NSPrivacyTrackingDomains</key>
 	<array/>
 	<key>NSPrivacyCollectedDataTypes</key>
-	<array>
-		<dict>
-			<key>NSPrivacyCollectedDataType</key>
-			<string>NSPrivacyCollectedDataTypeCoarseLocation</string>
-			<key>NSPrivacyCollectedDataTypeLinked</key>
-			<false/>
-			<key>NSPrivacyCollectedDataTypeTracking</key>
-			<false/>
-			<key>NSPrivacyCollectedDataTypePurposes</key>
-			<array>
-				<string>NSPrivacyCollectedDataTypePurposeAppFunctionality</string>
-			</array>
-		</dict>
-	</array>
+	<array/>
 	<key>NSPrivacyAccessedAPITypes</key>
 	<array/>
 </dict>
--- a/Shared/PrivacyInfo.xcprivacy
+++ b/Shared/PrivacyInfo.xcprivacy
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
+	<array>
+		<dict>
+			<key>NSPrivacyCollectedDataType</key>
+			<string>NSPrivacyCollectedDataTypeCoarseLocation</string>
+			<key>NSPrivacyCollectedDataTypeLinked</key>
+			<false/>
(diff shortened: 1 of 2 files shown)


Shared/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,27 @@
...
+		<dict>
+			<key>NSPrivacyCollectedDataType</key>
+			<string>NSPrivacyCollectedDataTypeCoarseLocation</string>
+			<key>NSPrivacyCollectedDataTypeLinked</key>
+			<false/>
+			<key>NSPrivacyCollectedDataTypeTracking</key>
+			<false/>
+			<key>NSPrivacyCollectedDataTypePurposes</key>
+			<array>
+				<string>NSPrivacyCollectedDataTypePurposeAppFunctionality</string>
```
yo1995, line 20: Here is the "[Privacy Nutrition Label](https://developer.apple.com/app-store/app-privacy-details/)" for the 100.x samples app. Since App Store Connect reviewers allowed us to publish the old app without adding this entry, I'm a little bit hesitant to add it. But if that is what Apple requires us to do now, then let's do it.

image

image

From the doc

> “Collect” refers to transmitting data off the device in a way that allows you and/or your third-party partners to access it for a period longer than what is necessary to service the transmitted request in real time.

I can see both sides on whether this app is counted as collecting location data or not. 🎭

yo1995 (dismissed):
Based on a similar PR https://github.com/duckduckgo/iOS/pull/1936 the format and the location where the manifest is added look good.

Shared/PrivacyInfo.xcprivacy
zkline101: @dg0yal Do you have any thoughts on this?

zkline101: [This](https://developer.apple.com/app-store/app-privacy-details/) is interesting too:
Screenshot 2023-11-09 at 9 54 00 AM

dg0yal: No we don't "collect" data as per this definition.

zkline101: @yo1995 Apple says this too, "For example, if you collect data solely for the purpose of app functionality, declare the data on your label and indicate that it is only being used for that purpose."

dg0yal: Yeah, but it doesn't change the definition of "collect"
For instance, you could be sending device location to a server to track a user's run. That would be data collection for app functionality

zkline101: I see, good point. Ok, sounds like I can remove this item from the privacy manifest then.

rolson: So if we had a sample that showed how to take your current location and sent it to a feature service, then we would need to report that in the nutrition label, right? Same if we had a sample that geocoded your current location, right? 

So we probably just need to consider this when implementing samples and how samples might affect what we "collect".

dg0yal: storing curring location in a feature service, yes i think so

geocoding your current location, probably not (does not satisfy the _allows you and/or your third-party partners to access it for a period longer than what is necessary to service the transmitted request in real time_   )

zkline101: That makes sense thank you. The only sample where we get the user's location is the show device location. In that sample, we are simply showing the user's location and are not transmitting it anywhere. So no need for this location item in our privacy manifest at this moment.

yo1995 (approved):

des12437 (approved):

zkline101:
Thanks!

PR #306, opened 2023-11-29
[Fix] Custom sheet auto dismiss

CalebRas:
## Description

This PR implements fixes to the following bugs that would occur when using the custom sheet modifier in the sample viewer:
1. If there was more than one custom sheet within a given view, some of the sheets would auto-dismiss upon opening, e.g. a sample containing settings sheet would cause the info sheet to auto-dismiss when opened.
2. Whenever a custom sheet was closed, `Modifying state during view update, this will cause undefined behavior.` would be thrown. 
3. Custom sheets dismissed using the `DismissAction` from the `Environment` wouldn't always re-present.

URL to custom sheet modifier file: [View+Sheet.swift](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Caleb/Fix-SheetExtensionAutoDismiss/Shared/Supporting%20Files/Extensions/View%2BSheet.swift)

## Linked Issue(s)

- `swift/issues/4778`

## How To Test

<details><summary>Test Code</summary>

```swift
struct DisplayMapView: View {
    /// A map with imagery basemap.
    @State private var map = Map(basemapStyle: .arcGISImagery)
    
    @State private var isShowingDoneSheet = false
    
    @State private var isShowingDragSheet = false
    
    @State private var isShowingDismissSheet = false
    
    var body: some View {
        // Creates a map view to display the map.
        MapView(map: map)
            .toolbar {
                ToolbarItemGroup(placement: .bottomBar) {
                    Button("Plain") {
                        isShowingDragSheet = true
                    }
                    .sheet(isPresented: $isShowingDragSheet, detents: [.medium, .large], dragIndicatorVisibility: .visible) {
                        print("Dismissed")
                    } content: {
                        Text("Some text.")
                    }
                    Spacer()
                    
                    Button("Done") {
                        isShowingDoneSheet = true
                    }
                    .sheet(isPresented: $isShowingDoneSheet, detents: [.medium]) {
                        print("Dismissed")
                    } content: {
                        NavigationView {
                            Text("Press Done.")
                                .toolbar {
                                    ToolbarItem(placement: .confirmationAction) {
                                        Button("Done") {
                                            isShowingDoneSheet = false
                                        }
                                    }
                                }
                        }
                    }
                    Spacer()
                    
                    Button("Dismiss") {
                        isShowingDismissSheet = true
                    }
                    .sheet(isPresented: $isShowingDismissSheet, detents: [.large]) {
                        print("Dismissed")
                    } content: {
                        DismissSheetView()
                    }
                }
            }
    }
}

private extension DisplayMapView {
    struct DismissSheetView: View {
        /// The action to dismiss the sheet.
        @Environment(\.dismiss) private var dismiss
        
        var body: some View {
            NavigationView {
                Text("Press dismiss")
                    .toolbar {
                        ToolbarItem(placement: .confirmationAction) {
                            Button("Dismiss") {
                                dismiss()
                            }
                        }
                    }
            }
        }
    }
}

```
</details>

- Copy the code above into the `Display map` sample.
- In the sample, verify the sheets open and close as expected using the buttons in the bottom toolbar.
- Verify that info sheet (top-right button) works as expected.
- Verify the rest of the samples using the custom sheet work as expected.

### DIFF
--- a/Shared/Supporting Files/Extensions/View+Sheet.swift
+++ b/Shared/Supporting Files/Extensions/View+Sheet.swift
@@ -26,6 +26,9 @@ extension View {
     ///   - idealHeight: The ideal height of the popover.
     ///   - onDismiss: A closure to execute when dismissing the sheet.
     ///   - content: A closure returning the content of the sheet.
+    ///
+    /// > Note: This modifier may not work with other UIKit derived views, such as an `Alert`.
+    ///  Using them both in the same view may cause undefined behavior.
     func sheet<Content>(
         isPresented: Binding<Bool>,
         detents: [Detent],
--- a/Shared/Supporting Files/Extensions/View+Sheet.swift
+++ b/Shared/Supporting Files/Extensions/View+Sheet.swift
@@ -176,28 +176,33 @@ private extension SheetModifier {
     func makeContentWithSheetWrapper(_ content: Content) -> some View {
         ZStack {
             content
-                .popover(
-                    isPresented: Binding(
-                        get: { isPresented && !isSheetLayout },
-                        set: { isPresented = $0 }
-                    )
-                ) {
+                .task(id: isPresented) {
+                    if isPresented {
+                        // Sleep to prevent appearing when other content is disappearing.
+                        try? await Task.sleep(nanoseconds: 100)
+                        
+                        if isSheetLayout {
+                            isSheetVisible = true
+                        } else {
+                            isPopoverVisible = true
+                        }
+                    } else {
+                        isSheetVisible = false
+                        isPopoverVisible = false
+                    }
+                }
+                .popover(isPresented: $isPopoverVisible) {
                     sheetContent
                         .frame(idealWidth: idealWidth, idealHeight: idealHeight)
-                        .onAppear { isPopoverVisible = true }
                         .onDisappear {
                             isPopoverVisible = false
-                            if !isPresented {
-                                onDismiss?()
-                            }
+                            isPresented = false
+                            onDismiss?()
                         }
                 }
             
             Sheet(
-                isPresented: Binding(
-                    get: { isPresented && !isPopoverVisible },
-                    set: { isPresented = $0 }
-                ),
+                isPresented: $isSheetVisible,
                 detents: detents.map { $0.sheetDetent },
                 selection: Binding(
                     get: {
@@ -212,9 +217,9 @@ private extension SheetModifier {
             ) {
                 sheetContent
                     .onDisappear {
-                        if !isPresented {
-                            onDismiss?()
-                        }
+                        isSheetVisible = false
+                        isPresented = false
+                        onDismiss?()
                     }
             }
             .fixedSize()
@@ -320,11 +325,6 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
             // not currently being dismissed, and is not an alert.
             rootViewController.dismiss(animated: presentedControllerIsHosting)
             model.isTransitioningFromPopover = !presentedControllerIsHosting
-        } else if hostingControllerIsBeingDismissed {
-            // Sets 'isPresented' to false when the hosting controller is being dismissed.
-            Task {
-                isPresented = false
-            }
         } else if wasPresenting {
             // Updates the sheet presentation controller and the root view of the hosting
             // controller if the sheet was already presenting.
--- a/Shared/Supporting Files/Extensions/View+Sheet.swift
+++ b/Shared/Supporting Files/Extensions/View+Sheet.swift
@@ -286,26 +286,26 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
         /// A Boolean value indicating whether the presented view controller is a hosting controller.
         let isPresentedControllerHostingType = rootViewController.presentedViewController is UIHostingController<Content>
         
+        /// A Boolean value indicating whether the presented view controller is an alert controller.
+        let isPresentedControllerAlertType = rootViewController.presentedViewController is UIAlertController
+        
+        /// A Boolean value indicating whether the sheet was already presenting.
+        let wasPresenting = rootViewController.presentedViewController != nil && isPresentedControllerHostingType
+        
+        /// A Boolean value indicating whether the hosting controller is being dismissed.
+        let isHostBeingDismissed = model.hostingController.isBeingDismissed
+        
         // Ensures that the device's layout is such that a sheet should be presented and
         // the hosting controller's sheet presentation controller exists.
         guard isSheetLayout,
               let sheet = model.hostingController.sheetPresentationController else {
             // Dismisses the sheet if it is being presented and a popover should
             // be presented instead.
-            if isPresentedControllerHostingType && !model.hostingController.isBeingDismissed {
+            if isPresentedControllerHostingType && !isHostBeingDismissed {
                 rootViewController.dismiss(animated: false)
             }
             return
         }
-        
-        /// A Boolean value indicating whether the sheet was already presenting.
-        let wasPresenting = rootViewController.presentedViewController != nil
-        
-        /// A Boolean value indicating whether the hosting controller is being dismissed.
-        let isHostBeingDismissed = model.hostingController.isBeingDismissed
-        
-        /// A Boolean value indicating whether the presented view controller is an alert controller.
-        let isPresentedControllerAlertType = rootViewController.presentedViewController is UIAlertController
         
         if isPresented && !wasPresenting {
             // Sets the sheet presentation controller's delegate.
@@ -313,17 +313,18 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
             // Configures the hosting controller's sheet presentation controller.
             configureSheetPresentationController(sheet)
             // Presents the hosting controller.
-            rootViewController.present(model.hostingController, animated: model.isTransitioningFromPopover ? false : true)
+            rootViewController.present(model.hostingController, animated: !model.isTransitioningFromPopover)
         } else if !isPresented && wasPresenting && !isHostBeingDismissed && !isPresentedControllerAlertType {
             // Dismisses the view controller presented by the root view controller
             // if 'isPresented' is false, but was presenting before (popover), is
             // not currently being dismissed, and is not an alert.
-            rootViewController.dismiss(animated: isPresentedControllerHostingType ? true : false)
+            rootViewController.dismiss(animated: isPresentedControllerHostingType)
             model.isTransitioningFromPopover = !isPresentedControllerHostingType
         } else if isHostBeingDismissed {
             // Sets 'isPresented' to false when the hosting controller is being dismissed.
-            isPresented = false
-            model.isTransitioningFromPopover = !isPresentedControllerHostingType
+            Task {
+                isPresented = false
+            }
         } else if wasPresenting {
             // Updates the sheet presentation controller and the root view of the hosting
             // controller if the sheet was already presenting.
--- a/Shared/Supporting Files/Extensions/View+Sheet.swift
+++ b/Shared/Supporting Files/Extensions/View+Sheet.swift
@@ -26,9 +26,8 @@ extension View {
     ///   - idealHeight: The ideal height of the popover.
     ///   - onDismiss: A closure to execute when dismissing the sheet.
     ///   - content: A closure returning the content of the sheet.
-    ///
-    /// > Note: This modifier may not work with other UIKit derived views, such as an `Alert`.
-    ///  Using them both in the same view may cause undefined behavior.
+    /// - Note: This modifier can have conflict with modal presentation views, such as an alert.
+    /// When the sheet is presented, it may cause the "already presenting" problem.
     func sheet<Content>(
         isPresented: Binding<Bool>,
         detents: [Detent],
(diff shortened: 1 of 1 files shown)


Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -286,44 +286,45 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
...
-            rootViewController.dismiss(animated: isPresentedControllerHostingType ? true : false)
+            rootViewController.dismiss(animated: isPresentedControllerHostingType)
             model.isTransitioningFromPopover = !isPresentedControllerHostingType
         } else if isHostBeingDismissed {
             // Sets 'isPresented' to false when the hosting controller is being dismissed.
-            isPresented = false
-            model.isTransitioningFromPopover = !isPresentedControllerHostingType
+            Task {
+                isPresented = false
+            }
```
yo1995, line 327: This is great!

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -286,44 +286,45 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
...
-        
         if isPresented && !wasPresenting {
             // Sets the sheet presentation controller's delegate.
             sheet.delegate = context.coordinator
             // Configures the hosting controller's sheet presentation controller.
             configureSheetPresentationController(sheet)
             // Presents the hosting controller.
-            rootViewController.present(model.hostingController, animated: model.isTransitioningFromPopover ? false : true)
+            rootViewController.present(model.hostingController, animated: !model.isTransitioningFromPopover)
         } else if !isPresented && wasPresenting && !isHostBeingDismissed && !isPresentedControllerAlertType {
```
yo1995, line 317: The changes solved most of the problems, but there is still an issue when an alert and sheet are in conflict. Run the test code on an iPad and add an alert button
```swift
Button("alert") {
    isShowingAlert = true
}
.alert(isPresented: $isShowingAlert) {
    Alert(title: Text("Alert!"))
}
```
When I tap either one of the sheet buttons, then tap the (grayed) alert button while the sheet is presenting, in console it throws the error

> Attempt to present *AlertController* on *UIHostingController: 0x111017200* (from <SwiftUI.UIKitNavigationController: 0x11782e400>) which is already presenting

I think there is still room for improvement to these conditions and I'll try to sort it out. 🧐

CalebRas: Wouldn't this work in that case? @yo1995 
```swift
let wasPresenting = rootViewController.presentedViewController != nil && isPresentedControllerHostingType && !isPresentedControllerAlertType
```

Shared/Supporting Files/Extensions/View+Sheet.swift
njarecha: The `Task` should be used to execute async code but `isPresented` is not async so what are we trying to do here?

njarecha (commented):
This is just initial review. I'll continue on testing the fix.

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -286,44 +286,45 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
         /// A Boolean value indicating whether the presented view controller is a hosting controller.
         let isPresentedControllerHostingType = rootViewController.presentedViewController is UIHostingController<Content>
         
+        /// A Boolean value indicating whether the presented view controller is an alert controller.
+        let isPresentedControllerAlertType = rootViewController.presentedViewController is UIAlertController
```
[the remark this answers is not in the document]
pgruenler: Maybe `presentedControllerIsAlertType`

Shared/Supporting Files/Extensions/View+Sheet.swift
yo1995: No it won't. I think the problem is that on iPad, the [formSheet](https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/formsheet) styled sheet is dismissed too late. So when the alert is presented, the sheet is still on the screen, which causes the error.

I thought I found a simple fix, but it was a red herring 🥲 . I'll keep looking for a little while, but it shouldn't hold up the review of this PR.

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -286,44 +286,45 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
...
+        
+        /// A Boolean value indicating whether the sheet was already presenting.
+        let wasPresenting = rootViewController.presentedViewController != nil && isPresentedControllerHostingType
+        
+        /// A Boolean value indicating whether the hosting controller is being dismissed.
+        let isHostBeingDismissed = model.hostingController.isBeingDismissed
```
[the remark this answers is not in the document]
njarecha: Just using `host` sounds little misleading as it is hosting controller.

Shared/Supporting Files/Extensions/View+Sheet.swift
CalebRas: The `Task` is used to updating State variables without modifying them while the view is updating otherwise you get: `Modifying state during view update, this will cause undefined behavior.`

njarecha: I don't think this is a right use of the `Task`. The warning comes up because the value is updated while the view is being updated. Should it happen after it finished updating?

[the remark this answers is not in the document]
CalebRas: Are you referring to the dismiss in the block above these lines? I am not sure this would work there. This needed when using `DismissAction` to dismiss the sheet since it otherwise you cannot re-present that specific sheet. It doesn't correctly set `isPresented` to false so the block above wouldn't ever run.

CalebRas: I am not sure how I would do that. Do you have any suggestions?

yo1995: I agree that it is not the right use of `Task` here. It seems that by wrapping the line in the block, the timing is slightly delayed so that the sheet can be dismissed prior to  `isPresented` is set to `false`.

In the original PR there was a different issue and eventually a `Task` block was used to work around that bug. https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/39#discussion_r933540923 You can find the task block in the condition branch a few lines below.

I'm OK if the threading problem can be mitigated by using a task block albeit it is not a right use - after all, this extension will be removed when we move to iOS 16.

njarecha: If there is no other way then I would suggest using the `DispatchQueue.main.async` instead of `Task`.

```swift
        } else if hostingControllerIsBeingDismissed {
            // Sets 'isPresented' to false when the hosting controller is being dismissed.
            DispatchQueue.main.async {
                isPresented = false
            }
        } else if wasPresenting {
            // Updates the sheet presentation controller and the root view of the hosting
            // controller if the sheet was already presenting.
            configureSheetPresentationController(sheet)
            DispatchQueue.main.async {
                model.hostingController.rootView = content
            }
        }
```

philium: If at all possible, GCD and modern Swift concurrency should not be mixed. [`MainActor.run(resultType:body:)`](https://developer.apple.com/documentation/swift/mainactor/run(resulttype:body:)) should be used instead of `DispatchQueue.main.async { ... }`.

njarecha: The above code is part of `func updateUIView(_ uiView: UIView, context: Context)` which is a synchronous function and use of `MainActor.run` requires using `await`. So it can't be used without using `Task` and we are trying to avoid using the `Task`.

rolson: I don't think using a Task is a problem in this case. Or at least I don't see why you would avoid it here. It's basically has the same cons that dispatch async has.

njarecha: I'm ok if we want to continue using `Task` here. My understanding is that `Task` should be used to execute `async` code from a `sync` context, not to convert the `sync` code to `async`.

njarecha (dismissed):
Changes looks good and improves for some use cases. We can continue looking for improving some other cases separately.

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -284,46 +284,47 @@ private struct Sheet<Content>: UIViewRepresentable where Content: View {
...
-            rootViewController.dismiss(animated: isPresentedControllerHostingType ? true : false)
-            model.isTransitioningFromPopover = !isPresentedControllerHostingType
-        } else if isHostBeingDismissed {
+            rootViewController.dismiss(animated: presentedControllerIsHosting)
+            model.isTransitioningFromPopover = !presentedControllerIsHosting
+        } else if hostingControllerIsBeingDismissed {
             // Sets 'isPresented' to false when the hosting controller is being dismissed.
-            isPresented = false
-            model.isTransitioningFromPopover = !isPresentedControllerHostingType
+            Task {
```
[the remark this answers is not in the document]
CalebRas: I wasn't able to reproduce this on iPhone, only on iPad. Can you verify this is fixed now?

Shared/Supporting Files/Extensions/View+Sheet.swift
CalebRas: This seems like it would require a solution similar to the timing bug fix, but I am not sure how I would go about implementing that. Should I just note this limitation as part of the documentation?

Shared/Supporting Files/Extensions/View+Sheet.swift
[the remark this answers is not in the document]
yo1995: I'm still seeing this occasionally (1 in 10) but since it doesn't cause the app to be in an unusable state, I think I'll settle with what we have now.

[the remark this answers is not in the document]
CalebRas: Upping the time on the `Task.sleep` may fix that. I just tried to go with the minimum value where I stopped seeing the issue, but it's so small that there probably a lot of wiggle room before it is visually noticeable.

yo1995 (dismissed):
I've run test on multiple iOS 15 to 17 iPhones and iPads. The change mitigates the sheet dismissal problem. Though there are still some minor bugs, I think it is good to get this in for now.

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -176,28 +179,33 @@ private extension SheetModifier {
...
-                .popover(
-                    isPresented: Binding(
-                        get: { isPresented && !isSheetLayout },
-                        set: { isPresented = $0 }
-                    )
-                ) {
+                .task(id: isPresented) {
+                    if isPresented {
+                        // Sleep to prevent appearing when other content is disappearing.
+                        try? await Task.sleep(nanoseconds: 1000)
```
njarecha, line 185: Generally, putting fixes with time will has other side effects and may manifest differently on different devices and situations. I'm ok with the change if you and @yo1995 are ok with this and have done enough testing.

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -26,6 +26,9 @@ extension View {
     ///   - idealHeight: The ideal height of the popover.
     ///   - onDismiss: A closure to execute when dismissing the sheet.
     ///   - content: A closure returning the content of the sheet.
+    ///
+    /// > Note: This modifier may not work with other UIKit derived views, such as an `Alert`.
+    ///  Using them both in the same view may cause undefined behavior.
```
yo1995, line 31:
```suggestion
    /// - Note: This modifier can have conflict with modal presentation views, such as an alert.
    /// When the sheet is presented, it may cause the "already presenting" problem.
```

- Let's use the recommended formatting in Apple [Markup](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/Note.html#//apple_ref/doc/uid/TP40016497-CH39-SW1) reference.
- [`Alert`](https://developer.apple.com/documentation/swiftui/alert) is deprecated. I rephrased the comment a bit.

Shared/Supporting Files/Extensions/View+Sheet.swift
```diff
@@ -212,9 +217,9 @@ private extension SheetModifier {
             ) {
                 sheetContent
                     .onDisappear {
-                        if !isPresented {
-                            onDismiss?()
-                        }
+                        isSheetVisible = false
```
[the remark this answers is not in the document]
yo1995: I'm not seeing the reappearance on iOS 17 iPhone and iPad. Do you have a screen recording of how it looks like?

Shared/Supporting Files/Extensions/View+Sheet.swift
yo1995: I agree. Using sleep is generally speaking an anti-pattern. Ideally this extension can be refactored but it would require effort that doesn't have much value.

I'm OK with using delay to mitigate the problem temporarily.

CalebRas: From my testing, it seems to work as expected. I will leave it in for now as a temporary solution.

yo1995 (approved):

CalebRas:
Thank you everyone!

PR #313, opened 2023-12-07
[Setup] Add favorites to the Sample Viewer

CalebRas:
## Description

This PR implements the ability to favorite samples in the Sample Viewer similar to the old iOS Sample Viewer. A "Favorites" category was added, where the user can see a list of their favorited samples. Samples can be added by pressing the plus button in the category or by long pressing on a sample row anywhere in the app. Samples in the list can also be removed and rearranged by tapping the "Edit" button in the category. A few other views were modified to best accommodate the new views. 

## Linked Issue(s)

- `swift/issues/4370`

## How To Test

- In the Favorites category, tap on the plus button to add a sample to favorites.
- Tap on "Edit" to reorder/delete the favorited samples.
- Tap and hold on a sample row anywhere in the Sample Viewer to favorite/unfavorite from the context menu.
- Close and reopen the app to ensure the favorited samples persist.

## Screenshots

|Before|After|
|:-:|:-:|
| Simulator Screenshot - iPhone 15 Pro - 2023-12-06 at 16 52 03 | Simulator Screenshot - iPhone 15 Pro - 2023-12-06 at 16 45 06 |

### Other: 
 Simulator Screenshot - iPhone 15 Pro - 2023-12-06 at 16 45 00   Simulator Screenshot - iPhone 15 Pro - 2023-12-06 at 16 56 58


Shared/Supporting Files/Views/SampleLink.swift
```diff
@@ -0,0 +1,144 @@
...
+                if isFavorited {
+                    HStack {
+                        Image(systemName: "star.fill")
+                            .resizable()
+                            .aspectRatio(contentMode: .fit)
+                            .frame(width: 8)
+                            .padding(.leading, -13)
+                        Spacer()
+                    }
+                }
```
CalebRas, line 72: Note to the reviewers: I was trying to emulate Apple Music's favorite star here, but I am not super happy with this solution since it is not future-proof, and it gets cut off when the context menu is showing on iPad. The alternatives are to either have the star as part of the title (then the title is offset compared to all the other unfavorited samples), or to get rid of this completely. Let me know what you think.

Shared/Supporting Files/Views/SampleLink.swift
```diff
@@ -0,0 +1,144 @@
...
+            .animation(.easeOut(duration: 0.2), value: isShowingDescription)
+        }
+    }
+}
+
+extension SampleLink {
+    /// Bolds given text found in the sample link's label.
+    /// - Parameter text: The text to bold.
+    /// - Returns: A new `SampleLink` with the bolded text.
+    func bolding(_ text: String) -> SampleLink {
```
yo1995, line 122: Instead of using a modifier, how about just pass the query as a parameter to the `SampleLink`? A default value can be provided to make the constructor look clean.
I feel it is out of place to pass an variable-length sequence as an argument of a view modifier. Typically it is an option, or a fixed length type.

Shared/Supporting Files/Views/SampleLink.swift
yo1995: Given that we cannot have something like Apple music, here are a few other ideas

- A normal sized star placed like Files app's tags may be acceptable
image
- If it makes the sample name display too short, we may use the info button in an innovative way. Toggle a background yellow color to indicate a sample is favorited? ☆ -> ⭐ 

I realized that the star UI also resembles the messages unread blue dot. I'll try to find if it is sth we can mimic within SwiftUI.

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+    }
+}
+
+private extension FavoritesView {
+    /// A view to add a favorite sample from a searchable list.
+    struct AddFavoriteView: View {
+        /// All samples retrieved from the Samples directory.
+        let samples: [Sample]
```
yo1995, line 68: I think the current list is OK. It would be better to filter out already favorited samples from the list though.

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,153 @@
...
+        .listStyle(.sidebar)
+        .toolbar {
+            ToolbarItemGroup(placement: .navigationBarTrailing) {
+                EditButton()
+                
+                Button {
+                    isShowingSheet = true
+                } label: {
+                    Image(systemName: "plus")
+                }
```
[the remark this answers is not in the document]
yo1995: Confirmed on iOS 17 for me as well 🐛

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,153 @@
...
+
+extension UserDefaults {
+    /// The key to read and write favorited sample names to the user defaults.
+    static var favoritedSamplesKey: String {
+        "favoritedSamples"
+    }
+}
+
+/// An extension allowing an array to be used with the app storage property wrapper.
+extension Array: RawRepresentable where Element: Codable {
```
yo1995, line 136:
```suggestion
extension Array: RawRepresentable where Element == String {
```
Since we are only using this extension here, let's limit it to the very specific typing requirements so it doesn't cause too much a side effect.

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,153 @@
...
+    /// Creates a new array from a given raw value.
+    /// - Parameter rawValue: The raw value of the array to create.
+    public init?(rawValue: String) {
+        guard let data = rawValue.data(using: .utf8),
+              let result = try? JSONDecoder().decode([Element].self, from: data)
+        else { return nil }
```
yo1995, line 142: I think there is some disparity here - the default encoded value is an empty array JSON, while the default decoded value is `nil`. What about both using empty array as default?

yo1995 (commented):
I'll add @dfeinzimer to the reviewers as well so we have more thoughts on the UI.

Shared/Supporting Files/Views/FavoritesView.swift
CalebRas: The full list is passed here because it is needed for the `AddFavoriteView`. Also, it makes so `CategoriesView` doesn't have handle an `AppStorage` variable, which is a bit unrelated to showing the categories. (This will be updated as part of `swift/issues/4842`, after this is merged.)

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,155 @@
...
+        
+        /// The search query in the search bar.
+        @State private var query = ""
+        
+        /// The samples displayed in the search list.
+        var displayedSamples: [Sample] {
+            if query.isEmpty {
+                return samples
+            } else {
+                return samples.filter { $0.name.contains(query) }
```
dfeinzimer, line 84: Use this to get case insensitive search.

```suggestion
                return samples.filter { $0.name.localizedStandardContains(query) }
```

Shared/Supporting Files/Views/CategoriesView.swift
```diff
@@ -55,47 +50,42 @@ struct CategoriesView: View {
...
-            }
-            .listStyle(.sidebar)
-        }
-    }
-    
-    struct CategoryTitleView: View {
-        /// The category name used to load the images from assets.
-        let category: String
+        /// The destination view for the category tile to present.
+        @ViewBuilder let destination: Content
```
yo1995, line 58: Hi @philium I'd like some advice here.

Another way would be passing a closure to build the view on site such as `@ViewBuilder let destination: () -> Content`. However I struggle to think of an advantage of doing that in this particular case. Do you recommend one way over another? Thank you.

philium: This seems to be re-inventing the navigation link wheel a bit here. I would suggest simplifying the tile to just be the tile view and then have the tile view be the label of the navigation link:

```swift
NavigationLink {
    SampleList(samples: samples(inCategory: category)
} label: {
    CategoryTile(name: category)
}
```

Having the navigation link embedded in the category tile view seems to complicate the tile view unnecessarily.

njarecha (changes requested):
Initial feedback. I'm looking a way to have favorite button after opening a sample as that is the most common way of making the sample favorite.

Shared/Supporting Files/Views/SampleLink.swift
```diff
@@ -0,0 +1,129 @@
...
+                    }
+                }
+                Spacer()
+                
+                if isFavorited {
+                    Image(systemName: "star.fill")
+                        .resizable()
+                        .aspectRatio(contentMode: .fit)
+                        .frame(width: 12)
+                }
```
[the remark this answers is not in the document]
CalebRas: Good idea!

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,155 @@
...
+            .searchable(text: $query, placement: .navigationBarDrawer(displayMode: .always))
+        }
+    }
+}
+
+// MARK: User Defaults
+
+extension UserDefaults {
+    /// The key to read and write favorited sample names to the user defaults.
+    static var favoritedSamplesKey: String {
```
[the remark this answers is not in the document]
CalebRas: Makes sense! Are you suggesting that I use `favorite` instead of `favorited` everywhere or just here?

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,155 @@
...
+        "favoritedSamples"
+    }
+}
+
+/// An extension allowing an array to be used with the app storage property wrapper.
+extension Array: RawRepresentable where Element == String {
```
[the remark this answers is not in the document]
CalebRas: Good call. I wasn't sure if I should add the key extension in the sample file, so let me know what you think.

Shared/Supporting Files/Views/FavoritesView.swift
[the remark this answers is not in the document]
njarecha: `favorite` or `favorites` whatever is application for specific variable.

Shared/Supporting Files/Views/SampleLink.swift
```diff
@@ -0,0 +1,129 @@
...
+                        }
+                    } else {
+                        favoritedNames.append(String(name.characters))
+                    }
+                } label: {
+                    if isFavorited {
+                        Label("Unfavorite", systemImage: "star.slash")
+                    } else {
+                        Label("Favorite", systemImage: "star")
+                    }
```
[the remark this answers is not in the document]
yo1995: While it might be a bit more verbose, consider using [`SymbolVariants`](https://developer.apple.com/documentation/swiftui/symbolvariants) such as

```swift
Label("Unfavorite", systemImage: "star")
    .symbolVariant(.slash)
```

It would make it easier if we need to further tune the symbols among platforms in the future.

njarecha:
There should be an easy way in the app to make any sample as favorite. There are examples like `News` or `Developer` app which adds a button along with navigation links back button. 

Screenshot 2023-12-13 at 3 47 54 PM

I tried the idea in the `SampleDetailView` and it seems to work. Please try this and see if we want to go with the idea.

```swift
struct SampleDetailView: View {
    /// The action to dismiss the sheet.
    @Environment(\.dismiss) private var dismiss: DismissAction
    
    /// The sample to display in the view.
    private let sample: Sample
    
    /// A Boolean value that indicates whether to present the sample's information view.
    @State private var isSampleInfoViewPresented = false
    
    /// An object to manage on-demand resources for a sample with dependencies.
    @StateObject private var onDemandResource: OnDemandResource
    
    /// The names of the favorited samples loaded from user defaults.
    @AppStorage(UserDefaults.favoritedSamplesKey) private var favoritedNames: [String] = []
    
    /// A Boolean value indicating whether the sample is favorited.
    private var isFavorited: Bool {
        favoritedNames.contains(sample.name)
    }
    
    /// A Boolean value indicating whether a sample should use on-demand resources.
    var usesOnDemandResources: Bool {
#if targetEnvironment(macCatalyst)
        // Mac Catalyst isn't supported by `NSBundleResourceRequest`. Instead, Xcode
        // put the offline data into the app bundle on Mac Catalyst.
        return false
#else
        return sample.hasDependencies
#endif
    }
    
    init(sample: Sample) {
        self.sample = sample
        self._onDemandResource = StateObject(
            wrappedValue: OnDemandResource(tags: [sample.nameInUpperCamelCase])
        )
    }
    
    var body: some View {
        Group {
            if usesOnDemandResources {
                // 'onDemandResource' is created in this branch.
                Group {
                    switch onDemandResource.requestState {
                    case .notStarted, .inProgress:
                        VStack {
                            ProgressView(onDemandResource.progress)
                            Button("Cancel") {
                                onDemandResource.cancel()
                            }
                        }
                        .padding()
                    case .cancelled:
                        VStack {
                            Image(systemName: "nosign")
                            Text("On-demand resources download canceled.")
                        }
                        .padding()
                    case .error:
                        VStack {
                            Image(systemName: "x.circle")
                            Text(onDemandResource.error!.localizedDescription)
                        }
                        .padding()
                    case .downloaded:
                        sample.makeBody()
                    }
                }
                .task {
                    guard case .notStarted = onDemandResource.requestState else { return }
                    await onDemandResource.download()
                }
            } else {
                // 'onDemandResource' is not created in this branch.
                sample.makeBody()
            }
        }
        .navigationTitle(sample.name)
        .navigationBarBackButtonHidden(true)
        .navigationBarTitleDisplayMode(.inline)
        .toolbar {
            ToolbarItem(placement: .navigationBarLeading) {
                HStack(spacing: 0) {
                    Button {
                        dismiss()
                    } label: {
                        Label("Back", systemImage: "chevron.backward")
                    }
                    
                    Button {
                        if isFavorited {
                            guard let index = favoritedNames.firstIndex(of: sample.name) else {
                                return
                            }
                            favoritedNames.remove(at: index)
                        } else {
                            favoritedNames.append(sample.name)
                        }
                    } label: {
                        if isFavorited {
                            Image(systemName: "star.fill")
                                .resizable()
                                .aspectRatio(contentMode: .fit)
                                .frame(width: 18)
                                .foregroundColor(.yellow)
                        } else {
                            Image(systemName: "star")
                                .resizable()
                                .aspectRatio(contentMode: .fit)
                                .frame(width: 18)
                        }
                    }
                }
            }
            
            ToolbarItem(placement: .navigationBarTrailing) {
                Button {
                    isSampleInfoViewPresented = true
                } label: {
                }
                .sheet(isPresented: $isSampleInfoViewPresented) {
                    if #available(iOS 16, *) {
                        NavigationStack {
                            SampleInfoView(sample: sample)
                        }
                    } else {
                        NavigationView {
                            SampleInfoView(sample: sample)
                        }
                    }
                }
            }
        }
    }
}
```

Simulator Screenshot - iPhone 15 Pro Max - 2023-12-13 at 14 58 12
Simulator Screenshot - iPhone 15 Pro Max - 2023-12-13 at 14 58 19

Based on discussion with @yo1995 and @CalebRas, we did not add the button to the navigation bar due to space constraint. The two buttons (custom back button and star button uses less space than system provided back button. There are samples with long names and their names are not fully displayed without favorite button.

Shared/Supporting Files/Extensions/Favorites.swift
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+
+extension String {
+    /// The key to read and write the names of the favorite samples to the user defaults.
+    static var favoriteSamplesKey: String {
+        "favoriteSamples"
+    }
```
njarecha, line 21: Does prefix `key` needed?

```suggestion
    static var favoriteSampleNames = "favoriteSampleNames"
```

Shared/Supporting Files/Extensions/Favorites.swift
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/// An extension allowing an array to be used with the app storage property wrapper.
+extension Array: RawRepresentable where Element == String {
```
njarecha, line 25: The file name should be `Array.swift`

Shared/Supporting Files/Views/FavoritesView.swift
```diff
@@ -0,0 +1,124 @@
...
+
+struct FavoritesView: View {
+    /// All samples retrieved from the Samples directory.
+    let samples: [Sample]
+    
+    /// A Boolean value indicating whether the add favorite sheet is showing.
+    @State private var addFavoriteSheetIsShowing = false
+    
+    /// The names of the favorite samples loaded from user defaults.
+    @AppStorage(.favoriteSamplesKey) private var favoriteNames: [String] = []
```
njarecha, line 25:
```suggestion
    @AppStorage(. favoriteSampleNames) private var favoriteSampleNames: [String] = []
```

Shared/Supporting Files/Views/SampleLink.swift
```diff
@@ -0,0 +1,135 @@
...
+                            .transition(.move(edge: .top).combined(with: .opacity))
+                    }
+                }
+                Spacer()
+                
+                if sampleIsFavorite {
+                    Image(systemName: "star.fill")
+                        .resizable()
+                        .aspectRatio(contentMode: .fit)
+                        .frame(width: 12)
```
njarecha, line 92: This size looks better
```suggestion
                        .frame(width: 18)
```
Simulator Screenshot - iPhone 15 Pro Max - 2023-12-13 at 16 29 31

njarecha (changes requested):
Some more suggestions

Shared/Supporting Files/Extensions/Favorites.swift
CalebRas: This there value to even having this if the title is the same as the value?

CalebRas:
> Based on discussion with @yo1995 and @CalebRas, we did not add the button to the navigation bar due to space constraint.

I was against having the button in the toolbar because it makes it appear cluttered. I would be more in favor of keeping the functionality to favorite in-sample that you propose and wrapping it in a `Menu` with everything else under a single button. That way, the toolbars remain clean, we can reuse the same actions for the `contextMenu`, and have a space to add additional functionality in the future. Thoughts @Esri/swift-samples-reviewers?

Simulator Screenshot - iPhone 15 Pro - 2023-12-13 at 16 48 52

Shared/Supporting Files/Extensions/Favorites.swift
njarecha: This is just a key which will be stored in the UserDefaults and the value will be an array of favorite sample names. We are defining this in the extension because it is used at multiple places.

njarecha (dismissed):
Looks good, adding favorite button for the sample can be done in a separate PR.

yo1995 (approved):

yo1995 (approved):

njarecha (approved):

PR #318, opened 2023-12-13
[Update] Sample toolbars style

CalebRas:
## Description

This PR updates sample toolbars to standardize their style and usage across the samples. Most of the changes fall under one of the following standards: 

- Items
    - `ToolbarItem` should be used for toolbars containing a single component.
    - Items should be intractable. 
    - Generally, the alignment of single items in a toolbar should be left up to the system. 
    - Where space permits, pickers should be placed directly in the toolbar (not in a `Menu`) for a better interface on Mac Catalyst.
    - No more than one item should be added to the top toolbar in a sample to reduce clutter on iOS.
- Naming
   - “Settings” should be used instead of “Options”.
   - Where space permits, the label of a settings button should be descriptive, e.g., “Visibility Settings”.
- Toggles
    - Toggles in the toolbar should use the default style (no switches).
    - Toggle names should dynamically indicate whether they are true or false to support Mac Catalyst, e.g., "Max Extent On".

**Note:** These standards are up for debate, and I would appreciate feedback.

## Linked Issue(s)

- `swift/issues/4851`

## How To Test

- Ensure updated samples work as expected on all platforms.

## Discussion

- Originally, this PR started out as an effort to center all the single toolbar items by changing the `ToolbarItem`s to `ToolbarGroup`s since there was some discrepancy across different OSs. This was determined to be an anti-pattern, and the alignment should instead be left up to the system. See the issue for more details.


philium:
> * Toggle names should dynamically indicate whether they are true or false to support Mac Catalyst, e.g., "Max Extent On".

I would suggest using enabled/disabled rather than on/off (ie. "Max Extend Enabled").

des12437 (approved):
Changes look good on iOS and Mac Catalyst.

Shared/Samples/Show viewshed from point in scene/ShowViewshedFromPointInSceneView.swift
```diff
@@ -37,8 +37,7 @@ struct ShowViewshedFromPointInSceneView: View {
                     .background(.thinMaterial, ignoresSafeAreaEdges: .horizontal)
             }
             .toolbar {
-                ToolbarItemGroup(placement: .bottomBar) {
-                    Spacer()
+                ToolbarItem(placement: .bottomBar) {
```
mhdostal, line 40: Is there a reason this is centered (in the image below) whereas "Light and Shadow" and "Orbit Around Plan" buttons are right-aligned?

CalebRas: My understanding is that is intended SwiftUI behavior. Starting on iOS 16, if a picker is centered, the following happens. That is why we decided that to leave the item alignment up to the system as there could be more breaking changes like this in the future.

https://github.com/Esri/arcgis-maps-sdk-swift-samples/assets/34120440/09201655-8b03-44f3-919c-393a20ff7356

mhdostal (approved):

philium:
> I would suggest using enabled/disabled rather than on/off (ie. "Max Extent Enabled").

I should clarify that in most (if not all) cases, Enabled shouldn't be needed as a suffix. While I couldn't find any explicit recommendations in the HIG, a cursory glance at how switches are used in iOS shows that toggle labels typically do not include the state as part of the label. The state is a function of the switch itself and including it in the label is odd, especially for VoiceOver. If it must be included in the label, it should be static, ie. it shouldn't change depending on the label's state.

yo1995:
>  If it must be included in the label, it should be static, ie. it shouldn't change depending on the label's state.

Thanks for the suggestion. Issue logged at #321

PR #323, opened 2023-12-18
[Update] Samples list accessing

CalebRas:
## Description

This PR updates how the list of samples is accessed by views within the Sample Viewer. Before, the list was passed down from `SamplesApp` as a parameter to the subsequent child views until it was eventually used. Those parameters were removed, and now, the views that need the list of samples access it through the static variable, ` SamplesApp.samples`.

## Linked Issue(s)

- `swift/issues/4842`

## How To Test

- Ensure the Sample Viewer works the same as before, particularly the categories, favorites, and searching.

### DIFF
--- a/Shared/SamplesApp.swift
+++ b/Shared/SamplesApp.swift
@@ -23,7 +23,7 @@ struct SamplesApp: App {
     
     var body: some SwiftUI.Scene {
         WindowGroup {
-            ContentView(samples: Self.samples)
+            ContentView()
         }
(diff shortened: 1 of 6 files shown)


yo1995 (approved):
Changes look good to me. I think this is a harmless change. In the past we are just passing a reference to the static property `SamplesApp.samples` across views, while now we access the static property directly.

pgruenler (approved):

PR #325, opened 2023-12-19
[Update] Refactor DownloadVectorTilesToLocalCacheView

### LINKED ISSUE #114: Fix the download conditional logic in `DownloadVectorTilesToLocalCacheView`
See https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/107#discussion_r1091301687

yo1995:
## Description

This PR refactors `DownloadVectorTilesToLocalCacheView`. Close #114 

## Linked Issue(s)

- https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/107#discussion_r1091301687
- #114 

## How To Test

- Run the sample and ensure the vector tiles download still works.
- Use Network Link Conditioner or other ways to emulate these scenarios
    - The basemap loads too slowly, so the `exportVectorTilesTask` is `nil`
    - The `exportVectorTilesTask` loads too slowly, so its load status is not `loaded`
    - In both cases, the download button should be disabled
- When reviewing code, you can hide whitespaces to simplify the review process

### DIFF
--- a/Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
+++ b/Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
@@ -233,8 +233,11 @@ private extension DownloadVectorTilesToLocalCacheView {
             self.exportVectorTilesTask = exportVectorTilesTask
         }
         
-        /// Downloads the vector tiles within the area of interest.
-        /// - Parameter extent: The area of interest's envelope to download vector tiles.
+        /// Downloads the vector tiles within the area of interest at given scale.
+        /// - Parameters:
+        ///   - extent: The area of interest's envelope to export vector tiles.
+        ///   - maxScale: The map scale which determines how far in to export
+        ///   the vector tiles. Set to `0` to include all levels of detail.
         func downloadVectorTiles(extent: Envelope, maxScale: Double) async throws {
             // Creates the parameters for the export vector tiles job.
             let parameters = try await exportVectorTilesTask.makeDefaultExportVectorTilesParameters(
--- a/Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
+++ b/Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
@@ -19,7 +19,7 @@ struct DownloadVectorTilesToLocalCacheView: View {
     /// A Boolean value indicating whether to download vector tiles.
     @State private var isDownloading = false
     
-    /// A Boolean value indicating whether to cancel and the job.
+    /// A Boolean value indicating whether to cancel the job.
     @State private var isCancellingJob = false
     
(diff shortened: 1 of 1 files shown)


Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
```diff
@@ -33,10 +33,10 @@ struct DownloadVectorTilesToLocalCacheView: View {
     
     var body: some View {
         GeometryReader { geometry in
-            MapViewReader { mapView in
+            MapViewReader { mapViewProxy in
                 MapView(map: model.map)
                     .interactionModes(isDownloading ? [] : [.pan, .zoom])
-                    .onScaleChanged { model.maxScale = $0 * 0.1 }
+                    .onScaleChanged { model.updateMaxScale($0) }
```
yo1995, line 39: There is a problem within Swift compiler that doesn't allow me to do the shorthand 🤷 

https://forums.swift.org/t/converting-function-value-of-type-mainactor-to-void-loses-global-actor-mainactor/59155

```swift
.onScaleChanged(model.updateMaxScale)
```

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
```diff
@@ -278,6 +287,14 @@ private extension DownloadVectorTilesToLocalCacheView {
             try? FileManager.default.removeItem(at: styleTemporaryURL)
         }
         
+        /// Updates the export's max scale.
+        func updateMaxScale(_ maxScale: Double) {
```
CalebRas, line 291: Do you need a doc comment for the `maxScale` parameter here?

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
```diff
@@ -103,19 +103,19 @@ struct DownloadVectorTilesToLocalCacheView: View {
...
+                                    // Shows results when the download finishes.
                                     isShowingResults = true
                                     // Sets downloading to false when the download finishes.
                                     isDownloading = false
                                 } catch {
                                     // Shows an alert if any errors occur.
                                     self.error = error
-                                    // Sets downloading to false when the download finishes.
+                                    // Sets downloading to false if any errors occur.
                                     isDownloading = false
```
CalebRas, line 119: Would it be better to defer `isDownload` in the `do` block instead of setting false both places?

Shared/Samples/Download vector tiles to local cache/DownloadVectorTilesToLocalCacheView.swift
yo1995: I adjusted the logic here to get the scale from the view.

CalebRas (approved):
Looks good!

njarecha (approved):
Looks good

PR #330, opened 2024-01-11
[New] Search for web map

CalebRas:
## Description

This PR implements `Search for web map` in `Cloud and Portal` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-SearchForWebMap/Shared/Samples/Search%20for%20web%20map)

## Linked Issue(s)

- `swift/issues/2260`

## How To Test

- Enter a query into the search bar.
- Tap on a result to see the map for the given portal item.
- Scroll to the bottom of the list to load more results.

## Screenshots

search-for-web-map


yo1995:
I'll review in lieu of Ryan as he's working on reviews this week.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,121 @@
...
+            
+            // Create the portal query parameters with the strings.
+            return PortalQueryParameters(query: "\(query) AND \(typeString) AND \(dateString)")
+        }
+    }
+}
+
+private extension Date {
+    /// The milliseconds between the date value and 00:00:00 UTC on 1 January 1970.
+    var millisecondsSince1970: Int64 {
```
yo1995, line 118: I was kinda surprised to find out this is the standard way to do date search with ArcGIS REST API. Probably worth paraphrasing from the source https://developers.arcgis.com/rest/users-groups-and-items/search-reference.htm#ESRI_SECTION1_B20F9637ADF04AD2B88172D43AA6FC42 to add some comment on why this is needed.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,121 @@
...
+            // Note: Web maps authored prior to July 2nd, 2014 are not supported.
+            let startDate = try Date(
+                "July 2, 2014",
+                strategy: Date.FormatStyle().day().month().year().parseStrategy
+            )
+            let dateRange = startDate.millisecondsSince1970...Date.now.millisecondsSince1970
+            let dateString = "uploaded:[\(dateRange.lowerBound) TO \(dateRange.upperBound)]"
+            
+            // Create a string to filter for web maps.
+            let typeString = "type:\"Web Map\""
```
yo1995, line 108:
```suggestion
            let typeString = #"type:"Web Map""#
```
Use raw string to make it look cleaner

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,121 @@
...
+    /// The view model for the sample.
+    class Model: ObservableObject {
+        /// A portal to ArcGIS Online to get the portal items from.
+        private let portal = Portal.arcGISOnline(connection: .anonymous)
+        
+        /// The query parameters for the next set of results based on the last results.
+        private var nextQueryParameters: PortalQueryParameters?
+        
+        /// The task used to find portal times thorough the portal.
+        private var task: Task<Void, Error>?
```
yo1995, line 28: I still tend to use https://developer.apple.com/documentation/swiftui/view/task(id:priority:_:) instead of a self managed task here. See comments below.

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,180 @@
...
+                LazyVStack {
+                    ForEach(model.portalItems, id: \.id) { item in
+                        NavigationLink {
+                            SafeMapView(map: Map(item: item))
+                                .navigationTitle(item.title)
+                        } label: {
+                            PortalItemRowView(item: item)
+                        }
+                        .buttonStyle(.plain)
+                        .onAppear {
```
yo1995, line 40: Change this to a task modifier.

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,180 @@
...
+            }
+            .opacity(!model.isLoadingResults && !query.isEmpty && model.portalItems.isEmpty ? 1 : 0)
+        }
+        .background(Color(.secondarySystemBackground))
+        .searchable(
+            text: $query,
+            placement: .navigationBarDrawer(displayMode: .always),
+            prompt: "Search for a web map"
+        )
+        .onChange(of: query) { _ in
```
yo1995, line 72: Change this to a `task(id:priority:_:)` modifier. If the id value changes, SwiftUI cancels and restarts the task.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,121 @@
...
+                strategy: Date.FormatStyle().day().month().year().parseStrategy
+            )
+            let dateRange = startDate.millisecondsSince1970...Date.now.millisecondsSince1970
+            let dateString = "uploaded:[\(dateRange.lowerBound) TO \(dateRange.upperBound)]"
+            
+            // Create a string to filter for web maps.
+            let typeString = "type:\"Web Map\""
+            
+            // Create the portal query parameters with the strings.
+            return PortalQueryParameters(query: "\(query) AND \(typeString) AND \(dateString)")
```
yo1995, line 111:
```suggestion
            let fullQuery = [query, typeString, dateString].joined(separator: " AND ")
            return PortalQueryParameters(query: fullQuery)
```

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,121 @@
...
+        /// The portal query parameters for a given query.
+        /// - Parameter query: The text query used to create the parameters.
+        /// - Returns: A new `PortalQueryParameters` object.
+        private func queryParameters(for query: String) throws -> PortalQueryParameters {
+            // Create a date string for a date range to search within.
+            // Note: Web maps authored prior to July 2nd, 2014 are not supported.
+            let startDate = try Date(
+                "July 2, 2014",
+                strategy: Date.FormatStyle().day().month().year().parseStrategy
+            )
```
yo1995, line 103: Maybe worth putting it as a static extension on Date to avoid calculating it each time a query is created.

yo1995 (commented):
Would you mind also adding the category to the design? Thank you. (common-samples/tree/main/designs/SearchForWebMap)

Some initial feedback

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
CalebRas: I didn't go with the `.task` modifiers for a couple of reasons:

1. There should probably only be one search task going at a time. If there were two task modifiers and the user were to start loading the next results and then quickly change the query, the next results would load with the new results from the new search.
2. Task modifiers run when the view appears, so if there was one watching the query, it would reload the results unnecessarily when the user entered and exited a map even if the query didn't change, causing the user to lose their place in the results list.
3. IMO, this solution is more readable and less complex as it doesn't require extra state variables to get the task modifiers to play nice.

Thoughts?

yo1995: Well, I think only having 1 task makes sense. If you go this route, please make sure the errors from each Task block are thrown, or handled within the model. Currently the errors are "swallowed" by the Task blocks.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,127 @@
...
+extension SearchForWebMapView {
+    /// The view model for the sample.
+    class Model: ObservableObject {
+        /// A portal to ArcGIS Online to get the portal items from.
+        private let portal = Portal.arcGISOnline(connection: .anonymous)
+        
+        /// The query parameters for the next set of results based on the last results.
+        private var nextQueryParameters: PortalQueryParameters?
+        
+        /// The task used to find portal times thorough the portal.
```
pgruenler, line 27:
```suggestion
        /// The task used to find portal items through the portal.
```

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,189 @@
...
+                        } catch {
+                            self.error = error
+                        }
+                    }
+                    .errorAlert(presentingError: $error)
+                    .onChange(of: error == nil) { _ in
+                        // Dismiss the view once the error alert is dismissed.
+                        guard error == nil else { return }
+                        dismiss()
+                    }
```
yo1995, line 125: If we call dismiss on `SafeMapView`, then both this view and the alert will be dismissed, as the alert is a subview to this view.

I don't think the auto-dismiss is too important. Also I think it is fine to not use `errorAlert`, but to implement a separate one for this view, so we have control over the done action.

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,189 @@
...
+                MapView(map: map)
+                    .task {
+                        mapIsLoading = true
+                        defer { mapIsLoading = false }
+                        
+                        // Show an alert for an error loading the map.
+                        do {
+                            try await map.load()
+                        } catch {
+                            self.error = error
```
yo1995, line 117: I'm discussing the error handling with Nimesh - there are lots of items on public AGOL where they use an Esri basemap, and some private layers, all in the same item. It causes  authentication problems that says

```
ArcGIS Maps SDK
Invalid API Key Error
The API key '{XXX}' is invalid, expired, or does not have access to the resource '{XXX}'.
Please visit https://developers.arcgis.com/api-keys to configure the API key to access the resource, or to create a new API key.
```

Ideally we can capture those errors as well, in addition to map load error.

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,189 @@
...
+                    
+                    Text(item.title)
+                    
+                    Spacer()
+                }
+                
+                HStack {
+                    Text(item.modificationDate?.formatted(
+                        Date.FormatStyle(date: .abbreviated, time: .omitted)
+                    ) ?? "")
```
yo1995, line 161: Maybe…?
```suggestion
                    ) ?? "Last Modified: Unknown")
```

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,189 @@
...
+                        image
+                            .resizable()
+                            .scaledToFit()
+                            .frame(height: 75)
+                    } placeholder: {
+                        Color(.lightGray)
+                            .frame(width: 110, height: 75)
+                    }
+                    .border(.primary)
+                    .padding([.leading, .top], 10)
```
yo1995, line 151:
```suggestion
                    AsyncImage(url: item.thumbnail?.url) { image in
                        image
                            .resizable()
                    } placeholder: {
                        Color(.lightGray)
                    }
                    .frame(width: 110, height: 75)
                    .border(.primary)
                    .padding([.leading, .top], 10)
```

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
yo1995: I'd suggest either trying out the task modifier route, or handle the error within the model. The problem with managing Task block within the model is that, it basically turns methods that should be `async` into synchronous, and we don't have control over their lifespans.

You may try sth like below that handles the result of the task each time it is changed, but this IMHO is pretty close to just calling the async methods in task modifiers 😹 

```swift
.task(id: model.task) {
    guard let task = model.task else { return }
    let result = await task.result
    switch result {
    case .success(let success):
        <#code#>
    case .failure(let failure):
        <#code#>
    }
}
```

Or you may use a `do {} catch {}` block with every throwing method, and store a published error on the model.

Shared/Samples/Search for web map/SearchForWebMapView.swift
philium: `formatted(…)` methods should only be used if there isn't an alternative. In this case, a better way would be to provide the value and format style via a `Text` initializer. That way, SwiftUI is able to efficiently update the view (only when the value and/or style changes, not every time the view is rebuilt).

```swift
if let modificationDate = item.modificationDate {
    Text(modificationDate, format: ...)
} else {
    Text("Unknown")
}
```

CalebRas: I used `formatted(…)` because I was trying avoid conditionally showing views. Does that not apply in this case

philium: Why avoid that? It shouldn't be done unnecessarily, but this is a case where it makes a lot of sense to do so.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,120 @@
...
+    /// A pair of two equatable types.
+    struct Pair<T: Equatable, U: Equatable>: Equatable {
+        var t: T
+        var u: U
+        
+        init(_ t: T, _ u: U) {
+            self.t = t
+            self.u = u
+        }
+    }
```
dfeinzimer, line 106: Is this still in use?

```suggestion
```

CalebRas: No, thank you

Shared/Samples/Search for web map/SearchForWebMapView.Views.swift
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+            .onAppear {
+                // Update the challenger handler to get any errors thrown.
+                ArcGISEnvironment.authenticationManager.arcGISAuthenticationChallengeHandler = errorHandler
+            }
+            .onDisappear {
+                // Reset the authentication challenge handler to use default handling.
+                ArcGISEnvironment.authenticationManager.arcGISAuthenticationChallengeHandler = nil
+            }
```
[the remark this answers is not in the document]
yo1995: Another thought would be using [`onLayerViewStateChanged(perform:)`](https://developers.arcgis.com/swift/api-reference/documentation/arcgis/mapview/onlayerviewstatechanged(perform:)) as pointed out by Phil G. As long as some layer are giving errors of `ArcGISAuthenticationError` type, this view can be destroyed.

yo1995 (commented):
More comments on authentication and error handling.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,109 @@
...
+        @Published private(set) var portalItems: [PortalItem] = []
+        
+        /// Finds the portal items that match the given query.
+        /// - Parameter query: The text query used to find the portal items.
+        func findItems(for query: String) async throws {
+            // Ensure that a new search is necessary.
+            guard query != lastQuery else { return }
+            
+            if query.isEmpty {
+                await MainActor.run {
```
yo1995, line 40: While it is great to have a clear sense to only use main thread to update UI and use arbitrary threads to run other async tasks, I prefer to simply mark the whole model as `@MainActor` instead.

Shared/Samples/Search for web map/SearchForWebMapView.Model.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+}
+
+private extension Date {
+    /// The date after which web maps are supported, July 2, 2014.
+    static let webMapSupportedDate = try? Date(
+        "July 2, 2014",
+        strategy: Date.FormatStyle().day().month().year().parseStrategy
+    )
```
yo1995, line 103: Maybe
```suggestion
    /// The date after which web maps are supported, July 2, 2014.
    static var webMapSupportedDate: Date {  // swiftlint:disable:next force_try
        try! Date(
            "July 2, 2014",
            strategy: Date.FormatStyle().day().month().year().parseStrategy
        )
    }
```

yo1995 (commented):
Looks good overall!

yo1995 (dismissed):

Shared/Samples/Search for web map/SearchForWebMapView.swift
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                    .padding()
+                }
+            }
+        }
+        .background(Color(.secondarySystemBackground))
+        .searchable(
+            text: $query,
+            placement: .navigationBarDrawer(displayMode: .always),
+            prompt: "Search for a web map"
```
dfeinzimer, line 75: Just a nitpick, but should this match the sample name? 

```suggestion
            prompt: "Search for web map"
```

Shared/Samples/Search for web map/SearchForWebMapView.Views.swift
```diff
@@ -0,0 +1,109 @@
...
+                        .overlay(.black)
+                    
+                    Text(item.owner)
+                        .foregroundColor(.teal)
+                    
+                    Spacer()
+                }
+                .font(.footnote)
+                .padding(10)
+                .background(Color(.darkGray))
```
dfeinzimer, line 101: The number of differing grays and the teal seem visually busy to me. I tweaked some things around and came up with this, take it or leave it.

```suggestion
                
                Divider()
                    .overlay(.black)
                
                HStack {
                    if let modificationDate = item.modificationDate {
                        Text(modificationDate, format: Date.FormatStyle(date: .abbreviated, time: .omitted))
                            .foregroundColor(Color(.systemGray5))
                    } else {
                        Text("Date: Unknown")
                            .foregroundColor(Color(.systemGray5))
                    }
                    
                    Divider()
                        .overlay(.black)
                    
                    Text(item.owner)
                        .foregroundColor(.accentColor)
                    
                    Spacer()
                }
                .font(.footnote)
                .padding()
```
As is:


Suggestion:


The padding in the suggestion could probably be reduced.

dfeinzimer (approved):
Looks good overall, just a couple of things - take or leave them.

Shared/Samples/Search for web map/SearchForWebMapView.swift
CalebRas: Looking to Apple's apps as a guide, it looks like the prompt should probably either be "Search" or what is being searched for, so I went with "Web Maps".

dfeinzimer (approved):

CalebRas:
Thank you guys!

PR #350, opened 2024-03-18
[Update] Replace deprecated safe area API in SampleInfoView

yo1995:
## Description

This PR replace deprecated API.

- [edgesIgnoringSafeArea(_:)](https://developer.apple.com/documentation/swiftui/view/edgesignoringsafearea(_:)) is deprecated, [ignoresSafeArea(_:edges:)](https://developer.apple.com/documentation/swiftui/view/ignoressafearea(_:edges:)) is used instead.

## How To Test

Open a source code pane, the view should still extend all the way to the bounds.

### DIFF
--- a/Shared/Supporting Files/Views/SampleInfoView.swift
+++ b/Shared/Supporting Files/Views/SampleInfoView.swift
@@ -33,7 +33,7 @@ struct SampleInfoView: View {
             WebView(htmlString: codeHTML)
                 .opacity(informationMode == .code ? 1 : 0)
         }
-        .edgesIgnoringSafeArea([.horizontal, .bottom])
+        .ignoresSafeArea(edges: [.horizontal, .bottom])
         .navigationBarTitleDisplayMode(.inline)
         .toolbar {
             ToolbarItem(placement: .principal) {


CalebRas (approved):

Shared/Supporting Files/Views/SampleInfoView.swift
```diff
@@ -33,7 +33,7 @@ struct SampleInfoView: View {
             WebView(htmlString: codeHTML)
                 .opacity(informationMode == .code ? 1 : 0)
         }
-        .edgesIgnoringSafeArea([.horizontal, .bottom])
+        .ignoresSafeArea(edges: [.horizontal, .bottom])
```
CalebRas, line 36: Is this modifier actually needed? I can't see a difference without it...

yo1995: Good question. It makes a tiny difference with/without it. 

Without bottom edge ignore, the README won't overlap with the screen bottom home screen drag indicator.
Without horizontal ignore, when zoom the webview/scroll view in landscape orientation, the content won't extend into left and right.

PR #355, opened 2024-03-21
[Update] Exclude AR samples from Mac Catalyst and Simulator

yo1995:
## Description

This PR uses a conditional view for the "Augmented Reality" samples so they aren't available for Mac Catalyst or Simulator target, as these platforms don't have camera and sensors.

App Store Connect reviewer rejected our Mac Catalyst build for AR sample not running on Mac Catalyst, so we need to make some change in the next submission. 

## Linked Issue(s)

- `swift/issues/5073`

## How To Test

Run the app on Mac Catalyst. AR samples shows a text for being not available.

## To Discuss

I have been thinking about the best way for "excluding" AR samples and couldn't decide which is the best, so this change is open for debate.

1. ⬆️ (Current approach) Add filtering logic to code generation, so AR sample root views aren't generated in the `SamplesApp.samples` static property.

- This is similar to option 3.

2. Display a text to let users who download this app from MAS or build the app on Simulator know that AR isn't available.

- This approach introduces fewest code changes
- However, unused source code is still compiled and included in the Mac Catalyst bundle.

3. Use `targetEnvironment` directive and `filter` method on all `SamplesApp.samples` static properties, so the AR samples aren't included for both targets.

- To make this work, I have to create a separate static property such as below, and swap all `SamplesApp.samples` occurrences with it.
```swift
    static let filteredSamples = SamplesApp.samples
#if targetEnvironment(macCatalyst) || targetEnvironment(simulator)
        .filter { sample in
            return sample.category != "Augmented Reality"
        }
#endif
```
- Unused code is still compiled and included in the bundle. So are the READMEs.
- Even the SwiftUI views for these samples are available, they are just "hidden" from the users.

4. Don't include AR samples' root views in the `GenerateSampleViewSourceCode.swift` script's build rule. Also, exclude the AR source code files from "Compile Sources" and "Copy Source Code Files" phases

image

- In order to exclude the files from compilation, anytime an AR source code is created, the sample author must go to `Build Phases > Compile Sources` to change the platform filter for the swift file. This might be easily overlooked by both an author and a PR reviewer.
- Simulator can still open AR samples, which isn't ideal.
- Also, the `GenerateSampleViewSourceCode.swift` script needs to incorporate a logic to filter out AR samples' metadata, so the excluded view's initializers aren't called in the generated code, which could lead to a build error if not handled correctly.
- Finally, the README copying script needs rework to exclude AR samples from being copied into the bundle.

5. More involved approach: Change the metadata json format to have a conditional `platforms` field; Rewrite many build phase scripts to conditionally exclude files; Rework the `GenerateSampleViewSourceCode` based on the new metadata schema, etc.

### DIFF
--- a/Scripts/GenerateSampleViewSourceCode.swift
+++ b/Scripts/GenerateSampleViewSourceCode.swift
@@ -143,6 +143,11 @@ private let arrayRepresentation = """
     [
             \(entries)
         ]
+    #if targetEnvironment(macCatalyst) || targetEnvironment(simulator)
+        // Exclude AR samples from Mac Catalyst and Simulator targets
+        // as they don't have camera and sensors available.
+        .filter { $0.category != "Augmented Reality" }
+    #endif
     """
 
 do {
--- a/Shared/Supporting Files/Views/CategoriesView.swift
+++ b/Shared/Supporting Files/Views/CategoriesView.swift
@@ -18,10 +18,6 @@ struct CategoriesView: View {
     /// The sample categories generated from the samples list.
     private let categories = Set(
         SamplesApp.samples.map(\.category)
-        #if targetEnvironment(macCatalyst)
-        // Exclude AR samples from Mac Catalyst app as they aren't suitable.
-            .filter { $0 != "Augmented Reality" }
-        #endif
     ).sorted()
     
     var body: some View {
--- a/Shared/Supporting Files/Views/CategoriesView.swift
+++ b/Shared/Supporting Files/Views/CategoriesView.swift
@@ -16,7 +16,13 @@ import SwiftUI
 
 struct CategoriesView: View {
     /// The sample categories generated from the samples list.
-    private let categories = Set(SamplesApp.samples.map(\.category)).sorted()
+    private let categories = Set(
+        SamplesApp.samples.map(\.category)
+        #if targetEnvironment(macCatalyst)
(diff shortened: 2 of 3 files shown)


yo1995:
@Esri/swift-samples-reviewers Feel free to share your thoughts on what is the best solution.

CalebRas:
I agree the solution you currently have is the probably the simplest and easiest. I would be more in favor of showing an alert with a disclaimer here when opening an AR sample on Mac Catalyst or Simulator and still letting the user access the incomplete sample. That way, we don't have to load it on device every run for development/testing of things like the UI. Also, doing so would preserve the View hierarchy on those platforms (that's not a huge deal, though).

philium:
Could we just exclude the Augmented Reality category in `SamplesApp+Samples.swift` like so?

```swift
    static let samples: [Sample] = [
        ...
    ]
#if targetEnvironment(macCatalyst)
    .filter { $0.category != "Augmented Reality" }
#endif
```

yo1995:
> Could we just exclude the Augmented Reality category in `SamplesApp+Samples.swift` like so?

Yes, that is a good option. I can make this change.
I worried that it tucks away some detail in a generated file (not in search results), which could confuse first time users. But that shouldn't be a big deal.

philium (approved):
Works as expected.

zkline101 (approved):

PR #358, opened 2024-03-27
[Fix] Custom Boolean bindings bug on Mac Catalyst

CalebRas:
## Description

This PR implements a fix for a bug where toggle states would not update on Mac Catalyst when a custom binding was used. The bindings were replaced with State variables that trigger updates to a given value through an `onChange` modifier.

Sample affected:
- Display dimensions
- Find route around barriers
- Group layers together

## Linked Issue(s)

- `swift/issues/4994`

## How To Test

- Test the toggles in the samples listed on Mac Catalyst.
- Ensure the samples' functionality is otherwise unchanged.

### DIFF
--- a/Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Views.swift
+++ b/Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Views.swift
@@ -16,42 +16,46 @@ import ArcGIS
 import SwiftUI
 
 extension FindRouteAroundBarriersView {
-    /// The list of settings for the sample.
-    struct SettingsList: View {
-        /// The view model for the sample.
-        @ObservedObject var model: Model
+    /// A list of settings for modifying route parameters.
+    struct RouteParametersSettingsList: View {
+        /// The route parameters to modify.
+        private let routeParameters: RouteParameters
         
         /// A Boolean value indicating whether routing will find the best sequence.
-        @State private var routingFindsBestSequence = false
+        @State private var routingFindsBestSequence: Bool
+        
+        /// A Boolean value indicating whether routing will preserve the first stop.
+        @State private var routePreservesFirstStop: Bool
+        
+        /// A Boolean value indicating whether routing will preserve the last stop.
+        @State private var routePreservesLastStop: Bool
+        
+        init(for routeParameters: RouteParameters) {
+            self.routeParameters = routeParameters
+            _routingFindsBestSequence = State(initialValue: routeParameters.findsBestSequence)
+            _routePreservesFirstStop = State(initialValue: routeParameters.preservesFirstStop)
+            _routePreservesLastStop = State(initialValue: routeParameters.preservesLastStop)
+        }
         
         var body: some View {
             List {
-                Toggle(isOn: $routingFindsBestSequence) {
-                    Text("Find Best Sequence")
-                }
-                .onChange(of: routingFindsBestSequence) { newValue in
-                    model.routeParameters.findsBestSequence = newValue
-                }
+                Toggle("Find Best Sequence", isOn: $routingFindsBestSequence)
+                    .onChange(of: routingFindsBestSequence) { newValue in
+                        routeParameters.findsBestSequence = newValue
+                    }
                 
                 Section {
-                    Toggle(isOn: Binding(
-                        get: { model.routeParameters.preservesFirstStop },
-                        set: { model.routeParameters.preservesFirstStop = $0 }
-                    )) {
-                        Text("Preserve First Stop")
-                    }
+                    Toggle("Preserve First Stop", isOn: $routePreservesFirstStop)
+                        .onChange(of: routePreservesFirstStop) { newValue in
+                            routeParameters.preservesFirstStop = newValue
+                        }
                     
-                    Toggle(isOn: Binding(
-                        get: { model.routeParameters.preservesLastStop },
-                        set: { model.routeParameters.preservesLastStop = $0 }
-                    )) {
-                        Text("Preserve Last Stop")
-                    }
+                    Toggle("Preserve Last Stop", isOn: $routePreservesLastStop)
+                        .onChange(of: routePreservesLastStop) { newValue in
+                            routeParameters.preservesLastStop = newValue
+                        }
                 }
                 .disabled(!routingFindsBestSequence)
-            }
-            .onAppear {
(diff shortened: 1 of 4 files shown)


Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Views.swift
```diff
@@ -16,43 +16,47 @@ import ArcGIS
 import SwiftUI
 
 extension FindRouteAroundBarriersView {
-    /// The list of settings for the sample.
-    struct SettingsList: View {
-        /// The view model for the sample.
-        @ObservedObject var model: Model
+    /// A list of settings for modifying route parameters.
+    struct RouteParametersSettingsList: View {
```
njarecha, line 20: I think this should be just `RouteParametersSettings`
```suggestion
    struct RouteParametersSettings: View {
```

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Views.swift
```diff
@@ -16,43 +16,47 @@ import ArcGIS
...
+        @State private var routePreservesFirstStop: Bool
+        
+        /// A Boolean value indicating whether routing will preserve the last stop.
+        @State private var routePreservesLastStop: Bool
+        
+        init(for routeParameters: RouteParameters) {
+            self.routeParameters = routeParameters
+            _routingFindsBestSequence = State(initialValue: routeParameters.findsBestSequence)
+            _routePreservesFirstStop = State(initialValue: routeParameters.preservesFirstStop)
+            _routePreservesLastStop = State(initialValue: routeParameters.preservesLastStop)
```
njarecha, line 37: Any specific reason to use `State(initialValue:)`?
```suggestion
            self.routingFindsBestSequence = routeParameters.findsBestSequence
            self.routePreservesFirstStop = routeParameters.preservesFirstStop
            self.routePreservesLastStop = routeParameters.preservesLastStop

```

Shared/Samples/Group layers together/GroupLayersTogetherView.GroupLayerListView.swift
```diff
@@ -116,4 +101,29 @@ extension GroupLayersTogetherView {
...
+        /// The title of the toggle.
+        private let title: String
+        
+        /// The layer.
+        private let layer: Layer
+        
+        /// A Boolean value indicating whether the layer's content is visible.
+        @State private var layerIsVisible: Bool
+        
+        init(_ title: String, layer: Layer) {
```
njarecha, line 116: Please add ref doc

Shared/Samples/Group layers together/GroupLayersTogetherView.GroupLayerListView.swift
```diff
@@ -116,4 +101,29 @@ extension GroupLayersTogetherView {
...
+            self.title = title
+            self.layer = layer
+            _layerIsVisible = State(initialValue: layer.isVisible)
```
njarecha, line 119:
```suggestion
            self.layerIsVisible = layer.isVisible
```

njarecha (changes requested):
Few suggestions

Shared/Samples/Group layers together/GroupLayersTogetherView.GroupLayerListView.swift
```diff
@@ -116,4 +101,33 @@ extension GroupLayersTogetherView {
...
+    /// A toggle for changing a given layer's visibility.
+    private struct LayerVisibilityToggle: View {
+        /// The title of the toggle.
+        private let title: String
+        
+        /// The layer with the visibility to change.
+        private let layer: Layer
+        
+        /// A Boolean value indicating whether the layer's content is visible.
+        @State private var layerIsVisible: Bool
```
njarecha, line 114: The `LayerVisibilityToggle` already uses the `Layer` so this should just be called, `isVisible`.

njarecha (approved):

CalebRas:
Thank you guys for reviewing!

Shared/Samples/Find route around barriers/FindRouteAroundBarriersView.Views.swift
philium: [The doc](https://developer.apple.com/documentation/swiftui/state/init(wrappedvalue:)) is pretty clear that state properties should not be set in an initializer, neither by explicitly calling `init(initialValue:)` nor by implicitly calling `init(wrappedValue:)`:

> You don’t call this initializer directly. Instead, SwiftUI calls it for you when you declare a property with the @State attribute and provide an initial value

The reason being is that state is a private, implementation detail. The parent view's body being re-computed shouldn't alter the view's state, which is what happens when a state property is set from inside an initializer. For values provided by an initializer, non-state properties should be used.

CalebRas: @philium sorry I missed your comment. Would it be better to use bindings here in this case?

philium: I took a closer look at what this sample is doing and in this case, I don't see an alternative option that isn't more complicated. As I mentioned, the concern with setting the state in an initializer is that the state will be reset every time the parent's body is re-computed, but in this case, that isn't an issue. (In an ideal world, `RouteParameters` would be a struct and the settings view would take a binding to it. Then everything would just work without needing separate state properties.)

A couple other things I noticed:

- The settings view initializer parameter label should be `routeParameters` instead of `for`. From the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/): "The first argument to initializer…calls should not form a phrase starting with the base name, e.g. x.makeWidget(cogCount: 47)"
- The settings view should use `Form` instead of `List`, as `Form` is "a container for grouping controls used for data entry, such as in settings or inspectors."

CalebRas: @philium Sounds good. I'll make another PR to address those issues.

yo1995: > The settings view should use Form instead of List

here is a [card](https://github.com/Esri/arcgis-maps-sdk-swift-samples/projects/1#card-89771848) for SprintX 😼

PR #361, opened 2024-03-28
[Update] `DowloadPortalItemData` script

CalebRas:
## Description

This PR updates the `DowloadPortalItemData` script to allow items to be deleted and re-downloaded when their entry in the `.downloaded_items.plist` is removed. This also fixes a bug where download failures would result in an empty folder, causing the scrip to think the item successfully downloaded. The PR additionally replaces the dispatch group in the script with async-await syntax.
URL to script: [DowloadPortalItemData](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Caleb/Fix-DowloadPortalItemDataScript/Scripts/DowloadPortalItemData.swift) 

## Linked Issue(s)

- `swift/issues/4891`

## How To Test

- Delete an item from the `.downloaded_items.plist`.
- Build the project and check the local console log to ensure the item re-downloaded.

### DIFF
--- a/Scripts/DowloadPortalItemData.swift
+++ b/Scripts/DowloadPortalItemData.swift
@@ -19,6 +19,7 @@
 // A mapping of item IDs to filenames is maintained in the download directory.
 // This mapping efficiently checks whether an item has already been downloaded.
 // If an item already exists, it will skip that item.
+// To delete and re-downloaded an item, remove it's entry in the plist.
 
 import Foundation
 
--- a/Scripts/DowloadPortalItemData.swift
+++ b/Scripts/DowloadPortalItemData.swift
@@ -161,10 +161,10 @@ func downloadFile(at sourceURL: URL, to downloadDirectory: URL, completion: @esc
                 
                 if isArchive {
                     let extractURL = downloadURL.pathExtension == "zip"
-                        // Uncompresses to directory named after archive.
-                        ? downloadURL.deletingPathExtension()
-                        // Uncompresses to appropriate subdirectory.
-                        : downloadURL.deletingLastPathComponent()
+                    // Uncompresses to directory named after archive.
+                    ? downloadURL.deletingPathExtension()
+                    // Uncompresses to appropriate subdirectory.
+                    : downloadURL.deletingLastPathComponent()
                     try uncompressArchive(at: temporaryURL, to: extractURL)
                 } else {
                     try FileManager.default.moveItem(at: temporaryURL, to: downloadURL)
--- a/Scripts/DowloadPortalItemData.swift
+++ b/Scripts/DowloadPortalItemData.swift
@@ -31,7 +31,7 @@ struct SampleDependency: Decodable {
 }
 
 /// A Portal Item and its data URL.
-struct PortalItem {
+struct PortalItem: Hashable {
     static let arcGISOnlinePortalURL = URL(string: "https://www.arcgis.com")!
     
     /// The identifier of the item.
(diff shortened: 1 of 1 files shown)


Scripts/DowloadPortalItemData.swift
```diff
@@ -241,39 +231,56 @@ let previousDownloadedItems: DownloadedItems = {
...
-    let destinationURL = downloadDirectoryURL.appendingPathComponent(portalItem.identifier, isDirectory: true)
-    // Checks if a directory exists or not, to see if an item is already downloaded.
-    if FileManager.default.fileExists(atPath: destinationURL.path) {
-        print("info: Item \(portalItem.identifier) has already been downloaded.")
-    } else {
+await withTaskGroup(of: Void.self) { group in
+    for portalItem in portalItems {
+        // Checks to see if an item is already downloaded.
+        guard downloadedItems[portalItem.identifier] == nil else {
+            print("note: Item already downloaded: \(portalItem.identifier)")
```
yo1995, line 238: Line 238 and 270 can use the same output phrasing.

Scripts/DowloadPortalItemData.swift
```diff
@@ -218,7 +208,7 @@ let portalItems: [PortalItem] = {
         // Omit the decoding errors from samples that don't have dependencies.
         let sampleDependencies = sampleJSONs
             .compactMap { try? parseJSON(at: $0) }
-        return sampleDependencies.flatMap(\.offlineData)
+        return Set(sampleDependencies.flatMap(\.offlineData))
```
yo1995, line 211: This is a great catch. We used to repeatedly download files…

Scripts/DowloadPortalItemData.swift
```diff
@@ -125,61 +126,50 @@ func uncompressArchive(at sourceURL: URL, to destinationURL: URL) throws {
...
-                completion(.failure(error))
-            }
-        } else if let error = error {
-            completion(.failure(error))
+///   - downloadDirectory: The directory to store the downloaded data in.
+/// - Throws: Exceptions when downloading and moving the file.
+/// - Returns: The name of the downloaded file.
+func downloadFile(from sourceURL: URL, to downloadDirectory: URL) async throws -> String {
+    let portalURLRequest = URLRequest(url: sourceURL)
+    let (temporaryURL, response) = try await URLSession.shared.download(for: portalURLRequest)
```
yo1995, line 137: What about using [func download(from: URL, delegate: (any URLSessionTaskDelegate)?) -> (URL, URLResponse)](https://developer.apple.com/documentation/foundation/urlsession/3767355-download) as we aren't customizing the URL request?

Scripts/DowloadPortalItemData.swift
```diff
@@ -125,61 +126,50 @@ func uncompressArchive(at sourceURL: URL, to destinationURL: URL) throws {
...
+    
+    guard let suggestedFilename = response.suggestedFilename else { return "" }
```
yo1995, line 139: This seems to be a behavioral change. The old script omits the case of no `suggestedFilename` and no error. Here it returns an empty string.

Scripts/DowloadPortalItemData.swift
```diff
@@ -241,39 +231,56 @@ let previousDownloadedItems: DownloadedItems = {
...
-portalItems.forEach { portalItem in
-    let destinationURL = downloadDirectoryURL.appendingPathComponent(portalItem.identifier, isDirectory: true)
-    // Checks if a directory exists or not, to see if an item is already downloaded.
-    if FileManager.default.fileExists(atPath: destinationURL.path) {
-        print("info: Item \(portalItem.identifier) has already been downloaded.")
-    } else {
+await withTaskGroup(of: Void.self) { group in
+    for portalItem in portalItems {
+        // Checks to see if an item is already downloaded.
+        guard downloadedItems[portalItem.identifier] == nil else {
```
yo1995, line 237:
```suggestion
        guard !downloadedItems.keys.contains(portalItem.identifier) else {
```
This might make more sense.

In current version of the script, it doesn't write a k-v pair with `nil` value to the plist, so the change should work just fine. 

However, your condition reads like "if the portal item ID is in the plist, but the value is `nil`, then the item wasn't downloaded." Which doesn't read exactly right.

Scripts/DowloadPortalItemData.swift
```diff
@@ -241,39 +231,56 @@ let previousDownloadedItems: DownloadedItems = {
...
+                }
+            } catch {
                 print("error: Error downloading item \(portalItem.identifier): \(error.localizedDescription)")
                 URLSession.shared.invalidateAndCancel()
                 exit(1)
             }
         }
     }
 }
-dispatchGroup.wait()
```
yo1995, line 276: [func waitForAll() async](https://developer.apple.com/documentation/swift/taskgroup/waitforall()) may be needed for the task group. Otherwise, when the last child task is added, the script moves directly to the `downloadedItems != previousDownloadedItems` equality check below, which might cause problems.

yo1995 (commented):
Some initial feedback below. Thanks for tackling this issue!

I feel like our previous download failures might be due to the repetition of portal item IDs without `Set`, that caused AGOL to limit our high-freq download requests.

- Would you mind also clean up all the verbs in the `//` comments to make them 3rd person singular when suitable? Thanks. 🙇

philium:
There is a lot going on in this PR, more than just fixing the download failure bug. Could the work be broken up? It seems like fixing the bug could be one PR, converting from GCD to Swift concurrency could be another, etc.

CalebRas:
> There is a lot going on in this PR, more than just fixing the download failure bug. Could the work be broken up? It seems like fixing the bug could be one PR, converting from GCD to Swift concurrency could be another, etc.

I'll update this PR to just fix the download failure bug.

CalebRas:
@philium & @yo1995 This PR should only include the fix changes (see updated description). I'm not sure what feedback applies now, so please unresolve any feedback you want me to address as part of this PR.

Scripts/DowloadPortalItemData.swift
```diff
@@ -245,31 +246,46 @@ var downloadedItems = previousDownloadedItems
 let dispatchGroup = DispatchGroup()
 
 portalItems.forEach { portalItem in
```
yo1995, line 248: [`forEach`](https://forums.swift.org/t/pitch-mutating-foreach/65950/10) can be replaced in the follow up PR.

yo1995 (approved):

Scripts/DowloadPortalItemData.swift
```diff
@@ -161,10 +162,10 @@ func downloadFile(at sourceURL: URL, to downloadDirectory: URL, completion: @esc
...
                 if isArchive {
                     let extractURL = downloadURL.pathExtension == "zip"
-                        // Uncompresses to directory named after archive.
-                        ? downloadURL.deletingPathExtension()
-                        // Uncompresses to appropriate subdirectory.
-                        : downloadURL.deletingLastPathComponent()
+                    // Uncompresses to directory named after archive.
+                    ? downloadURL.deletingPathExtension()
+                    // Uncompresses to appropriate subdirectory.
+                    : downloadURL.deletingLastPathComponent()
```
philium, line 168: Inadvertent whitespace changes?

Scripts/DowloadPortalItemData.swift
philium: More important than this is the guard statement inside the `forEach` loop. With that, filtering out duplicates isn't as important. Still good to do, but not as big of a deal.

Scripts/DowloadPortalItemData.swift
```diff
@@ -218,7 +219,7 @@ let portalItems: [PortalItem] = {
         // Omit the decoding errors from samples that don't have dependencies.
         let sampleDependencies = sampleJSONs
             .compactMap { try? parseJSON(at: $0) }
-        return sampleDependencies.flatMap(\.offlineData)
+        return Set(sampleDependencies.flatMap(\.offlineData))
```
philium, line 222: Flat mapping creates an intermediate array. That can be avoided by lazily flat mapping.

```suggestion
        return Set(sampleDependencies.lazy.flatMap(\.offlineData))
```

Scripts/DowloadPortalItemData.swift
```diff
@@ -245,31 +246,46 @@ var downloadedItems = previousDownloadedItems
 let dispatchGroup = DispatchGroup()
 
 portalItems.forEach { portalItem in
+    // Checks to see if an item is already downloaded.
+    guard !downloadedItems.keys.contains(portalItem.identifier) else {
+        print("note: Item already downloaded: \(portalItem.identifier)")
+        return
```
philium, line 252: Two suggestions in one (since they are intertwined):

1. I don't know how efficient `Dictionary.Keys.contains(_:)` is. Better to just subscript into the dictionary.
2. `forEach(_:)` is intended for use at the end of a method chain; a for loop should have been used instead.

```suggestion
for portalItem in portalItems {
    // Checks to see if an item is already downloaded.
    guard downloadedItems[portalItem.identifier] == nil else {
        print("note: Item already downloaded: \(portalItem.identifier)")
        continue
```

Scripts/DowloadPortalItemData.swift
```diff
@@ -245,31 +246,46 @@ var downloadedItems = previousDownloadedItems
...
+    }
+    
     let destinationURL = downloadDirectoryURL.appendingPathComponent(portalItem.identifier, isDirectory: true)
-    // Checks if a directory exists or not, to see if an item is already downloaded.
+    
+    // Deletes the directory if it already exists.
+    // This happens when the item is not in the plist and needs to be redownloaded.
     if FileManager.default.fileExists(atPath: destinationURL.path) {
```
philium, line 259: From [the doc](https://developer.apple.com/documentation/foundation/filemanager/1415645-fileexists):

> Attempting to predicate behavior based on the current state of the file system or a particular file on the file system is not recommended. Doing so can cause odd behavior or race conditions. It’s far better to attempt an operation (such as loading a file or creating a directory), check for errors, and handle those errors gracefully than it is to try to figure out ahead of time whether the operation will succeed.

Just try deleting the item and don't worry about the error. It is safe to assume that it will only throw an error if the directory doesn't exist, which is perfectly fine:

```swift
try? FileManager.default.removeItem(at: destinationURL)
```

Scripts/DowloadPortalItemData.swift
yo1995: I realized that I made wrong assumptions here: plist doesn't allow null values, as all the types that a plist supports aren't optional or NSNULL.

Scripts/DowloadPortalItemData.swift
yo1995: `contains` should have [asymptotic O(1) performance](https://forums.swift.org/t/does-dictionary-keys-contains-have-o-1-time-complexity/21647/14), though it is not clearly documented by Apple. I initially suggested this change to use contains, but I'm OK just use subscript.

philium: Thanks! I actually went and looked at the source, but it wasn't clear from that either. Since `keys` is an instance of a custom type (`Keys`), it makes sense that it could be implemented efficiently for hashable key types.

Scripts/DowloadPortalItemData.swift
CalebRas: This happened when I re-indented the file. Should it not be formatted like this?

philium: How the ternary operator is indented is a matter of preference, but it would be better to revert these changes since they are unrelated to the purpose of the PR.

Scripts/DowloadPortalItemData.swift
```diff
@@ -244,33 +245,43 @@ var downloadedItems = previousDownloadedItems
...
-            try FileManager.default.createDirectory(at: destinationURL, withIntermediateDirectories: false)
-        } catch {
-            print("error: Error creating download directory: \(error.localizedDescription).")
+    
+    // Deletes the directory if it already exists.
+    // This happens when the item is not in the plist and needs to be redownloaded.
+    do {
+        try FileManager.default.removeItem(at: destinationURL)
+        print("note: Deleted directory: \(portalItem.identifier)")
+    }
```
philium, line 262: We don't want to throw errors from the script because it gives an unsightly stack dump:

image

That's why elsewhere we catch any error and explicitly exit with a code of `1`. Also elsewhere, we only print a message for unexpected things (eg. errors). I would say that this print statement is unnecessary and should be removed. And as I said before, if `removeItem(at:)` does throw an error, it is almost certainly because the item (directory) doesn't exist, so we can just move on. No need to try to handle that error, which is why I suggested to just use `try?`.

philium (approved):
Looks good! Thanks for making those changes.

yo1995 (approved):

CalebRas:
Thank you for all your feedback you guys!

Issue #153, opened 2023-04-04
Examine potential retain cycle from unstructured Task blocks


yo1995:
https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/b068fdae82d66b88107a508093130d0a85e92200/Shared/Samples/Sketch%20on%20map/SketchOnMapView.swift#L216-L230

According to [this discussion](https://forums.swift.org/t/for-await-loop-wrapped-in-a-task-block/64159/2), there might be a retain cycle in these 2 Task blocks. Check in debugger and fix if needed.

mhdostal:
@philium The code in question is in a method on the model, but it's currently being run as the result of a button tap. I will move it into a `.task` and give it a whirl.  Thanks.

mhdostal:
The model method called when the user selects an external device (or mocked data) kicks off four tasks, one which will finish and three that will not finish (they're watching @Streamed properties). (I think) because of that, using the `.task` modifier didn't make a difference, I still needed to create `Task.detached` tasks and hold them in an array for cancellation. In other cases where there's only one task, using the `.task` modifier probably would work better.

philium:
A view can have more than one task modifier.

yo1995:
For a group of `for-await-in` infinite loops, I settled with sth like below

```swift
MapView(map: model.map)
    .task {
        do { try await startGeotriggerMonitors(monitors) } catch { // error handling }
    }

/// Starts the geotrigger monitors and handles posted notifications.
/// - Parameter geotriggerMonitors: The geotrigger monitors to start.
private func startGeotriggerMonitors(_ geotriggerMonitors: [GeotriggerMonitor]) async throws {
    await withThrowingTaskGroup(of: Void.self) { group in
        for monitor in geotriggerMonitors {
            group.addTask {
                try await monitor.start()
                for await newNotification in monitor.notifications where newNotification is FenceGeotriggerNotificationInfo {
                    await model.handleGeotriggerNotification(newNotification as! FenceGeotriggerNotificationInfo)
                }
            }
        }
    }
}
```

I think there are a few benefits:

1. task modifier's lifetime = view's lifetime, no need to manually cancel a unstructured task.
2. avoid the hassle of managing the lifecycle of unstructured task in the data model.
3. no need of multiple task modifiers - all the tasks are added to 1 task group. Once the task group is canceled, it will propagate to child tasks.

PR #246, opened 2023-08-11
[Update] `GeometryEditor` sample enhancements

zkline101:
## Description
Changes: 
- Renames the geometry editor sample from `Sketch on map` -> `Create and edit geometries`
- Add shape tool feature to the sample
- Add uniform scale toggle to the sample 

## Linked Issue(s)

- `swift/issues/4382`

## How To Test

The shape tool can be tested by clicking on "Shapes" in the main menu and the uniform scale toggle can be tested in the menu after clicking on a tool.

## Screenshots
Simulator Screenshot - iPhone 14 Pro Max - 2023-08-11 at 15 29 43
Simulator Screenshot - iPhone 14 Pro Max - 2023-08-11 at 15 30 05

PR #129, opened 2023-03-20
[New] Sketch on map

rolson:
## Description

This PR implements `Sketch on map` in `Display information` category.
URL to README [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/9953133126090fb2ce92a932245bf5d1376d982e/Shared/Samples/Sketch%20on%20the%20map/README.md)

## Linked Issue(s)

- `swift/issues/3580`

## How To Test

- Tap the pencil menu button and choose a tool for interactive editing
- Tap the pencil again for interactive editing options including undo/redo/delete/clear/save, etc

## Screenshots

sketch-on-map-1
sketch-on-map-2

PR #372, opened 2024-04-04
[New] Snap geometry edits

yo1995:
## Description

This PR implements `Snap geometry edits` in `Edit and Manage Data` category.

URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Ting/New-Snapping/Shared/Samples/Snap%20geometry%20edits/README.md

## Linked Issue(s)

- `swift/issues/5084`
- Kotlin https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/194

## How To Test

- Open the sample, use the geometry editor to draw all kinds of geometries
- After seeing the normal behavior, turn on snapping for all layers
- Continue drawing geometries
    - On iOS, when you drag a vertex to a feature, it should snap with highlight;
    - On Mac Catalyst, hover the mouse and it snaps
- Turn off some layers, and see they don't snap anymore.

## Screenshots

screenshot

## To Discuss

- We cannot reuse types from other samples, so unfortunately while https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/372/commits/db920126846151efdac5229b8fc6f60aa3cb2dfd is short and concise, I have to copy everything from `Create and edit geometries` to use the geometry editor. 😢 
- All the geometry editor code are copy pasted from the sample with only renaming and doc changes.
- "main", "lateral" are the thick and thin azure polylines; "system valve" are the green and red cat eyes; "service connection" are the "M"s; "curb stop valves" are the "x" shaped circles (need to zoom in to curb scale).
- I didn't include the "Tap to edit graphic" feature from the original design as I think it is not relevant to this sample.


CalebRas:
Reviewing for Mark since he is on vacation.

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.swift
```diff
@@ -0,0 +1,91 @@
...
+    @State private var layersAreLoaded = false
+    
+    /// A Boolean value indicating whether the snap settings are presented.
+    @State private var showsSnapSettings = false
+    
+    var body: some View {
+        MapView(map: map, graphicsOverlays: [model.graphicsOverlay])
+            .geometryEditor(model.geometryEditor)
+            .task {
+                // Load every layer in the webmap when the sample starts.
```
CalebRas, line 54:
```suggestion
                // Load every layer in the web map when the sample starts.
```

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.SnapSettingsView.swift
```diff
@@ -0,0 +1,85 @@
...
+                snappingEnabled = model.geometryEditor.snapSettings.isEnabled
+                
+                // Creates an array from snap source layers with their
+                // layer name and source settings.
+                snapSources = model.geometryEditor.snapSettings.sourceSettings.compactMap { sourceSettings in
+                    if let layer = sourceSettings.source as? FeatureLayer {
+                        return (layer.name, sourceSettings)
+                    } else {
+                        return nil
+                    }
```
CalebRas, line 75:
```suggestion
                    guard let layer = sourceSettings.source as? FeatureLayer else { return nil }
                    return (layer.name, sourceSettings)
```

Not a big deal, but this seems a little cleaner to me.

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.swift
```diff
@@ -0,0 +1,91 @@
...
+        // of data when editing, so full resolution is needed for valid snapping.
+        map.loadSettings.featureTilingMode = .enabledWithFullResolutionWhenSupported
+        return map
+    }()
+    
+    /// The model that is required by the geometry editor menu.
+    @StateObject private var model = GeometryEditorModel(
+        geometryEditor: GeometryEditor(),
+        graphicsOverlay: GraphicsOverlay(renderingMode: .dynamic)
+    )
```
CalebRas, line 39:
```suggestion
    @StateObject private var model = GeometryEditorModel()
```

Is there a reasons to create these objects here? Seems like it would be simpler to create them in the initalizer.

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.swift
```diff
@@ -0,0 +1,91 @@
...
+import SwiftUI
+
+/// A view that shows how to interact with the geometry editor with snapping.
+struct SnapGeometryEditsView: View {
+    /// The map to display in the view.
+    @State private var map: Map = {
+        let map = Map(
+            item: PortalItem(
+                portal: .arcGISOnline(connection: .anonymous),
+                // A stripped down Naperville water distribution network webmap.
```
CalebRas, line 25:
```suggestion
                // A stripped down Naperville water distribution network web map.
```

CalebRas (commented):
Nice sample! Mostly just suggestions. Feel free to veto any of them.

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.swift
yo1995: Copy pasted that from the Create and edit geometry sample. I tried to modify as few things as possible to avoid any confusion when we need to refactor both samples. But this looks like an innocuous change so I'll make it.

<sub>some 👿 soul yelling in the distance</sub> - <sup>it's not [dry](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) ;-)</sup>

yo1995:
Once I fix https://github.com/Esri/arcgis-maps-sdk-swift-samples/issues/153 in the samples, I'll request another review.

pgruenler (commented):
Looks good. I left a few comments.

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.GeometryEditorModel.swift
```diff
@@ -0,0 +1,155 @@
...
+                canClearCurrentSketch = geometry.map { !$0.isEmpty } ?? false
+                canSave = geometry?.sketchIsValid ?? false
+            }
+        }
+        
+        /// A Boolean value indicating if the geometry editor has started.
+        @Published private(set) var isStarted = false
+        
+        /// A Boolean value indicating if the scale mode is uniform.
+        @Published var shouldUniformScale = false {
```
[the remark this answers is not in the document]
yo1995: Was added in 109ede75703feb423d11b211d893bc319125f1a2 in #246 

I'm OK with the change; if I make the change here I'll apply the same change in a follow up PR to fix the retain cycle for the original sample.

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.GeometryEditorModel.swift
```diff
@@ -0,0 +1,155 @@
...
+                for await selection in geometryEditor.$selectedElement {
+                    self?.selection = selection
+                }
+            }
+        }
+        
+        /// Saves the current geometry to the graphics overlay and stops editing.
+        /// - Precondition: `canSave`
+        func save() {
+            precondition(canSave)
```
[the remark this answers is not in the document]
yo1995: Was added here dbabd11fdfa3b191588a84f296bcd914ea66878c see [this](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/129#discussion_r1146548483) comment.

I think it was added to force crash when a programmer error happens, i.e., save is called when the geometry is not valid. I thought just force unwrap the geometry should also be OK, but it may provide less crash info than the precondition?

pgruenler (dismissed):

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.GeometryEditorMenu.swift
```diff
@@ -0,0 +1,155 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+extension SnapGeometryEditsView {
+    /// A view that provides a menu for geometry editor functionality.
+    struct GeometryEditorMenu: View {
```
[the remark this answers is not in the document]
yo1995: It might be not ideal, but we sometimes have to duplicate code across multiple samples instead of reusing them. Here is an excerpt from Jenn Merritt

> My understanding is that any ArcGIS API shouldn't be reused across multiple samples so that they are individual and copy paste-able as you say. I think it's one of those situations where for the sample viewer app it's not necessarily best practice, but it keeps the samples separate if a user was really only interested in one area, rather than looking at the samples/sample viewer app as a whole. Then I think more sample viewer app/platform specific things like data management or other utils are fine to reuse. My 2c!

With that said, I do think geometry editor menu can be a good candidate for a toolkit component. Can you add a comment in `common-toolkit/issues/42` to summarize our findings in the latest implementation?

CalebRas (approved):

Shared/Samples/Snap geometry edits/SnapGeometryEditsView.GeometryEditorMenu.swift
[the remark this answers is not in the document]
yo1995: actually, I'll do this.

pgruenler (approved):

yo1995:
Thanks for the feedback!

PR #385, opened 2024-04-11
[Update] Add `NSLocationUsageDescription` for Mac Catalyst app

yo1995:
## Description

App Store Connect review rejected with the following comment

> If the macOS app requests access to location data, include the [NSLocationUsageDescription](https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationusagedescription?language=occ) key on the info.plist. The [NSLocationWhenInUseUsageDescription](https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationwheninuseusagedescription) key is only for iOS apps.

Include this key to make ASC reviewers happy. 🫤

## To Discuss

From reading this article: https://developer.apple.com/documentation/corelocation/requesting_authorization_to_use_location_services

> On macOS, When in Use and Always authorizations are functionally equivalent. Because macOS apps continue to run in the background after their initial launch, they are always in use. If you create your Mac app using Mac Catalyst, request authorization based on the needs of your iOS app.

It's not clear whether this key is needed or not for Mac Catalyst. Hope this change won't affect our next iOS submission, as the key was deprecated for iOS.

### DIFF
--- a/iOS/Info.plist
+++ b/iOS/Info.plist
@@ -81,6 +81,8 @@
 		<string>com.eos-gnss.positioningsource</string>
 		<string>com.geneq.sxbluegpssource</string>
 	</array>
+	<key>NSLocationUsageDescription</key>
+	<string>Your location is used to show your position on the map on Mac Catalyst.</string>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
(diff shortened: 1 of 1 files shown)


zkline101 (approved):
Reviewing for Mark since he is out.

CalebRas (approved):

yo1995:
It seems the plist sorting order is wrong somehow. Let me fix that real quick 🤔

zkline101 (approved):

CalebRas (approved):

PR #387, opened 2024-04-15
[Fix] Resolve the Task introduced retain cycle

yo1995:
## Description

This PR fixes the retain cycle discussed in #153 .

## Linked Issue(s)

- #153 

## How To Test

use `deinit` to see if the model is deallocated after sample is closed.

```swift
deinit {
    print("✅ cleared")
}
```

### DIFF
--- a/Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
+++ b/Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
@@ -20,25 +20,13 @@ struct CreateAndEditGeometriesView: View {
     /// The map to display in the view.
     @State private var map = Map(basemapStyle: .arcGISTopographic)
     
-    /// The model that is required by the menu.
-    @StateObject var model = GeometryEditorMenuModel()
+    /// The view model for this sample.
+    @StateObject private var model = GeometryEditorModel()
     
     var body: some View {
         VStack {
-            MapView(map: map, graphicsOverlays: [model.graphicsOverlay])
+            MapView(map: map, graphicsOverlays: [model.geometryOverlay])
                 .geometryEditor(model.geometryEditor)
-                .task {
-                    for await geometry in model.geometryEditor.$geometry {
-                        // Update geometry when there is an update.
-                        model.onGeometryChanged(geometry)
-                    }
-                }
-                .task {
-                    for await selection in model.geometryEditor.$selectedElement {
-                        // Update selected element when there is an update.
-                        model.onSelectedElementChanged(selection)
-                    }
-                }
         }
         .toolbar {
             ToolbarItem(placement: .primaryAction) {
@@ -51,7 +39,13 @@ struct CreateAndEditGeometriesView: View {
 /// A view that provides a menu for geometry editor functionality.
 struct GeometryEditorMenu: View {
     /// The model for the menu.
-    @ObservedObject var model: GeometryEditorMenuModel
+    @ObservedObject var model: GeometryEditorModel
+    
+    /// The currently selected element.
+    @State private var selectedElement: GeometryEditorElement?
+    
+    /// The current geometry of the geometry editor.
+    @State private var geometry: Geometry?
     
     var body: some View {
         Menu {
@@ -64,6 +58,18 @@ struct GeometryEditorMenu: View {
             }
         } label: {
             Label("Geometry Editor", systemImage: "pencil.tip.crop.circle")
+        }
+        .task {
+            for await geometry in model.geometryEditor.$geometry {
+                // Update geometry when there is an update.
+                self.geometry = geometry
+            }
+        }
+        .task {
+            for await element in model.geometryEditor.$selectedElement {
+                // Update selected element when there is an update.
+                selectedElement = element
+            }
         }
     }
 }
@@ -177,14 +183,14 @@ extension GeometryEditorMenu {
             } label: {
                 Label("Undo", systemImage: "arrow.uturn.backward")
             }
-            .disabled(!model.canUndo)
+            .disabled(!canUndo)
             
             Button {
                 model.geometryEditor.redo()
             } label: {
                 Label("Redo", systemImage: "arrow.uturn.forward")
             }
-            .disabled(!model.canRedo)
+            .disabled(!canRedo)
             
             Button {
                 model.geometryEditor.deleteSelectedElement()
@@ -200,7 +206,7 @@ extension GeometryEditorMenu {
             } label: {
                 Label("Clear Current Sketch", systemImage: "trash")
             }
-            .disabled(!model.canClearCurrentSketch)
+            .disabled(!canClearCurrentSketch)
             
             Divider()
             
@@ -209,7 +215,7 @@ extension GeometryEditorMenu {
             } label: {
                 Label("Save Sketch", systemImage: "square.and.arrow.down")
             }
-            .disabled(!model.canSave)
+            .disabled(!canSave)
             
             Button {
                 model.stop()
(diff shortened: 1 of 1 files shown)


Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -21,15 +21,24 @@ struct CreateAndEditGeometriesView: View {
...
+    @StateObject var model = GeometryEditorMenuModel()
     
     var body: some View {
         VStack {
             MapView(map: map, graphicsOverlays: [model.graphicsOverlay])
                 .geometryEditor(model.geometryEditor)
+                .task {
+                    for await geometry in model.geometryEditor.$geometry {
+                        // Update geometry when there is an update.
+                        model.onGeometryChanged(geometry)
```
philium, line 33: For both this and `onSelectedElementChanged`, the names are a bit odd. Obviously if a new value is being set, it is because the value changed. Since the change happened in the view, not the model, a more appropriate name would just be `setGeometry(_:)` or `setSelection(_:)`. However, since the implementation of both of those methods simply sets a property on the model, it would make sense just to make those properties settable by the view and do away with the methods that aren't really pulling their own weight.

With all that being said, as far as I can tell, the model doesn't actually even use the selection, which is a pretty good indication that the property should be on a view instead of on the model. And since it is only used by `GeometryEditorMenu`, it seems like the selection should be on that view:

```swift
@State private var selectedElement: GeometryEditorElement?
```

And the associated task modifier should be as well.

It is likely the same should be done for `geometry`; I just haven't dug into it enough to know for sure. While one might argue that the model is "using" it due to it having a property observer, whatever is being done in that property observer could also be done in the associated view. In my opinion, that's not a strong enough argument to put the property in the model. Remember, observable properties on a model equal less efficient view updates. The view model should have only what is necessary.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -21,15 +21,24 @@ struct CreateAndEditGeometriesView: View {
...
+                    }
+                }
+                .task {
+                    for await selection in model.geometryEditor.$selectedElement {
```
philium, line 37: A better name for the loop variable would be `selectedElement` or perhaps even just `element`.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -226,10 +235,10 @@ extension GeometryEditorMenu {
 @MainActor
 class GeometryEditorMenuModel: ObservableObject {
     /// The geometry editor.
-    let geometryEditor: GeometryEditor
+    let geometryEditor = GeometryEditor()
     
     /// The graphics overlay used to save geometries to.
-    let graphicsOverlay: GraphicsOverlay
+    let graphicsOverlay = GraphicsOverlay(renderingMode: .dynamic)
```
philium, line 241: A better name for this might be `geometryOverlay`. You could go with `geometriesOverlay`, but it is well understood that the adjectival noun (geometry) could be referring to more than one geometry. For example, we say "car mechanic", not "cars mechanic", and we understand that the mechanic may work (and most likely works) on more than one car.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
yo1995: This is good suggestion. I'll apply and refactor the sample a bit.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -21,15 +21,24 @@ struct CreateAndEditGeometriesView: View {
     @State private var map = Map(basemapStyle: .arcGISTopographic)
     
     /// The model that is required by the menu.
-    @StateObject var model = GeometryEditorMenuModel(
-        geometryEditor: GeometryEditor(),
-        graphicsOverlay: GraphicsOverlay(renderingMode: .dynamic)
-    )
+    @StateObject var model = GeometryEditorMenuModel()
```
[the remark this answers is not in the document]
yo1995: I have renamed it to remove the word "menu" to reflect the fact that the model is a general view model for the whole sample.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -56,10 +59,22 @@ struct GeometryEditorMenu: View {
         } label: {
             Label("Geometry Editor", systemImage: "pencil.tip.crop.circle")
         }
+        .task {
```
yo1995, line 62: Should these task modifiers be attached to the `editMenuContent`, or should they remains attached to the `Menu`? The former seems more correct semantically - you only needs updates when edit a geometry; but the latter seems more convenient - don't need to worry about the conditional view's lifecycle vs when the task modifiers are created, etc. 🤔

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -211,56 +226,51 @@ extension GeometryEditorMenu {
...
-extension GeometryEditorMenu {
+private extension GeometryEditorMenu {
     /// A Boolean value indicating whether the selection can be deleted.
     ///
     /// In some instances deleting the selection may be invalid. One example would be the mid vertex
     /// of a line.
-    var deleteButtonIsDisabled: Bool {
-        guard let selection = model.selection else { return true }
-        return !selection.canBeDeleted
+    private var deleteButtonIsDisabled: Bool {
```
philium, line 234: Since the extension itself is private, annotating the members as private is superfluous.

```suggestion
    var deleteButtonIsDisabled: Bool {
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -211,56 +226,51 @@ extension GeometryEditorMenu {
...
-class GeometryEditorMenuModel: ObservableObject {
-    /// The geometry editor.
-    let geometryEditor: GeometryEditor
-    
-    /// The graphics overlay used to save geometries to.
-    let graphicsOverlay: GraphicsOverlay
     
     /// A Boolean value indicating if the geometry editor can perform an undo.
-    @Published private(set) var canUndo = false
+    private var canUndo: Bool {
```
philium, line 240:
```suggestion
    var canUndo: Bool {
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -211,56 +226,51 @@ extension GeometryEditorMenu {
...
+        return model.geometryEditor.canUndo
+    }
     
     /// A Boolean value indicating if the geometry editor can perform a redo.
-    @Published private(set) var canRedo = false
-    
-    /// The currently selected element.
-    @Published private(set) var selection: GeometryEditorElement?
+    private var canRedo: Bool {
```
philium, line 245:
```suggestion
    var canRedo: Bool {
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -211,56 +226,51 @@ extension GeometryEditorMenu {
...
+        return model.geometryEditor.canRedo
+    }
     
     /// A Boolean value indicating if the geometry can be saved to a graphics overlay.
-    @Published private(set) var canSave = false
+    private var canSave: Bool {
```
philium, line 250:
```suggestion
    var canSave: Bool {
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -211,56 +226,51 @@ extension GeometryEditorMenu {
...
+        return geometry?.sketchIsValid ?? false
+    }
     
     /// A Boolean value indicating if the geometry can be cleared from the geometry editor.
-    @Published private(set) var canClearCurrentSketch = false
+    private var canClearCurrentSketch: Bool {
```
philium, line 255:
```suggestion
    var canClearCurrentSketch: Bool {
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -274,39 +284,20 @@ class GeometryEditorMenuModel: ObservableObject {
...
-        Task { [weak self, geometryEditor] in
-            for await selection in geometryEditor.$selectedElement {
-                self?.selection = selection
-            }
-        }
-    }
-    
     /// Saves the current geometry to the graphics overlay and stops editing.
-    /// - Precondition: `canSave`
+    /// - Precondition: Whether or not the geometry is from a valid sketch.
```
philium, line 288: This precondition description doesn't make sense. It should describe what the condition is, ie. what would cause a failure.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -211,56 +226,51 @@ extension GeometryEditorMenu {
...
-            canSave = geometry?.sketchIsValid ?? false
-        }
-    }
-    
     /// A Boolean value indicating if the geometry editor has started.
-    @Published var isStarted = false
+    @Published private(set) var isStarted = false
     
     /// A Boolean value indicating if the scale mode is uniform.
     @Published var shouldUniformScale = false {
```
dfeinzimer, line 276: Something seems off with this name.. maybe `isUniformScale`?

yo1995: Phil G had the same question [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/372#discussion_r1561367356)… I think it warrants a change. 😥

philium (approved):

yo1995:
Thank you! I'll retrofit the other sample with some comments in this PR.

PR #401, opened 2024-05-01
[New] Edit with branch versioning

CalebRas:
## Description

This PR implements `Edit with branch versioning` in `Edit and Manage Data` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-EditWithBranchVersioning/Shared/Samples/Edit%20with%20branch%20versioning)

## Linked Issue(s)

- `swift/issues/5287`

## How To Test

- Tap "Create", enter a version name, and selected an access to create a version.
- Tap "Switch" to switch to the new version.
- Tap on the feature on feature to select it.
    - Tap on the edit button to update feature's damage type.
    - Or, tap on map to move the feature.

## Screenshots
edit-with-branch-versioning-1 edit-with-branch-versioning-2


Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.Views.swift
```diff
@@ -0,0 +1,152 @@
...
+    /// All the version access cases.
+    static var allCases: [Self] { [.public, .protected, .private] }
+    
+    /// The text description of the version access.
+    var description: String {
+        switch self {
+        case .public: return "Any portal user can view and edit the version."
+        case .protected: return "Any portal user can view the version, but only the version owner, feature layer owner, and portal administrator can edit it."
+        case .private: return "Only the version owner, feature layer owner, and portal administrator can view and edit the version."
+        @unknown default: return "Unknown version access."
```
philium, line 149: Return keywords aren't needed:

```suggestion
        case .public: "Any portal user can view and edit the version."
        case .protected: "Any portal user can view the version, but only the version owner, feature layer owner, and portal administrator can edit it."
        case .private: "Only the version owner, feature layer owner, and portal administrator can view and edit the version."
        @unknown default: "Unknown version access."
```

philium (commented):
I haven't finished my review (I haven't gotten to `EditWithBranchVersioningView.swift`), but here is my feedback so far.

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,250 @@
...
+                            selectedAction = .updateFeature
+                        }
+                    },
+                    message: { _ in
+                        Text("Do you want to move the selected feature?")
+                    }
+                )
+                .task(id: selectedAction) {
+                    guard let selectedAction else { return }
+                    defer { self.selectedAction = nil }
```
philium, line 94: It is not entirely clear why the selected action is being nilled out when this scope is exited. You could add a comment, but consider not using a defer block. It doesn't really fit the mental model, particularly since this is an async function and the defer block has no way of knowing what the current state is when it gets executed.

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,250 @@
...
+                    calloutPlacement = nil
+                    
+                    do {
+                        switch selectedAction {
+                        case .setUp:
+                            try await model.setUp()
+
```
philium, line 101: We usually don't put a blank line in between cases. Is there a particular reason why you did it here? if not, please remove them.

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,250 @@
...
+                        case .selectFeature(let screenPoint, let mapPoint):
+                            model.clearSelection()
+                            
+                            // Identifies the feature on the feature layer at the tapped point.
+                            let result = try await mapViewProxy.identify(
+                                on: model.featureLayer,
+                                screenPoint: screenPoint,
+                                tolerance: 10
+                            )
+                            guard let feature = result.geoElements.first as? Feature else { return }
```
philium, line 120: A guard statement nested this deep really breaks the mental model for guard statements. Consider using an if statement instead.

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,250 @@
...
+    private var editDamageTypeButton: some View {
+        Button("Edit", systemImage: "pencil") {
+            isSettingDamageType = true
+        }
+        .imageScale(.large)
+        .labelStyle(.iconOnly)
+        .disabled(model.onDefaultVersion)
+        .confirmationDialog("Damage Type", isPresented: $isSettingDamageType, titleVisibility: .visible) {
+            ForEach(DamageType.allCases, id: \.self) { damageType in
+                Button(damageType.rawValue) {
```
philium, line 163: Similar to `VersionAccess`, it would be better not to use the case name (non-localized) as a user-facing (localized) string as it isn't best practice for a production app.

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,250 @@
...
+                }
+            }
+        } message: {
+            Text("Choose a version to switch to.")
+        }
+    }
+}
+
+/// An asynchronous action associated with the sample.
+private enum AsyncAction: Equatable {
```
philium, line 211: This is very similar to `Model.State`. Is it possible to unify the two?

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,250 @@
...
+
+/// The damage type of a feature.
+private enum DamageType: String, CaseIterable {
+    case destroyed = "Destroyed"
+    case major = "Major"
+    case minor = "Minor"
+    case affected = "Affected"
+    case inaccessible = "Inaccessible"
+    case `default` = "Default"
+}
```
philium, line 236: I mentioned this earlier, but an enum's raw value isn't intended for user-facing strings. Add a `title` property and switch over `self` to determine which string to use as the title.

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,258 @@
...
+    case makeVersion
+    
+    /// Switches to a version with an associated version name.
+    case switchToVersion(version: String)
+    
+    /// Select a feature identified from an associated screen and map point.
+    case selectFeature(screenPoint: CGPoint, mapPoint: Point)
+    
+    /// Updates the selected feature in it's feature table.
+    case updateFeature
```
philium, line 226: We don't usually separate enum cases with a blank line.

```suggestion
    /// Sets up the sample.
    case setUp
    /// Makes a version with the current version parameters.
    case makeVersion
    /// Switches to a version with an associated version name.
    case switchToVersion(version: String)
    /// Select a feature identified from an associated screen and map point.
    case selectFeature(screenPoint: CGPoint, mapPoint: Point)
    /// Updates the selected feature in it's feature table.
    case updateFeature
```

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+}
+
+private extension String {
+    /// The key for a feature's damage type attribute.
+    static let damageTypeKey = "typdamage"
+    
+    /// The key for a feature's place name attribute.
+    static let placeNameKey = "placename"
```
philium, line 251: We don't usually separate stored properties with a blank line.

```suggestion
    /// The key for a feature's damage type attribute.
    static let damageTypeKey = "typdamage"
    /// The key for a feature's place name attribute.
    static let placeNameKey = "placename"
```

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.Views.swift
```diff
@@ -0,0 +1,104 @@
...
+                        Button("Add") {
+                            let parameters = ServiceVersionParameters()
+                            parameters.access = selectedVersionAccess
+                            parameters.name = versionName
+                            parameters.description = versionDescription
+                            
+                            action(parameters)
+                            dismiss()
+                        }
+                        .disabled(versionName.isEmpty)
```
philium, line 83: If the name can include spaces, which it looks like it can, then the button should also be disabled when the last character is a space. That way, the user can still enter a space (followed by additional non-whitespace characters), but not be able to add while the name ends in a space.

CalebRas: It looks like beginning and trailing spaces are actually trimmed when the version is made, and the errors I was getting testing those were due to the version already existing. (Not sure why the readme specified that it couldn't begin with a space, though.)

philium: In that case, you'll want to trim leading/trailing whitespace before verifying the character count. Theoretically, someone could be prevented from submitting a valid name because they have accidentally included leading/trailing whitespace that makes the name go over the character limit.

CalebRas: @philium I am not sure how I would do this without trimming the whitespace, which would then prevent using spaces at all. Any suggestions?

philium: You would do it here:

```swift
.disabled(versionName.isEmpty || versionName.last!.isWhitespace)
```

philium (approved):

Shared/Samples/Edit with branch versioning/EditWithBranchVersioningView.swift
```diff
@@ -0,0 +1,253 @@
...
+                }
+            }
+        } message: {
+            Text("Choose a version to switch to.")
+        }
+    }
+}
+
+/// An asynchronous action associated with the sample.
+private enum AsyncAction: Equatable {
```
zkline101, line 212: I thought this was a pretty neat idea

zkline101 (approved):
Looks good

zkline101:
I noticed that this sample can get into a state where it looks like I cannot unselect the callout no matter where I tap. Is there another way to cancel it?



Simulator Screen Recording - iPhone 15 Pro Max - 2024-05-10 at 16 05 43

CalebRas:
> I noticed that this sample can get into a state where it looks like I cannot unselect the callout no matter where I tap. Is there another way to cancel it?

Tapping "Cancel" also dismisses the callout now. Thanks for catching this!

zkline101:
Should the feature be unselected as well?

zkline101 (dismissed):

zkline101 (approved):

philium (approved):

zkline101 (approved):

CalebRas:
Thank you for reviewing this!

PR #411, opened 2024-05-16
[New] Filter by definition expression or display filter

CalebRas:
## Description

This PR implements `Filter by definition expression or display filter` in `Search and Query` category.
URL to README: [README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-FilterByDefinitionExpressionOrDisplayFilter/Shared/Samples/Filter%20by%20definition%20expression%20or%20display%20filter)

## Linked Issue(s)

- `swift/issues/5369`

## How To Test

- Tap "Expression" to filter the feature layer with the definition expression.
- Tap "Display Filter" to filter the results with the display filter definition.
- Tap "None" to view feature layer without filtering.
- See the feature count change when you pan, zoom, and update the filtering selection.

## Screenshots

filter-by-definition-expression-or-display-filter


Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+                        guard let viewExtent = mapViewProxy.envelope(
+                            fromViewRect: viewRect
+                        ) else { return }
+                        
+                        do {
+                            // Gets the feature count contained within the envelope.
+                            featureCount = try await model.queryFeatureCount(extent: viewExtent)
+                        } catch {
+                            self.error = error
+                        }
```
philium, line 54: The feature count isn't updated when there is an error. In cases like this, it is often better to combine the success value and the failure error using `Result`:

```swift
@State private var queryResult: Result<Int, Error>?
```

That is more stateful as now there is an initial state (`nil`), a success state, and a failure state, and they are all mutually exclusive. As for the error, instead of presenting an alert, consider displaying it in the status bar at the top of the map.

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+                    .onDrawStatusChanged { drawStatus = $0 }
+                    .task(id: drawStatus) {
+                        // Updates the feature count when the map view finishes drawing.
+                        guard drawStatus == .completed else { return }
+                        
+                        // Creates an envelope from the frame of the map view.
+                        let viewRect = geometryProxy.frame(in: .local)
+                        guard let viewExtent = mapViewProxy.envelope(
+                            fromViewRect: viewRect
+                        ) else { return }
```
philium, line 47: I found this somewhat difficult to read. Consider setting the guard statement apparent with a blank line and not breaking the call to `envelope(fromViewRest:)`:

```suggestion
                        let viewRect = geometryProxy.frame(in: .local)
                        
                        guard let viewExtent = mapViewProxy.envelope(fromViewRect: viewRect) else {
                            return
                        }
```

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+                            featureCount = try await model.queryFeatureCount(extent: viewExtent)
+                        } catch {
+                            self.error = error
+                        }
+                    }
+                    .errorAlert(presentingError: $error)
+            }
+        }
+        .overlay(alignment: .top) {
+            Text("Feature count: \(featureCount)")
```
philium, line 60: Consider something a bit simpler:

```suggestion
            Text("\(featureCount) feature(s)")
```

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+            )
+            return map
+        }()
+        
+        /// A feature layer of San Fransisco 311 incidents.
+        private let featureLayer: FeatureLayer = {
+            // Creates the feature layer from a feature service URL.
+            let featureTable = ServiceFeatureTable(url: .sanFransiscoIncidentsFeatureLayer)
+            return FeatureLayer(featureTable: featureTable)
+        }()
```
philium, line 103:
```suggestion
        private let featureLayer = FeatureLayer(
            // Creates the feature layer from a feature service URL.
            featureTable: ServiceFeatureTable(url: .sanFransiscoIncidentsFeatureLayer)
        )
```

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+        
+        /// The parameters for querying the feature layer.
+        private let queryParameters = QueryParameters()
```
philium, line 106: There isn't really a need to re-use the query parameters. Personally, I would just create them where needed as needed.

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+                name: "Trees",
+                whereClause: "req_type LIKE 'Tree Maintenance or Damage'"
+            )
+            
+            // Creates a display filter definition from the display filter.
+            let treesDisplayFilterDefinition = ManualDisplayFilterDefinition(
+                activeFilter: treesDisplayFilter,
+                availableFilters: [treesDisplayFilter]
+            )
+            return treesDisplayFilterDefinition
```
philium, line 124:
```suggestion
            return ManualDisplayFilterDefinition(
                activeFilter: treesDisplayFilter,
                availableFilters: [treesDisplayFilter]
            )
```

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+        }()
+        
+        init() {
+            map.addOperationalLayer(featureLayer)
+        }
+        
+        /// Filters the feature layer based on a given filter mode.
+        /// - Parameter filterMode: The mode indicating how to filter the layer.
+        func filterFeatureLayer(filterMode: FilterMode) {
```
philium, line 133: `filterMode` -> `mode` (since _filter_ is included in the name)

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+            // Sets the feature layer's definition expression.
+            featureLayer.definitionExpression = filterMode == .definitionExpression
+            ? treesDefinitionExpression
+            : ""
+            
+            // Sets the feature layer's display filter definition.
+            featureLayer.displayFilterDefinition = filterMode == .displayFilterDefinition
+            ? treesDisplayFilterDefinition
+            : nil
```
philium, line 142: Consider using a switch statement, making this code much more readable:

```suggestion
            let (definitionExpression, displayFilterDefinition): (String, DisplayFilterDefinition?) = switch filterMode {
            case .definitionExpression:
                (treesDefinitionExpression, nil)
            case .displayFilterDefinition:
                ("", treesDisplayFilterDefinition)
            case .none:
                ("", nil)
            }
            // Sets the feature layer's definition expression.
            featureLayer.definitionExpression = definitionExpression
            // Sets the feature layer's display filter definition.
            featureLayer.displayFilterDefinition = displayFilterDefinition
```

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+        }
+        
+        /// Queries the count of features on the feature layer within a given extent.
+        /// - Parameter extent: The extent to query.
+        /// - Returns: The number of features within the extent.
+        func queryFeatureCount(extent: Envelope) async throws -> Int {
```
philium, line 148: A better name for this method might be something like `numberOfFeatures(withinExtent:)`.

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+            guard let featureTable = featureLayer.featureTable else { return 0 }
```
philium, line 149: We shouldn't ever attempt to fail gracefully when we know that failure isn't an option. In this case, we know that the layer will have a table (since we created it ourselves), so just force unwrap.

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,184 @@
...
+            
+            queryParameters.geometry = extent
+            let featureCount = try await featureTable.queryFeatureCount(using: queryParameters)
+            
+            return featureCount
```
philium, line 154:
```suggestion
            return try await featureTable.queryFeatureCount(using: queryParameters)
```

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
CalebRas: This also seems a little convoluted to me. Would it make sense to have two switch statements here? (One for each constant)

philium: No, there should definitely only be one switch statement. You could move the setting of the feature layer properties into the cases, which I originally started doing, but this way is less error prone (a case can't forget to set one of the properties).

Shared/Samples/Filter by definition expression or display filter/FilterByDefinitionExpressionOrDisplayFilterView.swift
```diff
@@ -0,0 +1,180 @@
...
+                        let viewRect = geometryProxy.frame(in: .local)
+                        
+                        guard let viewExtent = mapViewProxy.envelope(fromViewRect: viewRect) else {
+                            return
+                        }
+                        
+                        featureCountResult = await Result {
+                            // Gets the feature count contained within the envelope.
+                            try await model.numberOfFeatures(withinExtent: viewExtent)
+                        }
```
philium, line 50: `Result.init(awaiting:)` is provided by the toolkit and, in my opinion, [it really shouldn't be](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/issues/732). Consider not using it.

```suggestion
                        do {
                            // Gets the feature count contained within the envelope.
                            let count = try await model.numberOfFeatures(withinExtent: viewExtent)
                            featureCountResult = .success(count)
                        } catch {
                            featureCountResult = .failure(error)
                        }
```

philium (approved):

philium (approved):

njarecha (approved):
Looks good

CalebRas:
Thank you for your feedback you guys!

PR #419, opened 2024-05-24
[Fix] Address build error and warnings in ConfigureBasemapStyleParametersView.swift

philium:
Related to recent API changes. They weren't intended to introduce build errors, so these changes and the API changes may need to be reverted before the next release. For now, addressing them was the easiest approach.

### DIFF
--- a/Shared/Samples/Configure basemap style parameters/ConfigureBasemapStyleParametersView.swift
+++ b/Shared/Samples/Configure basemap style parameters/ConfigureBasemapStyleParametersView.swift
@@ -20,7 +20,7 @@ struct ConfigureBasemapStyleParametersView: View {
     @StateObject private var model = Model()
     
     /// The selected basemap style language strategy.
-    @State private var selectedLanguage: BasemapStyleLanguage = .global
+    @State private var selectedLanguage: BasemapStyleLanguage = .strategic(.global)
     
     /// The selected locale.
     @State private var selectedLocale: Locale = .current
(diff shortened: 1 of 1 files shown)


zkline101 (approved):

yo1995 (commented):
Overall looks good

Shared/Samples/Configure basemap style parameters/ConfigureBasemapStyleParametersView.swift
```diff
@@ -20,70 +20,56 @@ struct ConfigureBasemapStyleParametersView: View {
...
-                        // Use the default language setting for the basemap style.
-                        BasemapStyleLanguage.default,
-                        // Use the global language (English) for basemap labels.
-                        BasemapStyleLanguage.global,
-                        // Uses country-local language for basemap labels.
-                        BasemapStyleLanguage.local,
-                        // Use the system locale language for basemap labels.
-                        BasemapStyleLanguage.applicationLocale
-                    ],
+                    [BasemapStyleLanguageStrategy.default, .global, .local, .applicationLocale],
```
yo1995, line 40: Can you retain the comments so the audience can see the difference between the strategies more easily? Thank you.

philium: Aren't the comments simply repeating what is in the documentation?

yo1995: Yes, I thought adding some comment would help the reader to spot the key place to look at in the code. OK not to add.

yo1995 (approved):

PR #427, opened 2024-06-10
[New] Show viewshed from point on map

chriswebb09:
## Description

This PR implements **`Show viewshed from point on map`** in `Analysis` category.

`swift/issues/5493`

## How To Test

Load map, tap point on map, wait until drawing state is finished. There should be red dot on the map where you tapped as well as the viewshed graphic overlay. 

## Screenshots
show-viewshed-from-point-on-map


Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+        return map
+    }()
+    
+    var body: some View {
+        MapView(map: map, graphicsOverlays: [inputGraphicsOverlay, resultGraphicsOverlay])
+            .onSingleTapGesture { _, tapPoint in
+                self.tapScreenPoint = tapPoint
+            }
+            .overlay(alignment: .center) {
+                // Sets locating indication when currentDrawStatus is true.
```
chriswebb09, line 76: This comment needs to be updated.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+        switch result {
+        case .success(let output):
+            if let resultFeatures = output.outputs["Viewshed_Result"] as? GeoprocessingFeatures {
+                self.processFeatures(resultFeatures: resultFeatures)
+            }
+        case .failure(let error):
+            // Sets error to be displayed.
+            self.error = error
+        case .none:
+            // This case should never execute.
```
chriswebb09, line 155: None case is appearing because geoprocessingJob is now optional vs force unwrapped. The reasoning is this case should not execute and if it does execute it shouldn't crash all the samples. Let me know if this is wrong.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+                // Sets current draw status to false after process tap point operation
+                // completes.
+                defer { currentDrawStatus = false }
+                await process(at: tapScreenPoint)
+            }
+            .errorAlert(presentingError: $error)
+    }
+    
+    private func process(at tapScreenPoint: Point) async {
+        self.addGraphic(at: tapScreenPoint)
```
chriswebb09, line 97: potentially get rid of the self. references so that code is standardized.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+        let geoprocess = GeoprocessingTask(url: .viewshedURL)
+        return geoprocess
+    }()
+    /// Handles the execution of the geoprocessing task and gets the result.
+    @State private var geoprocessingJob: GeoprocessingJob?
+    /// Sets map's initial viewpoint to Vanoise National Park in France.
+    @State private var map: Map = {
+        let map = Map(basemapStyle: .arcGISTopographicBase)
+        map.initialViewpoint = Viewpoint(
+            center: Point(latitude: 45.3790902612337, longitude: 6.84905317262762),
```
yo1995, line 64: I have to use this GIS cliché meme here again… https://xkcd.com/2170/ 😉 

```suggestion
            center: Point(latitude: 45.379, longitude: 6.849),
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+import ArcGIS
+import SwiftUI
+
+struct ShowViewshedFromPointOnMapView: View {
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    /// The point on the screen where the user tapped.
+    @State private var tapScreenPoint: Point?
+    /// The graphics overlay that shows where the user tapped on the screen.
+    @State private var inputGraphicsOverlay = {
```
yo1995, line 24: Currently, there are 5 ArcGIS types which don't emit state changes that got held as a state property - inputGraphicsOverlay, resultGraphicsOverlay, geoprocessingTask, geoprocessingJob, and map.

While it is OK to hold them as state properties, my personal preference would be - if there are more than 3 types that need to be held, let's create a data model for them. i.e.,

```swift
import ArcGIS
import SwiftUI

struct ShowViewshedFromPointOnMapView: View {
    /// The model used to store the geo model and other expensive objects
    /// used in this view.
    private class Model: ObservableObject {
        /// A map with topographic basemap.
        let map = Map(basemapStyle: .arcGISTopographicBase)

        // …
    }
    
    /// The data model for the sample.
    
    var body: some View {
        MapView(map: model.map)
        // …
    }
}
```

Data manipulating methods can also be moved to the data model.

If the data model is even longer, feel free to put it in an extension to the view, like in other samples.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct ShowViewshedFromPointOnMapView: View {
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    /// The point on the screen where the user tapped.
+    @State private var tapScreenPoint: Point?
```
yo1995, line 22: Just a styling choice - would you mind adding a newline between each property? Thanks.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+            outline: nil
+        )
+        let resultRenderer = SimpleRenderer(symbol: fillSymbol)
+        resultGraphicsOverlay.renderer = resultRenderer
+        return resultGraphicsOverlay
+    }()
+    /// The current draw status of the map.
+    @State private var currentDrawStatus: Bool = false
+    /// The processing task that references the online resource through the url.
+    @State private var geoprocessingTask: GeoprocessingTask = {
```
yo1995, line 54:
```suggestion
    @State private var geoprocessingTask = GeoprocessingTask(url: .viewshedURL)
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+                    ProgressView("Drawing…")
+                        .padding()
+                        .background(.ultraThinMaterial)
+                        .cornerRadius(10)
+                        .shadow(radius: 50)
+                }
+            }
+            .task(id: tapScreenPoint) {
+                guard let tapScreenPoint else { return }
+                currentDrawStatus = true
```
yo1995, line 87: This Boolean can have a better name. It could be something like `geoprocessingInProgress`

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
yo1995: Add graphic and calculate viewshed can be called directly in the task modifier, so this helper method can be removed.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,183 @@
...
+            .errorAlert(presentingError: $error)
+    }
+    
+    private func process(at tapScreenPoint: Point) async {
+        self.addGraphic(at: tapScreenPoint)
+        await self.calculateViewshed(at: tapScreenPoint)
+    }
+    
+    /// Removes previously tapped location from overlay and draws new dot on tap location.
+    private func addGraphic(at tapPoint: Point) {
```
yo1995, line 102: To add the docstring template, use Editor > Structure > Add Documentation, or use its shortcut.

We add docstring for each method, unless it is trivial and self-explanatory (which in most cases, can be refactored away)

yo1995 (commented):
Some initial feedback

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: @yo1995 let me know if this is what you were expecting.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+            }
+        }
+    }
+}
+
+private extension ShowViewshedFromPointOnMapView {
+    private class Model: ObservableObject {
+        /// A map with topographic basemap.
+        /// Sets map's initial viewpoint to Vanoise National Park in France.
+        var map: Map = {
```
yo1995, line 147:
```suggestion
        let map: Map = {
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+            let map = Map(basemapStyle: .arcGISTopographicBase)
```
yo1995, line 148:
```suggestion
            let map = Map(basemapStyle: .arcGISTopographic)
```
The base style doesn't have the elevation hillshade.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+                let graphic = Graphic(geometry: feature.geometry)
+                // Sets the graphic on the overlay to display to the user.
+                model.resultGraphicsOverlay.addGraphic(graphic)
+            }
+        }
+    }
+}
+
+private extension ShowViewshedFromPointOnMapView {
+    private class Model: ObservableObject {
```
yo1995, line 144:
```suggestion
   class Model: ObservableObject {
```
As the extension is already private, no need to repeat access level.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+/// Description
```
yo1995, line 18:
```suggestion
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+struct ShowViewshedFromPointOnMapView: View {
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    
+    /// The point on the screen where the user tapped.
+    @State private var tapScreenPoint: Point?
+    
+    /// The current geoprocessing status..
```
yo1995, line 26:
```suggestion
    /// The current geoprocessing status.
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+    @State private var geoprocessingInProgress: Bool = false
```
yo1995, line 27:
```suggestion
    @State private var geoprocessingInProgress = false
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+    @StateObject private var model = Model()
+    
+    var body: some View {
+        MapView(map: model.map, graphicsOverlays: [model.inputGraphicsOverlay, model.resultGraphicsOverlay])
+            .onSingleTapGesture { _, tapPoint in
+                self.tapScreenPoint = tapPoint
+            }
+            .overlay(alignment: .center) {
+                // Sets indication when geoprocessingInProgress is true.
+                if geoprocessingInProgress == true {
```
yo1995, line 39:
```suggestion
                if geoprocessingInProgress {
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+    /// The current geoprocessing status..
+    @State private var geoprocessingInProgress: Bool = false
+    
+    /// The data model for the sample.
+    @StateObject private var model = Model()
+    
+    var body: some View {
+        MapView(map: model.map, graphicsOverlays: [model.inputGraphicsOverlay, model.resultGraphicsOverlay])
+            .onSingleTapGesture { _, tapPoint in
+                self.tapScreenPoint = tapPoint
```
yo1995, line 35:
```suggestion
                tapScreenPoint = tapPoint
```
Also it would be better to rename `tapScreenPoint` to `tapLocation` or `mapPoint`,  to distinguish the concept of a "screen point" (a point in the screen's coordinates) vs a "map point" (a point in the map's coordinate system)

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+                }
+            }
+            .task(id: tapScreenPoint) {
+                guard let tapScreenPoint else { return }
+                geoprocessingInProgress = true
+                // Sets current geoprocessingInProgress to false after process geoprocessing
+                // completes.
+                defer { geoprocessingInProgress = false }
+                addGraphic(at: tapScreenPoint)
+                await calculateViewshed(at: tapScreenPoint)
```
yo1995, line 54:
```suggestion
                // Sets current geoprocessingInProgress to false after process geoprocessing
                // completes.
                defer { geoprocessingInProgress = false }
                addGraphic(at: tapScreenPoint)
                geoprocessingInProgress = true
                await calculateViewshed(at: tapScreenPoint)
```
Not a big deal, but it is better to set the Boolean right prior to where the async method is called.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+            }
+            .errorAlert(presentingError: $error)
+    }
+    
+    /// Removes previously tapped location from overlay and draws new dot on tap location.
+    /// - Parameter tapPoint: Location that the user tapped on the map.
+
```
yo1995, line 61:
```suggestion
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+            }
+            .errorAlert(presentingError: $error)
+    }
+    
+    /// Removes previously tapped location from overlay and draws new dot on tap location.
+    /// - Parameter tapPoint: Location that the user tapped on the map.
+    
+    private func addGraphic(at tapPoint: Point) {
+        model.inputGraphicsOverlay.removeAllGraphics()
+        let graphic = Graphic(geometry: tapPoint)
```
yo1995, line 64:
```suggestion
    private func addGraphic(at point: Point) {
        model.inputGraphicsOverlay.removeAllGraphics()
        let graphic = Graphic(geometry: point)
```
Also, you can move the methods to the model, so the method has less "side effect".

```swift
model.addGraphic(at: tapPoint)

// In the model…

        /// Removes previously tapped location from overlay and draws new dot at tap location.
        /// - Parameter point: Location that the new dot to be drawn.
        func addGraphic(at point: Point) {
            inputGraphicsOverlay.removeAllGraphics()
            let graphic = Graphic(geometry: point)
            inputGraphicsOverlay.addGraphic(graphic)
        }
```

yo1995 (commented):
Some more feedback. All the methods that deal with the objects held by the data model can be moved into the model as well.

chriswebb09:
show-viewshed-from-point-on-map
updated screenshot

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+            let renderer = SimpleRenderer(symbol: pointSymbol)
+            inputGraphicsOverlay.renderer = renderer
+            return inputGraphicsOverlay
+        }()
+        
+        /// Handles the execution of the geoprocessing task and gets the result.
+        var geoprocessingJob: GeoprocessingJob?
+        
+        /// Removes previously tapped location from overlay and draws new dot on tap location.
+        /// - Parameter tapPoint: Location that the user tapped on the map.
```
chriswebb09, line 180: Need to update comment to reflect new param name

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+            // Red dot marker that is added to graphic overlay on map on the location where the user taps.
+            let pointSymbol = SimpleMarkerSymbol(style: .circle, color: .red, size: 10)
+            // Sets the renderer to draw the dot symbol on the graphics overlay.
+            let renderer = SimpleRenderer(symbol: pointSymbol)
+            inputGraphicsOverlay.renderer = renderer
+            return inputGraphicsOverlay
+        }()
+        
+        /// Handles the execution of the geoprocessing task and gets the result.
+        var geoprocessingJob: GeoprocessingJob?
```
yo1995, line 106:
```suggestion
        private var geoprocessingJob: GeoprocessingJob?
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+                center: Point(latitude: 45.379, longitude: 6.849),
+                scale: 144447
+            )
+            return map
+        }()
+        
+        /// The processing task that references the online resource through the url.
+        let geoprocessingTask = GeoprocessingTask(url: .viewshedURL)
+        
+        ///  The graphics overlay that displays the viewshed for the tapped location.
```
yo1995, line 75:
```suggestion
        /// The graphics overlay that displays the viewshed for the tapped location.
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+private extension ShowViewshedFromPointOnMapView {
+    class Model: ObservableObject {
+        /// The error shown in the error alert.
+        @Published var error: Error?
+        
+        /// A map with topographic basemap.
+        /// Sets map's initial viewpoint to Vanoise National Park in France.
+        let map: Map = {
+            let map = Map(basemapStyle: .arcGISTopographic)
+            map.initialViewpoint = Viewpoint(
```
yo1995, line 65:
```suggestion
        /// A map with topographic basemap.
        let map: Map = {
            let map = Map(basemapStyle: .arcGISTopographic)
            // Sets map's initial viewpoint to Vanoise National Park in France.
            map.initialViewpoint = Viewpoint(
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+            let renderer = SimpleRenderer(symbol: pointSymbol)
+            inputGraphicsOverlay.renderer = renderer
+            return inputGraphicsOverlay
+        }()
+        
+        /// Handles the execution of the geoprocessing task and gets the result.
+        var geoprocessingJob: GeoprocessingJob?
+        
+        /// Removes previously tapped location from overlay and draws new dot on tap location.
+        /// - Parameter point: Location that the user tapped on the map.
```
yo1995, line 109:
```suggestion
        /// - Parameter point: Location to add a dot graphic on the map.
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+        func addGraphic(at point: Point) {
```
yo1995, line 110: Rename it to `addInputGraphic` would be more fitting because there are 2 graphics overlays.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+                // Sets indication when geoprocessingInProgress is true.
+                if geoprocessingInProgress {
+                    ProgressView("Geoprocessing in progress…")
+                        .padding()
+                        .background(.ultraThinMaterial)
+                        .cornerRadius(10)
+                        .shadow(radius: 50)
+                }
+            }
+            .task(id: tapScreenPoint) {
```
yo1995, line 43: If you tap the map again while the geoprocessing job is running, you'll see an error in the log

> Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.

It's a good learning opportunity to understand why it happens. If you have more questions, we can discuss.

yo1995, line 43: If you tap the map again while the geoprocessing job is running, you'll see that the translucent hud is no longer appearing. Try to fix it 😉 and think about creative ways to actually derive the in progress status from the job.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,197 @@
...
+            // Sets the parameters spatial reference to the point tapped on the map.
+            params.processSpatialReference = featureCollectionTable.spatialReference
+            params.outputSpatialReference = featureCollectionTable.spatialReference
+            // Create a feature with the feature collection table which has the reference to the tapped location.
+            let geoprocessingFeature = GeoprocessingFeatures(features: featureCollectionTable)
+            // Sets the observation point to the tapped location.
+            params.setInputValue(geoprocessingFeature, forKey: "Input_Observation_Point")
+            // Creates geoprocessing job and kicks off process of getting viewshed for the point.
+            geoprocessingJob = geoprocessingTask.makeJob(parameters: params)
+            geoprocessingJob?.start()
```
yo1995, line 144: Don't forget to release the reference the job after it finishes by setting it to `nil`. Otherwise, the memory won't be recycled, and the lifespan of  the object will be longer than needed.

yo1995 (commented):
More comments

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: should be handled now by the defer

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: This should be fixed nopw

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+
+struct ShowViewshedFromPointOnMapView: View {
+    /// The error shown in the error alert.
+    @State var error: Error?
```
yo1995, line 20:
```suggestion
    @State private var error: Error?
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+                geoprocessingInProgress = true
+                do {
+                    try await model.calculateViewshed(at: tapLocation)
+                    geoprocessingInProgress = false
+                } catch {
+                    self.error = error
+                    // This is set because errorAlert hides cancellation errors,
+                    // however when kicking off new job, we cancel any in progress
+                    // jobs which leads to wrong behavior.
+                    if !($error.wrappedValue is CancellationError) {
```
yo1995, line 50:
```suggestion
                    if !(error is CancellationError) {
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+            )
+            do {
+                // Creates a feature for the point tapped.
+                let feature = featureCollectionTable.makeFeature(geometry: point)
+                // Asynchronously adds that feature to the table.
+                try await featureCollectionTable.add(feature)
+                try await performGeoprocessing(featureCollectionTable)
+            } catch {
+                throw error
+            }
```
yo1995, line 144: If a method is throwing, no need to do-catch then throw the error again. Same for other methods below.

```suggestion
            // Creates a feature for the point tapped.
            let feature = featureCollectionTable.makeFeature(geometry: point)
            // Asynchronously adds that feature to the table.
            try await featureCollectionTable.add(feature)
            try await performGeoprocessing(featureCollectionTable)
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+        }
+        
+        /// If the feature set is returned from the geoprocessing, it iterates through each feature and adds it to the
+        /// graphic overlay to display to the user.
+        /// - Parameter resultFeatures: Passes on the results of the geoprocessing so that they can be displayed.
+        private func processFeatures(resultFeatures: GeoprocessingFeatures) {
+            if let featureSet = resultFeatures.features {
+                // Iterates through the feature set.
+                for feature in featureSet.features().makeIterator() {
```
yo1995, line 183:
```suggestion
                for feature in featureSet.features() {
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,202 @@
...
+            case .success(let output):
+                if let resultFeatures = output.outputs["Viewshed_Result"] as? GeoprocessingFeatures {
+                    processFeatures(resultFeatures: resultFeatures)
+                }
+            case .failure(let error):
+                throw error
+            case .none:
+                // This case should never execute.
+                break
+            }
```
yo1995, line 174: A few comment on this section

- Instead of using the `result`, use `output` of the `geoprocessingJob` can save you the switch-case
- Instead of unwrapping the feature set in the `processFeatures` method, you can unwrap it here, then pass the non-optional feature set into the following method. Therefore, `processFeatures` will be more intuitive.
- Instead of using the instance property `geoprocessingJob` to hold the job reference, I tend to use a local variable `job`. It saves several conditional chaining. You can also force unwrap if you prefer that style.

```swift
let job = geoprocessingTask.makeJob(parameters: params)
geoprocessingJob = job
defer { geoprocessingJob = nil }
job.start()
// Get the result of the geoprocessing job asynchronously.
let output = try await job.output
if let resultFeatures = output.outputs["Viewshed_Result"] as? GeoprocessingFeatures,
   let featureSet = resultFeatures.features {
    processFeatures(in: featureSet)  // private func processFeatures(in featureSet: FeatureSet)
}
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,188 @@
...
+            // Sets the observation point to the tapped location.
+            params.setInputValue(geoprocessingFeature, forKey: "Input_Observation_Point")
+            // Creates geoprocessing job and kicks off process of getting viewshed for the point.
+            geoprocessingJob = geoprocessingTask.makeJob(parameters: params)
+            geoprocessingJob?.start()
+            // Get the result of the geoprocessing job asynchronously.
+            let output = try await geoprocessingJob?.output
+            if let resultFeatures = output?.outputs["Viewshed_Result"] as? GeoprocessingFeatures,
+               let featureSet = resultFeatures.features {
+                processFeatures(in: featureSet)  // private func processFeatures(in featureSet: FeatureSet)
```
chriswebb09, line 161: needs to fix

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,188 @@
...
+            )
+            // Creates a feature for the point tapped.
+            let feature = featureCollectionTable.makeFeature(geometry: point)
+            // Asynchronously adds that feature to the table.
+            try await featureCollectionTable.add(feature)
+            try await performGeoprocessing(featureCollectionTable)
+        }
+        
+        /// Contains the logic for the geoprocessing and passes the result on to another function to display.
+        /// - Parameter featureCollectionTable: Holds the tapped location feature
```
yo1995, line 144:
```suggestion
        /// - Parameter featureCollectionTable: Holds the tapped location feature.
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,188 @@
...
+        private func performGeoprocessing(_ featureCollectionTable: FeatureCollectionTable) async throws {
+            let params = GeoprocessingParameters(executionType: .synchronousExecute)
+            // Sets the parameters spatial reference to the point tapped on the map.
+            params.processSpatialReference = featureCollectionTable.spatialReference
+            params.outputSpatialReference = featureCollectionTable.spatialReference
+            // Create a feature with the feature collection table which has the reference to the tapped location.
```
yo1995, line 150:
```suggestion
            // Creates a feature with the feature collection table which has the reference to the tapped location.
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,188 @@
...
+        private func processFeatures(in featureSet: FeatureSet) {
+            // Iterates through the feature set.
+            for feature in featureSet.features() {
+                // Creates the graphic for each feature's geometry.
+                let graphic = Graphic(geometry: feature.geometry)
+                // Sets the graphic on the overlay to display to the user.
+                resultGraphicsOverlay.addGraphic(graphic)
+            }
+        }
+    }
```
yo1995, line 177: No need to changed, can also be simplified into

```swift
resultGraphicsOverlay.addGraphics(featureSet.features().map { Graphic(geometry: $0.geometry) })
```

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,188 @@
...
+            let params = GeoprocessingParameters(executionType: .synchronousExecute)
+            // Sets the parameters spatial reference to the point tapped on the map.
+            params.processSpatialReference = featureCollectionTable.spatialReference
+            params.outputSpatialReference = featureCollectionTable.spatialReference
+            // Create a feature with the feature collection table which has the reference to the tapped location.
+            let geoprocessingFeature = GeoprocessingFeatures(features: featureCollectionTable)
+            // Sets the observation point to the tapped location.
+            params.setInputValue(geoprocessingFeature, forKey: "Input_Observation_Point")
+            // Creates geoprocessing job and kicks off process of getting viewshed for the point.
+            geoprocessingJob = geoprocessingTask.makeJob(parameters: params)
```
yo1995, line 155: I don't think the [comment](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/427#discussion_r1639053258) is fully addressed.

- job reference can be set to `nil` after it finishes. You can defer that
- either hold a local job reference, or force unwrap the `geoprocessingJob` so it doesn't get conditionally chained 3 times - `geoprocessingJob?.start()`, `geoprocessingJob?.output` and `output?.outputs["Viewshed_Result"]`

yo1995 (commented):
Mostly looking good

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: since it's a one liner, would it need to be in its own function?

yo1995: If you make this change, then no need to be in a separate method.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: Should be resolved now per the previous comments suggestions.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: Okay, yeah I added it because I was updating it for the other suggestions.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,181 @@
...
+            // Creates a feature with the feature collection table which has the reference to the tapped location.
+            let geoprocessingFeature = GeoprocessingFeatures(features: featureCollectionTable)
+            // Sets the observation point to the tapped location.
+            params.setInputValue(geoprocessingFeature, forKey: "Input_Observation_Point")
+            // Creates geoprocessing job and kicks off process of getting viewshed for the point.
+            let job = geoprocessingTask.makeJob(parameters: params)
+            geoprocessingJob = job
+            defer { geoprocessingJob = nil }
+            job.start()
+            // Get the result of the geoprocessing job asynchronously.
```
yo1995, line 159:
```suggestion
            // Gets the result of the geoprocessing job asynchronously.
```

yo1995 (approved):

yo1995 (approved):

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
```diff
@@ -0,0 +1,181 @@
...
+    
+    /// The data model for the sample.
+    @StateObject private var model = Model()
+    
+    var body: some View {
+        MapView(map: model.map, graphicsOverlays: [model.inputGraphicsOverlay, model.resultGraphicsOverlay])
+            .onSingleTapGesture { _, tapPoint in
+                tapLocation = tapPoint
+            }
+        // Disables tap while geoprocessing is in progress.
```
mhdostal, line 36: Just a spacing thing...
```suggestion
            // Disables tap while geoprocessing is in progress.
```

mhdostal (approved):
Looks great, just the one comment.

Shared/Samples/Show viewshed from point on map/ShowViewshedFromPointOnMapView.swift
chriswebb09: Yeah that was bothering me as well, Xcode's code formatter moved it back every time.

PR #434, opened 2024-07-01
[New] Show service area

chriswebb09:
## Description

This PR implements **`Show service area`** in `Analysis` category.

`swift/issues/5546`

## How To Test

Open view and load map. Add in facilities and barriers to map and then tap show service. The service graphics should render correctly in relation facilities and barriers. 

## Screenshots

show-service-area

PR #436, opened 2024-07-02
[Fix] Strict concurrency warnings

yo1995:
## Description

Fix some easy to fix concurrency warnings. Mostly just `static var = ` -> `static var` computed

To test, enable strict concurrency = complete from build settings.

|Warnings|
|-|
|iShot_2024-07-02_16 05 04|
|iShot_2024-07-02_16 05 54|
|iShot_2024-07-02_16 08 36|

### DIFF
--- a/Shared/Samples/Show grid/ShowGridView.swift
+++ b/Shared/Samples/Show grid/ShowGridView.swift
@@ -251,15 +251,17 @@ private extension ShowGridView {
 }
 
 private extension ArcGIS.Grid.LabelPosition {
-    static var allCases: [Self] = [
-        .allSides,
-        .center,
-        .topLeft,
-        .topRight,
-        .bottomLeft,
-        .bottomRight,
-        .geographic
-    ]
+    static var allCases: [Self] {
+        return [
+            .allSides,
+            .center,
+            .topLeft,
+            .topRight,
+            .bottomLeft,
+            .bottomRight,
+            .geographic
+        ]
+    }
     
     var label: String {
         switch self {
@@ -276,7 +278,7 @@ private extension ArcGIS.Grid.LabelPosition {
 }
 
 private extension LatitudeLongitudeGrid.LabelFormat {
-    static var allCases: [Self] = [.decimalDegrees, .degreesMinutesSeconds]
+    static var allCases: [Self] { [.decimalDegrees, .degreesMinutesSeconds] }
     
     var label: String {
         switch self {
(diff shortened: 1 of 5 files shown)


Shared/Samples/Show grid/ShowGridView.swift
```diff
@@ -251,15 +251,17 @@ private extension ShowGridView {
...
-    static var allCases: [Self] = [
-        .allSides,
-        .center,
-        .topLeft,
-        .topRight,
-        .bottomLeft,
-        .bottomRight,
-        .geographic
-    ]
+    static var allCases: [Self] {
```
njarecha, line 254: Why not use `let` like other cases?

yo1995: We generally favor using computed properties, see Phil's comment [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/24#discussion_r921690047). 

When it comes to [Plain-Old-Data](https://en.wikipedia.org/wiki/Passive_data_structure)-like types, such as `Int` or `Double` or `Char`, because the compiler optimization [results in the same assembly code](https://forums.swift.org/t/static-computed-property-vs-static-let-constant/20968/3) for `static let` vs `static var` computed, it doesn't matter too much which to use. For more complex data types such as collections and strings, we prefer computed var.

I can change them all to `static var`

Shared/Samples/Show grid/ShowGridView.swift
```diff
@@ -251,15 +251,17 @@ private extension ShowGridView {
...
+        return [
+            .allSides,
+            .center,
+            .topLeft,
+            .topRight,
+            .bottomLeft,
+            .bottomRight,
+            .geographic
+        ]
```
njarecha, line 263:
```suggestion
        [
            .allSides,
            .center,
            .topLeft,
            .topRight,
            .bottomLeft,
            .bottomRight,
            .geographic
        ]
```

njarecha (changes requested):
Looks good, just one suggestion.

Shared/Samples/Show grid/ShowGridView.swift
yo1995: image

Without return it would trigger a linter warning. Similar to [this](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/434#discussion_r1664910339)

njarecha (dismissed):

mhdostal (approved):

yo1995:
Thanks for the suggestions.

PR #439, opened 2024-07-04
[Fix] Strict concurrency warnings

yo1995:
## Description

This PR fixes some concurrency warnings. 

To test, enable strict concurrency = complete from build settings.

### DIFF
--- a/Shared/Supporting Files/Views/WebView.swift
+++ b/Shared/Supporting Files/Views/WebView.swift
@@ -41,19 +41,18 @@ struct WebView: UIViewRepresentable {
     }
     
     class Coordinator: NSObject, WKNavigationDelegate {
-        func webView(
-            _ webView: WKWebView,
-            decidePolicyFor navigationAction: WKNavigationAction,
-            decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
-        ) {
+        func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
             switch navigationAction.navigationType {
             case .linkActivated:
-                if let url = navigationAction.request.url, UIApplication.shared.canOpenURL(url) {
-                    UIApplication.shared.open(url)
+                if let url = navigationAction.request.url,
+                   await UIApplication.shared.canOpenURL(url) {
+                    Task { @MainActor in
+                        UIApplication.shared.open(url)
+                    }
                 }
-                decisionHandler(.cancel)
+                return .cancel
             default:
-                decisionHandler(.allow)
+                return .allow
             }
         }
     }
--- a/Shared/Supporting Files/Views/WebView.swift
+++ b/Shared/Supporting Files/Views/WebView.swift
@@ -41,14 +41,12 @@ struct WebView: UIViewRepresentable {
     }
     
     class Coordinator: NSObject, WKNavigationDelegate {
+        @MainActor
         func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
             switch navigationAction.navigationType {
             case .linkActivated:
-                if let url = navigationAction.request.url,
-                   await UIApplication.shared.canOpenURL(url) {
-                    Task { @MainActor in
-                        UIApplication.shared.open(url)
-                    }
+                if let url = navigationAction.request.url, UIApplication.shared.canOpenURL(url) {
+                    await UIApplication.shared.open(url)
                 }
                 return .cancel
             default:
--- a/Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
+++ b/Shared/Samples/Add custom dynamic entity data source/AddCustomDynamicEntityDataSourceView.Vessel.swift
@@ -26,7 +26,7 @@ extension AddCustomDynamicEntityDataSourceView {
         /// The location of the vessel.
         let geometry: Geometry
         /// The attributes of the vessel.
-        let attributes: [String: Any]
+        let attributes: [String: any Sendable]
     }
 }
 
@@ -57,7 +57,7 @@ extension AddCustomDynamicEntityDataSourceView.Vessel: Decodable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         
         let geometry = try container.decode(Geometry.self, forKey: .geometry)
-        let attributes: [String: Any] = try {
+        let attributes: [String: any Sendable] = try {
             let attributes = try container.decode(Attributes.self, forKey: .attributes)
             return [
                 Attributes.CodingKeys.mmsi.rawValue: attributes.mmsi,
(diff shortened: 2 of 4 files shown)


Shared/Supporting Files/Views/WebView.swift
```diff
@@ -41,19 +41,18 @@ struct WebView: UIViewRepresentable {
...
-        ) {
+        func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
             switch navigationAction.navigationType {
             case .linkActivated:
-                if let url = navigationAction.request.url, UIApplication.shared.canOpenURL(url) {
-                    UIApplication.shared.open(url)
+                if let url = navigationAction.request.url,
+                   await UIApplication.shared.canOpenURL(url) {
+                    Task { @MainActor in
+                        UIApplication.shared.open(url)
```
yo1995, line 50: I didn't find a way to suppress the warning for complaining that the `open` method takes in a non-Sendable dictionary, which literally is an empty dictionary by default. Thus I created its own little main actor context.

Shared/Supporting Files/Views/WebView.swift
```diff
@@ -41,19 +41,18 @@ struct WebView: UIViewRepresentable {
...
-            decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
-        ) {
+        func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
             switch navigationAction.navigationType {
             case .linkActivated:
-                if let url = navigationAction.request.url, UIApplication.shared.canOpenURL(url) {
-                    UIApplication.shared.open(url)
+                if let url = navigationAction.request.url,
+                   await UIApplication.shared.canOpenURL(url) {
+                    Task { @MainActor in
```
njarecha, line 49: Use deteched task 
```suggestion
                    Task.detached { @MainActor in
```

njarecha (changes requested):
Just one suggestion

Shared/Supporting Files/Views/WebView.swift
rolson: I'm not sure there is a difference here?

Shared/Supporting Files/Views/WebView.swift
rolson: `WKNavigationDelegate` is annotated as `MainActor`. So I think that you can try 2 things here instead:
1. Try annotating `func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {` as MainActor
2. Use `MainActor.assumeIsolated` instead of kicking off a Task

njarecha: Annotating the delegate as MainActor causes warning, 

Screenshot 2024-07-10 at 8 24 51 AM

To fix, it will have to add `extension WKNavigationAction: @unchecked Sendable {}`

Then it works with this,

```
    class Coordinator: NSObject, WKNavigationDelegate {
        @MainActor
        func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
            switch navigationAction.navigationType {
            case .linkActivated:
                if let url = navigationAction.request.url,
                    UIApplication.shared.canOpenURL(url) {
                    await UIApplication.shared.open(url)
                }
                return .cancel
            default:
                return .allow
            }
        }
    }
```

rolson: > To fix, it will have to add extension WKNavigationAction: @unchecked Sendable {}

We shouldn't do that.

There is no warning in Xcode 16 building with this code. The warning is only in Xcode 15.4.

```swift
    class Coordinator: NSObject, WKNavigationDelegate {
        @MainActor
        func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
            switch navigationAction.navigationType {
            case .linkActivated:
                if let url = navigationAction.request.url, UIApplication.shared.canOpenURL(url) {
                    await UIApplication.shared.open(url)
                }
                return .cancel
            default:
                return .allow
            }
        }
    }
```

I believe the reason is that WKNavigationAction is annotated as MainActor in the latest SDK:

```swift
@MainActor open class WKNavigationAction : NSObject {
```

Shared/Supporting Files/Views/WebView.swift
```diff
@@ -41,19 +41,18 @@ struct WebView: UIViewRepresentable {
...
-            decidePolicyFor navigationAction: WKNavigationAction,
-            decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
-        ) {
+        func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
             switch navigationAction.navigationType {
             case .linkActivated:
-                if let url = navigationAction.request.url, UIApplication.shared.canOpenURL(url) {
-                    UIApplication.shared.open(url)
+                if let url = navigationAction.request.url,
+                   await UIApplication.shared.canOpenURL(url) {
```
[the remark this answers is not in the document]
rolson: For some reason the compiler expects that it is async. I'm guessing because it translates the function with the  completion handler to an async method. Even though the completion handler is optional and a default argument of `nil` is provided:

[uiapplication](https://developer.apple.com/documentation/uikit/uiapplication/1648685-open)

Shared/Supporting Files/Views/WebView.swift
rolson: I do not see the warning with Complete Strict Concurrency with Xcode 16 Beta 3. But you can also try it before merging if you'd like.

rolson: (Yes, I had that setting `strict concurrency = complete` when I tried it)

njarecha: Should we be fixing this based on Xcode 16 (still in beta) or Xcode 15.4?

yo1995: I've tested with Xcode 16b3 and it indeed doesn't give warning. I have the same question as Nimesh though - should I use the Xcode 16 compatible fix that still gives warning in Xcode 15, or should I keep it as-is?

rolson: I would fix for Xcode 16. There is not a lot of point in fixing the warning now, if you have to re-fix it later.

njarecha (approved):
Looks good

Shared/Supporting Files/Views/WebView.swift
philium: Whatever solution you go with, make sure that it works—without warning—on the minimum Xcode version (Xcode 15.0).

rolson (approved):

Shared/Supporting Files/Views/WebView.swift
yo1995: > make sure that it works—without warning—on the minimum Xcode version (Xcode 15.0)

OK, I think my strategy will be turning on complete strict concurrency after `(the min req becomes Xcode 16) && (most if not all strict concurrency warnings are addressed)`. That way, those Xcode 16+ concurrency fixes won't cause trouble in any Xcode 15.0 min req releases.

yo1995:
Thanks for the thoughts.

PR #452, opened 2024-07-16
[Update] Create and edit geometries sample

des12437:
## Description

This PR updates `Create and edit geometries`
URL to README: https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/destiny/Update-Create-and-edit-geometries-sample/Shared/Samples/Create%20and%20edit%20geometries#readme

This PR updates the sample to match the sample design with the following changes:
- Updates the basemap and initial viewpoint
- Changes the styling of the graphics
- Adds initial graphics
- Allows deleting all graphics including the initial graphics

## Screenshots

create-and-edit-geometries-1  create-and-edit-geometries-2


dfeinzimer (approved):

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -387,6 +415,115 @@ private class GeometryEditorModel: ObservableObject {
     }
 }
 
+private extension Geometry {
+    /// The area around the island of Inis Meáin (Aran Islands) in Ireland.
+    static var aranIslands: Envelope {
+        Envelope(center: Point(x: -9.5920, y: 53.08230, spatialReference: .wgs84), width: 1, height: 1, depth: 1)
+    }
+
```
yo1995, line 423:
```suggestion
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -314,6 +321,28 @@ private class GeometryEditorModel: ObservableObject {
...
+        
+        let houseGraphic = Graphic(geometry: .house(), symbol: .point)
+        
+        geometryOverlay.addGraphics([
+            boundaryGraphic,
+            road1Graphic,
+            road2Graphic,
+            outbuildingsGraphic,
+            houseGraphic
+        ])
```
yo1995, line 341: The existing geometries are there for showing the editing workflow. Can you add the identify operation so that editing existing geometries also works, as described in the implementation details? Thank you. See around `onIdentifyGraphicsOverlayStatusChanged` symbol in the Qt example.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -413,24 +440,30 @@ private class GeometryEditorModel: ObservableObject {
         geometryEditor.start(withType: geometryType)
         isStarted = true
     }
+    
+    /// Starts editing a given graphic with the geometry editor.
+    /// - Parameter graphic: The graphic to edit.
+    func startEditing(withGraphic graphic: Graphic) {
+        guard let geometry = graphic.geometry else { return }
+        selectedGraphic = graphic
+        selectedGraphic?.isVisible = false
```
yo1995, line 449:
```suggestion
        graphic.isVisible = false
```

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -364,6 +390,7 @@ private class GeometryEditorModel: ObservableObject {
     func stop() {
         geometryEditor.stop()
```
yo1995, line 391: After reading the implementation details, I like the idea to showcase that `stop` method returns the final geometry, because it might be something that a user won't discover by merely following this sample, without reading the doc.

Instead of hiding then removing the graphic being edited, can you modify the geometry instead? Thanks.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -380,11 +453,136 @@ private class GeometryEditorModel: ObservableObject {
     ///   - tool: The tool to draw with.
     ///   - geometryType: The type of geometry to draw.
     func startEditing(with tool: GeometryEditorTool, geometryType: Geometry.Type) {
+        selectedGraphic = nil
```
yo1995, line 456: Why is this line needed? 🤔

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -380,11 +453,136 @@ private class GeometryEditorModel: ObservableObject {
...
+    /// - Parameter graphic: The graphic to edit.
+    func startEditing(withGraphic graphic: Graphic) {
+        guard let geometry = graphic.geometry else { return }
+        
+        switch geometry {
+        case is Point, is Multipoint:
+            geometryEditor.tool = VertexTool()
+        default:
+            break
+        }
```
yo1995, line 473: Is this needed?

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -380,11 +453,136 @@ private class GeometryEditorModel: ObservableObject {
...
+        selectedGraphic = nil
         configureGeometryEditorTool(tool, scaleMode: scaleMode)
         geometryEditor.tool = tool
         geometryEditor.start(withType: geometryType)
         isStarted = true
     }
+    
+    /// Starts editing a given graphic with the geometry editor.
+    /// - Parameter graphic: The graphic to edit.
+    func startEditing(withGraphic graphic: Graphic) {
```
yo1995, line 465: I'm not entirely sure here, do you think it is Swift API Guideline compliant to name it `func startEditing(with graphic: Graphic)`?

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
```diff
@@ -380,11 +453,136 @@ private class GeometryEditorModel: ObservableObject {
...
+        guard let geometry = graphic.geometry else { return }
```
yo1995, line 466:
```suggestion
        let geometry = graphic.geometry!
```
Since we can tap on the graphic, the geometry must be present.

yo1995 (commented):
Overall looks good. A few minor questions

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
des12437: Yes, it sets the geometry editor tool based on the geometry. This logic is specified in the sample design implementation details.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
des12437: This line can be removed.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
des12437: Removed the repeated type information from the method signature.

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
yo1995: I'll ask the design author Andy why it is needed. Without these lines, the geometry editor still works, so I prefer removing them. Will let you know tomorrow if he gets back.

yo1995 (commented):
LGTM with one thing to confirm.

dfeinzimer (dismissed):

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
des12437: Removed the unnecessary lines.

dfeinzimer (approved):

yo1995 (approved):

Shared/Samples/Create and edit geometries/CreateAndEditGeometriesView.swift
yo1995: After consulting bistromath team, this logic needs to be added back.

yo1995:
A update that I don't know where to put, so just leave here as a reference, in case we revisit these [changes](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/452#discussion_r1683596409) in the future 🤔 

- If this logic is not added, there is a case such as using shape tool to edit `Point` will result in the GE not being able to handle the geometry so the geometry will be hidden by the GE
- If this logic is expanded to, say, edit every geometry with vertex tool, it totally works, but narrows down the tool to vertex tool will probably give a false indication to a user that one can only edit a geometry using vertex tool, which isn't true.
- Thus, the SDC decided although this logic is not perfect, it strikes a balance between being simple and being able to not surprise a user with unexpected behavior.

PR #460, opened 2024-07-17
[Fix] Remove workaround for toolbar bug in AR samples

des12437:
## Description

This PR removes the workaround for an issue in the AR samples where AR content would cover the toolbar.

## Linked Issue(s)

- `swift/issues/5112`

## How To Test
- Use the `Augment reality to collect data`, `Augment reality to navigate route`, or  `Augment reality to show hidden infrastructure` samples and verify that AR content does not cover the toolbar.
- Test on an iOS 16 device to verify fix on all iOS versions.

## Screenshots

|Before|After|


des12437:
Closing since this fix does not work for iOS 16.

PR #462, opened 2024-07-19
[New] Show device location using indoor positioning 

chriswebb09:
## Description

This PR implements **`Show device location using indoor positioning`** in `Map` category.

- `swift/issues/5653` (internal test data in the issue)
- `common-samples/pull/3939`

## How To Test

Load map and wait for campus to load. You should see your location relative to the layout of the office.

Please note:

- The sample data doesn't work in Building M due to no pathway table was surveyed, and not in Building L either maybe due to dead batteries. To see it in action, use the internal data in the issue linked above.
- Both the `indoorPositioningDefinition` and the manual `ips_positioning` creation of the LDS should work, so make sure you test both condition branches

## Screenshots

351891359-cf184145-a083-48f0-a5a0-3dbc6ca7c106


Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,220 @@
...
+        /// Sets the indoor datasource on the location display depending on
+        /// whether the map contains an IndoorDefinition.
+        /// - Parameter map: The map which is checked for an indoor definition.
+        private func setIndoorDatasource(for type: DataSourceType) async throws {
+            await indoorsLocationDataSource?.stop()
+            indoorsLocationDataSource = nil
+            try await map.floorManager?.load()
+            switch type {
+            case .indoorDefinition:
+                if try await loadAndCheckForIndoorDefinition(map: map),
```
chriswebb09, line 105: @yo1995 Given that we are selecting the mode, what should be the behavior if indoorsDefinition is selected but not present in the map?

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,120 @@
...
+                        .background(.ultraThinMaterial)
+                        .cornerRadius(10)
+                        .shadow(radius: 50)
+                }
+            }
+            .task(id: selectedDataSourceType) {
+                model.isLoading = true
+                do {
+                    try await model.map.load()
+                    mapIsLoaded = true
```
chriswebb09, line 57: mapIsLoading = false

Samples.xcodeproj/project.pbxproj
```diff
@@ -522,6 +527,9 @@
 			dstPath = "";
 			dstSubfolderSpec = 7;
 			files = (
+				9503056F2C46ED010091B32D /* ShowDeviceLocationUsingIndoorPositioningView.Model.swift in Copy Source Code Files */,
+				9503056C2C46EC8A0091B32D /* ShowServiceAreaView.swift in Copy Source Code Files */,
```
yo1995, line 531: Can you move this change into a separate PR? Thanks.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  ShowDeviceLocationUsingIndoorPositioningView.swift
+//  Samples
```
yo1995, line 3:
```suggestion
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+                positioningTable: positioningTable,
+                pathwaysTable: pathwaysLayer.featureTable as? ArcGISFeatureTable,
+                levelsTable: levelsLayer.featureTable as? ArcGISFeatureTable,
+                positioningID: positioningID
+            )
+        }
+        
+        /// The method that updates the location when the indoors location datasource is triggered.
+        /// - Parameter floorManager: The floor manager that filters what is displayed on the map by floor.
+        private func dataChangesOnLocationUpdate() async throws {
```
yo1995, line 144: The docstring doesn't match the method's signature

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,207 @@
...
+
+import ArcGIS
+import CoreLocation
+import SwiftUI
+
+extension ShowDeviceLocationUsingIndoorPositioningView {
+    @MainActor
+    class Model: ObservableObject {
+        /// Basic map with topographic style.
+        var map = Map(url: .indoorsMap)!
```
yo1995, line 23:
```suggestion
        let map = Map(url: .indoorsMap)!
```

Shared/Samples/Show device location using indoor positioning/README.md
```diff
@@ -0,0 +1,50 @@
...
+* MapView
+
+## About the data
+
+This sample uses an [IPS-enabled web map](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) that displays Building L on the Esri Redlands campus. Please note: you would only be able to use the indoor positioning functionalities when you are inside this building. Swap the web map to test with your own IPS setup.
+
+## Additional information
+
+* Location and Bluetooth permissions are required for this sample.
+* To learn more about IPS, read the [Indoor positioning](https://developers.arcgis.com/ios/device-location/indoor-positioning/) article on ArcGIS Developer website.
```
yo1995, line 45:
```suggestion
* To learn more about IPS, read the [Indoor positioning](https://developers.arcgis.com/swift/device-location/indoor-positioning/) article on Esri Developer website.
```

Shared/Samples/Show device location using indoor positioning/README.md
```diff
@@ -0,0 +1,50 @@
...
+* IndoorsLocationDataSource
+* LocationChangeHandlerDelegate
+* LocationDisplay
+* LocationDisplayAutoPanMode
+* Map
+* MapView
+
+## About the data
+
+This sample uses an [IPS-enabled web map](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) that displays Building L on the Esri Redlands campus. Please note: you would only be able to use the indoor positioning functionalities when you are inside this building. Swap the web map to test with your own IPS setup.
```
yo1995, line 40:
```suggestion
This sample uses an [IPS-aware web map](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) that displays Building L on the Esri Redlands campus. Please note: you would only be able to use the indoor positioning functionalities when you are inside this building. Swap the web map to test with your own IPS setup.
```

Samples.xcodeproj/project.pbxproj
chriswebb09: You meant just the ShowServiceAreaView one correct?

yo1995: Right. We don't include change that has side effects in a PR even if it is trivial. Because if the PR is reverted in the future, the unrelated change will be reverted as well, which might cause problems.
Instead, create a quick fix PR is better.

chriswebb09: Okay I'll create quickfix pr

chriswebb09: Looks like it is already fixed in v.next

chriswebb09:
IMG_4126
updated UI

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,226 @@
...
+                }
+                labelTextTrailing += "Data source: \(source)"
+            } else {
+                labelTextLeading = "No floor data."
+            }
+        }
+        
+        /// Starts the location display to show user's location on the map.
+        private func requestLocationServicesAuthorizationIfNecessary() async throws {
+            // Request location permission if it has not yet been determined.
```
chriswebb09, line 214: Requests location permission if it has not yet been determined.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,70 @@
...
+                        .cornerRadius(10)
+                        .shadow(radius: 50)
+                }
+            }
+            .task {
+                model.isLoading = true
+                do {
+                    try await model.map.load()
+                    try await model.displayIndoorData()
+                    // Since the method dataChangesOnLocationUpdate listens for new location changes, it is important
```
chriswebb09, line 52: Referenced method has been renamed to updateAndDisplayOnLocationChange()

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,226 @@
...
+            } else {
+                indoorsLocationDataSource = try await createIndoorLocationDataSource(map: map)
+            }
+            // This ensures that the details of the inside of the building, like room layouts are displayed.
+            for featLayer in map.operationalLayers {
+                if featLayer.name == "Transitions" || featLayer.name == "Details" {
+                    featLayer.isVisible = true
+                }
+            }
+            guard let dataSource = indoorsLocationDataSource else { return }
```
chriswebb09, line 111: Add comment that the datasource should always be there, but createIndoorLocationDataSource throws an optional value so it cannot be guaranteed. Best option is you get to this point without a datasource is to return (ideally an error would have been thrown before this point and the flow broken.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,234 @@
...
+        
+        /// A function that attempts to load an indoor definition attached to the map
+        /// and returns a boolean value based whether it is loaded.
+        /// - Parameter map: The map that contains the IndoorDefinition.
+        /// - Returns: A boolean value for whether the IndoorDefinition is loaded.
+        private func indoorDefinitionIsLoaded(map: Map) async throws -> Bool {
+            guard map.indoorPositioningDefinition?.loadStatus != .loaded else { return true }
+            try await map.indoorPositioningDefinition?.load()
+            return map.indoorPositioningDefinition?.loadStatus == .loaded
+        }
```
yo1995, line 92: This method seems to be extraneous. The sample only loads the web map once, so the load status will likely be either `notLoaded` -> `loaded` or -> `failed`. Also, if you really need the retry workflow, there is the `retryLoad` method for Loadable resources for certain specific situations.

I suggest move the nullability check of `map.indoorPositioningDefinition` and `try await map.indoorPositioningDefinition.load()` method call to the spot where it is called.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,234 @@
...
+        
+        /// Sets the indoor datasource on the location display depending on
+        /// whether the map contains an IndoorDefinition.
+        private func setIndoorDatasource() async throws {
+            try await map.floorManager?.load()
+            // If an indoor definition exists in the map, it gets loaded and sets the IndoorsDataSource to pull information
+            // from the definition.
+            if try await indoorDefinitionIsLoaded(map: map),
+               let indoorPositioningDefinition = map.indoorPositioningDefinition {
```
yo1995, line 101: The `indoorPositioningDefinition` needs to be not nil before it can be loaded. Please refactor.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,234 @@
...
+                indoorsLocationDataSource = IndoorsLocationDataSource(definition: indoorPositioningDefinition)
+                // Otherwise the IndoorsDataSource attempts to create itself using IPS table information.
```
yo1995, line 103: When commenting in `if-else` conditions, we typically put it in their respective scope. See `DownloadPreplannedMapAreaView.Model` line 102 for an example.

Shared/Samples/Show device location using indoor positioning/README.md
yo1995, line 1: The design update PR has went in `common-samples/pull/3939`, please add the latest changes to the README here. Thank you!

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    
+    var body: some View {
+        MapView(map: model.map)
+            .locationDisplay(model.locationDisplay)
+            .overlay(alignment: .top) {
+                HStack {
+                    Text(model.labelTextLeading)
+                        .frame(width: 140)
```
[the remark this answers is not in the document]
chriswebb09: I think no width is better than maxwidth 
No width:
IMG_4213
Max width: 
Screenshot 2024-08-01 at 2 10 17 PM

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,66 @@
...
+                        .shadow(radius: 50)
+                        .multilineTextAlignment(.center)
+                }
+            }
+            .task {
+                do {
+                    try await model.loadIndoorData()
+                    try await model.updateDisplayOnLocationChange()
+                } catch {
+                    model.isLoading = false
```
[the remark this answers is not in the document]
chriswebb09: I moved the isLoading defer into the model.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,66 @@
...
+                        .padding()
+                        .background(.ultraThinMaterial)
+                        .cornerRadius(10)
+                        .shadow(radius: 50)
+                        .multilineTextAlignment(.center)
+                }
+            }
+            .task {
+                do {
+                    try await model.loadIndoorData()
```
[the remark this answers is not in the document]
chriswebb09: I've moved it to view and it launches before sourcing the indoor data.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,234 @@
...
+        let map = Map(url: .indoorsMap)!
+        
+        /// The value of the current floor with -1 being used to represent floor that has not been set.
+        private var currentFloor: Int = -1
+        
+        /// The number of BLE sensors which are being used for indoor location.
+        private var sensorCount: Int = -1
+        
+        /// Counts the number of satellites which being used for the GPS location.
+        private var satelliteCount: Int = -1
```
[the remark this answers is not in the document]
chriswebb09: Converted to optionals.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,234 @@
...
+        private var indoorsLocationDataSource: IndoorsLocationDataSource?
+        
+        /// The map's location display.
+        private(set) var locationDisplay = LocationDisplay(dataSource: SystemLocationDataSource())
+        
+        /// The location manager which handles the location data.
+        private let locationManager = CLLocationManager()
+        
+        /// The measurement formatter for sensor accuracy.
+        private let measurementFormatter: MeasurementFormatter = {
```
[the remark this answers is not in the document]
chriswebb09: I think I've got it, but I'm taking my cues from what has already been implemented.

Shared/Samples/Show device location using indoor positioning/README.md
chriswebb09: Sure thing, let me know if this is what you were expecting in terms of README updates.

chriswebb09:
@philium should be ready for you to look at when you get the chance.

Shared/Samples/Show device location using indoor positioning/show-device-location-using-indoor-positioning.png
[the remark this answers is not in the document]
chriswebb09: updated

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import CoreLocation
+import SwiftUI
```
philium, line 17: This file doesn't depend on SwiftUI. It does depend on Combine.

```suggestion
import Combine
import CoreLocation
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+            try await map.load()
+            try await setIndoorDatasource()
+        }
+        
+        /// Stops the location data source.
+        func stopLocationDataSource() {
+            Task {
+                await locationDisplay.dataSource.stop()
+            }
+        }
```
philium, line 66: Instead of hiding the fact that calling this method performs an asynchronous operation, I would suggest making it async and letting the caller wrap the call in a task if needed, just like the other methods of this model.

```swift
func stopLocationDataSource() async {
    await locationDisplay.dataSource.stop()
}
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+        /// An IPS-aware web map for all three floors of Esri Building L in Redlands.
+        let map = Map(url: .indoorsMap)!
+        
+        /// The value of the current floor with -1 being used to represent floor that has not been set.
+        private var currentFloor: Int?
+        
+        /// The number of BLE sensors which are being used for indoor location.
+        private var sensorCount: Int?
+        
+        /// Counts the number of satellites which being used for the GPS location.
```
philium, line 31:
```suggestion
        /// The number of satellites which are being used for the GPS location.
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+        
+        /// Stops the location data source.
+        func stopLocationDataSource() {
+            Task {
+                await locationDisplay.dataSource.stop()
+            }
+        }
+        
+        /// Sets the indoor datasource on the location display depending on
+        /// whether the map contains an IndoorDefinition.
```
philium, line 69: Not a huge deal, but the map may contain an indoor definition (instance), not an IndoorDefinition (class).

```suggestion
        /// whether the map contains an indoor definition.
```

Same elsewhere.

Note: When referring to a type or member in a documentation comment, use code voice (enclose in parentheses).

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+            }
+            // This ensures that the details of the inside of the building, like room layouts are displayed.
+            for featLayer in map.operationalLayers {
+                if featLayer.name == "Transitions" || featLayer.name == "Details" {
+                    featLayer.isVisible = true
+                }
+            }
+            // The indoorsLocationDataSource should always be there. Since the createIndoorLocationDataSourceFromTables returns
+            // an optional value it cannot be guaranteed. Best option if you get to this point without a datasource
+            // is to return (ideally an error would have been thrown before this point and the flow broken.)
```
philium, line 88: Since one of the goals of samples is to model best practices for our customers, we should do what is ideal if at all possible.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+                indoorsLocationDataSource = IndoorsLocationDataSource(definition: indoorPositioningDefinition)
+            } else {
+                indoorsLocationDataSource = try await createIndoorLocationDataSourceFromTables(map: map)
+            }
+            // This ensures that the details of the inside of the building, like room layouts are displayed.
+            for featLayer in map.operationalLayers {
+                if featLayer.name == "Transitions" || featLayer.name == "Details" {
+                    featLayer.isVisible = true
+                }
+            }
```
philium, line 85: The type of `operationalLayers` is `[Layer]`, not `[FeatureLayer]`, so a better name for the loop variable would be `layer`. Also, you could use a where clause instead of an if statement to filter the layers.

```suggestion
            for layer in map.operationalLayers where layer.name == "Transitions" || layer.name == "Details" {
                layer.isVisible = true
            }
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+        private func createIndoorLocationDataSourceFromTables(map: Map) async throws -> IndoorsLocationDataSource? {
+            // Gets the positioning table from the map.
+            guard let positioningTable = map.tables.first(where: { $0.displayName == "IPS_Positioning" }) else { return nil }
+            // Creates and configures the query parameters.
+            let queryParameters = QueryParameters()
+            queryParameters.maxFeatures = 1
+            queryParameters.whereClause = "1 = 1"
+            // Queries positioning table to get the positioning ID.
+            let queryResult = try await positioningTable.queryFeatures(using: queryParameters)
+            guard let feature = queryResult.features().makeIterator().next() else { return nil }
```
philium, line 108: Here's another way of doing this.

```suggestion
            guard let feature = queryResult.features().first(where: { _ in true }) else { return nil }
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+            let serviceFeatureTable = positioningTable as! ServiceFeatureTable
+            let positioningID = feature.attributes[serviceFeatureTable.globalIDField] as? UUID
+            
+            // Gets the pathways layer (optional for creating the IndoorsLocationDataSource).
+            let pathwaysLayer = map.operationalLayers.first(where: { $0.name == "Pathways" }) as! FeatureLayer
+            // Gets the levels layer (optional for creating the IndoorsLocationDataSource).
+            let levelsLayer = map.operationalLayers.first(where: { $0.name == "Levels" }) as! FeatureLayer
```
philium, line 115: Here you're assuming that these layers will always be there by force unwrapping them, but you didn't do that for the positioning table. How come?

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+            
+            // Setting up IndoorsLocationDataSource with positioning, pathways tables and positioning ID.
+            // positioningTable - the "IPS_Positioning" feature table from an IPS-aware map.
+            // pathwaysTable - An ArcGISFeatureTable that contains pathways as per the ArcGIS Indoors Information Model.
+            // Setting this property enables path snapping of locations provided by the IndoorsLocationDataSource.
+            // levelsTable - An ArcGISFeatureTable that contains floor levels in accordance with the ArcGIS Indoors Information Model.
+            // Providing this table enables the retrieval of a location's floor level ID.
+            // positioningID - an ID which identifies a specific row in the positioningTable that should be used for setting up IPS.
```
philium, line 123: This is essentially just duplicating the documentation for the parameters of `IndoorsLocationDataSource.init(positioningTable:pathwaysTable:levelsTable:positioningID:)`. I would suggest getting rid of it.

```suggestion
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+            return IndoorsLocationDataSource(
+                positioningTable: positioningTable,
+                pathwaysTable: pathwaysLayer.featureTable as? ArcGISFeatureTable,
+                levelsTable: levelsLayer.featureTable as? ArcGISFeatureTable,
+                positioningID: positioningID
+            )
+        }
+        
+        /// The method that updates the location when the indoors location datasource is triggered.
```
philium, line 132:
```suggestion
        /// Updates the location when the indoors location datasource is triggered.
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+                if let floorLevel = location.additionalSourceProperties[.floor] as? Int,
+                   (floorLevel + 1) != currentFloor {
+                    // Sets the currentFloor to the new floor level and adds one, since location uses
+                    // zero based flooring system.
+                    currentFloor = floorLevel + 1
+                    // The floor manager is used to filter so that only the data from the current floor is
+                    // displayed to the user.
+                    if let floorManager = map.floorManager {
+                        floorManager.levels.forEach {
+                            $0.isVisible = currentFloor == $0.levelNumber
```
philium, line 149: `Sequence.forEach(_:)` is intended to only be used at the end of a method chain. It shouldn't be used when a for loop would work instead.

```suggestion
                        for level in floorManager.levels {
                            level.isVisible = currentFloor == level.levelNumber
```

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+                    sensorCount = location.additionalSourceProperties[.transmitterCount] as? Int ?? 0
+                }
+                horizontalAccuracy = location.horizontalAccuracy
+                // Updates every time the location changes.
+                getStatusLabelText(source: source)
+            }
+        }
+        
+        /// Updates the labels on the view with the current state of the indoors data source.
+        private func getStatusLabelText(source: String) {
```
philium, line 169: The prefix `get` implies that the method returns something, but in fact it doesn't return anything. Not only that, this is something the view should be doing, not the model. The model shouldn't be determining user-facing strings, particularly ones containing values that should be formatted appropriately. (After all, we want to model best practices to our customers, which includes formatting data appropriately.) The view should be doing that. Please move this logic and the corresponding properties into the view. The view should use the state of the model to determine what should be displayed.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,202 @@
...
+        URL(string: "https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba")!
+    }
+}
+
+private extension FormatStyle where Self == Measurement<UnitLength>.FormatStyle {
+    /// The format style for the distances.
+    static var distance: Self {
+        .measurement(width: .abbreviated, usage: .asProvided, numberFormatStyle: .number.precision(.fractionLength(2)))
+    }
+}
```
philium, line 202: Data formatting logic in the model is a big code smell. Data should be formatted based on the current locale, but the model doesn't have that information (because it comes from the environment of the view).

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,75 @@
...
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    
+    var body: some View {
+        MapView(map: model.map)
+            .locationDisplay(model.locationDisplay)
+            .overlay(alignment: .top) {
+                HStack(alignment: .center, spacing: 10) {
+                    Text(model.labelTextLeading)
+                    Text(model.labelTextTrailing)
```
philium, line 31: The HUD text is a little difficult to read. Consider making the text darker and/or the background not as transparent. Even better, consider using a [material](https://developer.apple.com/documentation/swiftui/material/) for the background so that the text has a vibrancy effect.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,75 @@
...
+                .opacity(model.labelTextLeading.isEmpty ? 0 : 0.5)
+            }
+            .overlay(alignment: .center) {
+                if model.isLoading {
+                    ProgressView("Loading\n indoor data…")
+                        .padding()
+                        .background(.ultraThinMaterial)
+                        .cornerRadius(10)
+                        .shadow(radius: 50)
+                        .multilineTextAlignment(.center)
```
philium, line 44: Consider using a multiline string literal.

```suggestion
                    ProgressView(
                        """
                        Loading
                        indoor data…
                        """
                    )
                    .padding()
                    .background(.ultraThinMaterial)
                    .cornerRadius(10)
                    .shadow(radius: 50)
                    .multilineTextAlignment(.center)
```

philium (commented):
Thank you for implementing this sample! I can tell that you have put a lot of time and effort into it. I have reviewed your PR and have left several comments.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
chriswebb09: I had it previously with the if let, but I looked at the how we have it in the guide online and wanted to keep the code as close to what was posted to users on the web as possible. I would probably not force unwrap them if it was purely my implementation, because its possible the data could not be there and then you'd have the app store version crash always without much explanation.

philium: If we're genuinely concerned that the data may change, then we should be consistently safe and not force unwrap any of the data. Check with your team, but I think the likelihood of the data changing is so low that it would be fine to force unwrap, provided that it is guaranteed that the map has already been loaded.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
chriswebb09: So should I steer this sample code further away from what is in the indoors developer documentation for iOS?

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
chriswebb09: Okay sounds good!

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
chriswebb09: Thanks for the tip!

chriswebb09:
Screenshot 2024-08-08 at 11 47 57 AM

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
chriswebb09: I've posted updated screenshot. I'll update the project screenshot if it looks good to you.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,109 @@
...
+    
+    /// Updates the labels on the view with the current state of the indoors data source.
+    private func updateStatusTextDisplay() {
+        labelTextLeading = ""
+        labelTextTrailing = ""
+        if let currentFloor = model.currentFloor {
+            labelTextLeading += "Current floor: \(currentFloor)\n"
+            if let horizontalAccuracy = model.horizontalAccuracy {
+                let measuredAccuracy = Measurement(value: horizontalAccuracy, unit: UnitLength.meters)
+                labelTextLeading += "Accuracy: \(measuredAccuracy.formatted(.measurement(width: .abbreviated, usage: .asProvided, numberFormatStyle: .number.precision(.fractionLength(2)))))"
```
chriswebb09, line 93: It looks very messy if the number isn't rounded to two decimal places. @philium

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,154 @@
...
+        /// Updates the location when the indoors location datasource is triggered.
+        func updateDisplayOnLocationChange(locationUpdate: () -> Void) async throws {
+            for await location in locationDisplay.dataSource.locations {
+                // Since this listens for new location changes, it is important
+                // to ensure any blocking UI is dismissed once location updates begins.
+                // Floors in location are zero indexed however floorManager levels begin at one. Since
+                // it is necessary to display the same information to the user as the floor manager levelNumber
+                // one is added to the floor level value.
+                if let floorLevel = location.additionalSourceProperties[.floor] as? Int,
+                   (floorLevel + 1) != currentFloor {
```
philium, line 122: This is akin to a [Yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions), which we avoid as a matter of style. Please use `floorLevel != currentFloor - 1` instead.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,109 @@
...
+
+struct ShowDeviceLocationUsingIndoorPositioningView: View {
+    /// The data model for the sample.
+    @StateObject private var model = Model()
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    /// This is the published value of the data that is displayed.
+    @State private var labelTextLeading: String = ""
+    /// This is the published value of the data that is displayed.
+    @State private var labelTextTrailing: String = ""
```
philium, line 27: User-visible text should not be calculated ahead of time as the underlying data may not the only thing that affects what is displayed to the user. For formatted values, the value may be re-formatted due to a locale change as the user moves across a border. Please re-factor to calculate the text to display to the user where it is needed. Instead of concatenating strings together with newlines, just use separate Text values in an H stack or V stack (whichever is appropriate). Instead of calling a `formatted` method on the value to be formatted, allow SwiftUI to do the formatting so that it can efficiently format the value only when needed. For example: `Text("Accuracy: \(accuracyMeasurement, format: ...)")`.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+    
+    /// Updates the labels on the view with the current state of the indoors data source.
+    private func updateStatusTextDisplay() {
+        labelTextLeading = ""
+        labelTextTrailing = ""
+        if let currentFloor = model.currentFloor {
+            labelTextLeading += "Current floor: \(currentFloor)\n"
+            if let horizontalAccuracy = model.horizontalAccuracy {
+                let measuredAccuracy = Measurement(value: horizontalAccuracy, unit: UnitLength.meters)
```
philium, line 92: A better name for this would be `accuracyMeasurement`.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,100 @@
...
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    /// This is the text description of the current floor that is displayed.
+    @State private var currentFloorText: String = ""
+    /// This is the text description of the data source that is displayed.
+    @State private var sourceText: String = ""
+    /// This is the text description of the number sensor of the data that is displayed.
+    @State private var sensorText: String = ""
+    /// This is the text description of the horizontal accuracy data that is displayed.
+    @State private var accuracyText: String = ""
```
philium, line 31: These are no longer being used.

```suggestion
```

Whenever you refactor code, don't just stop there. Always do the next step of asking "In light of these changes, what else needs to change?" A good thing to look for is code (types, variables, functions, etc.) that is no longer needed.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,100 @@
...
+        MapView(map: model.map)
+            .locationDisplay(model.locationDisplay)
+            .overlay(alignment: .top) {
+                HStack(alignment: .center, spacing: 10) {
+                    VStack(alignment: .leading) {
+                        Text("Current floor: \(model.currentFloor ?? 0)")
+                        Text(
+                            """
+                          Accuracy: \(Measurement(value: model.horizontalAccuracy ?? 0.0, unit: UnitLength.meters), format: .measurement(width: .abbreviated, usage: .asProvided, numberFormatStyle: .number.precision(.fractionLength(2))))
+                          """
```
philium, line 43: A multiline string literal, as the name implies, should only be used when the string literal contains multiple lines.

```suggestion
                            "Accuracy: \(Measurement(value: model.horizontalAccuracy ?? 0.0, unit: UnitLength.meters), format: .measurement(width: .abbreviated, usage: .asProvided, numberFormatStyle: .number.precision(.fractionLength(2))))"
```

philium (commented):
Thank you for making those changes. Just a couple more, as far as I can see. Getting close!

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
chriswebb09: Thanks for pointing this out, appreciate the feedback.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
chriswebb09: Updated! Thanks for the tip!

chriswebb09:
> Thank you for making those changes. Just a couple more, as far as I can see. Getting close!

Glad to make the changes, I appreciate all your feedback!

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -21,27 +21,15 @@ struct ShowDeviceLocationUsingIndoorPositioningView: View {
...
-    /// This is the text description of the current floor that is displayed.
-    @State private var currentFloorText: String = ""
-    /// This is the text description of the data source that is displayed.
-    @State private var sourceText: String = ""
-    /// This is the text description of the number sensor of the data that is displayed.
-    @State private var sensorText: String = ""
-    /// This is the text description of the horizontal accuracy data that is displayed.
-    @State private var accuracyText: String = ""
-    
+
```
philium, line 24: Notice this change? It indicates that the whitespace on the line changed. When reviewing your code before committing (which you should do every time!), pay careful attention to whitespace-only changes. In this case, the whitespace was correct before and now it is not. This is something to catch and fix before committing. Please revert.

philium (approved):
Looks good! Thank you.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
chriswebb09: Okay, I think it is fixed

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,154 @@
...
+            // Start the location display to zoom to the user's current location.
+            try await locationDisplay.dataSource.start()
+        }
+        
+        /// Creates an indoor location datasource from the maps tables if there is no indoors definition.
+        /// - Parameter map: The map which contains the tables from which the data source is constructed.
+        /// - Returns: Returns a configured IndoorsLocationDataSource created from the IPS position table.
+        private func createIndoorLocationDataSourceFromTables(map: Map) async throws -> IndoorsLocationDataSource? {
+            // Gets the positioning table from the map.
+            guard let positioningTable = map.tables.first(where: { $0.displayName == "IPS_Positioning" }) else { return nil }
```
yo1995, line 88: Should use `tableName` instead of `displayName`

yo1995 (commented):
Some stuff I'm gonna address

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,154 @@
...
+            isLoading = true
+            defer { isLoading = false }
+            try await map.load()
+            try await setIndoorDatasource()
+        }
+        
+        /// Sets the indoor datasource on the location display depending on
+        /// whether the map contains an indoor definition.
+        private func setIndoorDatasource() async throws {
+            try await map.floorManager?.load()
```
yo1995, line 61: After further examination, this web map does not contain a floor manager. 🤦

yo1995: So I enabled floor-awareness of the webmap, and some hacks can be removed.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,103 @@
...
+                if model.isLoading {
+                    ProgressView(
+                        """
+                        Loading
+                        indoor data…
+                        """
+                    )
+                    .padding()
+                    .background(.ultraThinMaterial)
+                    .cornerRadius(10)
```
CalebRas, line 73: `cornerRadius(_:antialiased:)` is deprecated in iOS 18, but I am okay if you want to leave as is for now.

CalebRas (approved):
Looks good!

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
yo1995: Will fix.
Can you log an issue in this repo so we can clean them up later? Thanks.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.Model.swift
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+private extension PortalItem.ID {
+    /// Esri campus Building L IPS data.
+    static var indoorsMap: Self { Self("8fa941613b4b4b2b8a34ad4cdc3e4bba")! }
+}
+
+private extension ShowDeviceLocationUsingIndoorPositioningView.Model {
+    enum SetupError: LocalizedError {
```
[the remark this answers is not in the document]
yo1995: No, there were more errors I tried to throw but reduced them to only one. Will change it to a struct.

Shared/Samples/Show device location using indoor positioning/ShowDeviceLocationUsingIndoorPositioningView.swift
```diff
@@ -0,0 +1,103 @@
...
+                        }
+                    }
+                    .opacity(model.currentFloor != nil ? 1 : 0)
+                    VStack(alignment: .leading) {
+                        let sourceType = model.positionSource == "GNSS" ? "Satellites" : "Transmitters"
+                        Text("Data source: \(model.positionSource ?? "None")")
+                        Text("Number of \(sourceType): \(model.signalSourceCount ?? 0)")
+                    }
+                }
+                .frame(maxWidth: .infinity)
```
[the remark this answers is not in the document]
yo1995: I'll skip recapturing a screenshot this time as the change is subtle <sub>(and I'm not in building M to live test it) 😬  </sub>

CalebRas (approved):

philium (approved):

chriswebb09:
😁

yo1995:
Thank you all for the feedback. Chris, great job for pulling all the pieces together!

chriswebb09:
Thank you guys for getting it across the finish line!

PR #476, opened 2024-07-24
[Fix] Editing bug in `Create and edit geometries`

des12437:
## Description

This PR fixes a bug in the `Create and edit geometries` sample introduced in #452 by setting the selected graphic to `nil` when editing is cancelled. Otherwise the model still has the selected graphic set which causes unexpected behavior when adding a new graphic, resulting in the previously selected graphic to disappear.


yo1995 (approved):

dfeinzimer (approved):

des12437:
Thank you both.

PR #491, opened 2024-08-07
[Update] Sample type and member access levels

### LINKED ISSUE #375: Examine access levels for all sample types
Currently some types aren't correctly limited to the least access level. An example

https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/05b3a37f1419d9bedc98fe5a82a62635e63a2a4e/Shared/Samples/Create%20and%20edit%20geometries/CreateAndEditGeometriesView.swift#L227

The data model type can be `private`, but is `internal` right now. In a very rare case, a type might be reused by another sample, which causes inconvenience to fix.

We can go through each sample and  make them correct.

CalebRas:
## Description

This PR updates the access levels of types and members used by samples to make them more encapsulated. This was done to: 
- Prevent types and members being reused in other samples. (Samples should be stand alone to allow users to copy and paste them, see related [comment](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/372#discussion_r1567718805).)
- Prevent existing declarations interfering with new ones. (E.g., the internal `Model` used by `Edit geodatabase with transactions` was preventing any new samples from implementing their own file-private `Model`.)

Overview of changes (corresponding to a commit):
- Make types file-private or nested (or both).
- Make extensions private were possible.
- Make sample view members (properties, methods, & nested types) private were possible.

## Linked Issue(s)

Closes #375.

## How To Test

There are no functionality changes in this PR so ensuring the project builds should be sufficient.

### DIFF
--- a/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
+++ b/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
@@ -205,13 +205,16 @@ extension DownloadPreplannedMapAreaView {
     }
 }
 
-extension DownloadPreplannedMapAreaView.OfflineMapModel: Hashable {
-    typealias OfflineMapModel = DownloadPreplannedMapAreaView.OfflineMapModel
-    
-    static func == (lhs: OfflineMapModel, rhs: OfflineMapModel) -> Bool {
+extension DownloadPreplannedMapAreaView.OfflineMapModel: Equatable {
+    static func == (
+        lhs: DownloadPreplannedMapAreaView.OfflineMapModel,
+        rhs: DownloadPreplannedMapAreaView.OfflineMapModel
+    ) -> Bool {
         lhs === rhs
     }
-    
+}
+
+extension DownloadPreplannedMapAreaView.OfflineMapModel: Hashable {
     func hash(into hasher: inout Hasher) {
         hasher.combine(ObjectIdentifier(self))
     }
--- a/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
+++ b/Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
@@ -155,7 +155,7 @@ extension DownloadPreplannedMapAreaView {
     }
 }
 
-extension DownloadPreplannedMapAreaView.Model {
+extension DownloadPreplannedMapAreaView {
     /// A type that specifies the currently selected map.
     enum SelectedMap: Hashable {
         /// The online version of the map.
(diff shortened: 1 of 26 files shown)


des12437 (dismissed):

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
```diff
@@ -165,45 +165,49 @@ extension DownloadPreplannedMapAreaView.Model {
...
+            Task { [job] in
+                // Cancel any outstanding job.
+                await job?.cancel()
+            }
         }
     }
 }
 
-extension OfflineMapModel: Hashable {
+extension DownloadPreplannedMapAreaView.OfflineMapModel: Hashable {
```
philium, line 208: Not sure if there's anything that can be done about it at this point, but the implementation of the requirements of these protocols do not fit the requirements as outlined in the documentation. Most likely `ObjectIdentifier` (which adopts `Equatable` and `Hashable`) should have been used (as evidenced by the implementation of `hash(into:)`) instead of the model itself.

Also, FYI: equatable and hashable are really two separate protocols and are typically adopted using two separate extensions. When adopting a protocol via an extension, it is helpful to only include the requirements for that protocol in that extension, and nothing else. The typealias is unnecessary. If `Self` can't be used, then just use the full type name, even if it is long. After all, that is how the compiler synthesizes it.

Shared/Samples/Show device location with NMEA data sources/FileNMEASentenceReader.swift
```diff
@@ -14,91 +14,93 @@
...
+        private(set) var isStarted = false
+        
+        /// Loads locations from NMEA sentences.
+        /// Reads mock NMEA sentences line by line and group them by the timestamp.
+        /// - Parameters:
+        ///   - nmeaSourceFile: The URL of the NMEA source file.
+        ///   - speed: The playback time interval in second.
+        init(url: URL, interval: TimeInterval = 1.0) {
+            // An empty container for NMEA data.
+            var dataBySeconds = [Data]()
```
philium, line 42: Not a big deal, but the syntactic sugar for arrays/dictionaries isn't intended to be used when initializing a type (per Apple devs on the forums).

```suggestion
            var dataBySeconds: [Data] = []
```

Shared/Samples/Show device location with NMEA data sources/FileNMEASentenceReader.swift
```diff
@@ -14,91 +14,93 @@
...
+            
+            if let nmeaStrings = try? String(contentsOf: url, encoding: .utf8).components(separatedBy: .newlines).filter({ !$0.isEmpty }) {
+                // A temporary container for the NMEA sentences at current timestamp.
+                var currentTimestamp = [String]()
```
philium, line 46:
```suggestion
                var currentTimestamp: [String] = []
```

philium (dismissed):

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
CalebRas: I agree the original implementation isn't the best, but I will leave it as is for now to avoid refactoring the sample in this PR. I did remove the typealias, though, and broke the extension into two, one per conformance. Let me know if you had something different in mind.

Shared/Samples/Show device location with NMEA data sources/FileNMEASentenceReader.swift
CalebRas: In an effort to keep the changes in this PR related to access level, I created an issue for these. That way, they can all be addressed at once. 

- #493

Shared/Samples/Download preplanned map area/DownloadPreplannedMapAreaView.Model.swift
philium: Sounds good. Thanks!

Shared/Samples/Show device location with NMEA data sources/FileNMEASentenceReader.swift
philium: Thank you!

philium (approved):

des12437 (approved):

PR #508, opened 2024-09-20
[Update] Bump version to Xcode 16

yo1995:
## Description

This PR bumps version numbers to Xcode 16. Related to https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/883

## How To Test

Build and run the app on Xcode 16 without error. Xcode 15 won't open the project.


yo1995:
When the Swift Language Version is bumped to Swift 6.0, there are a few errors and warnings that pops up. Upon discussion with Phil R, I'll address these errors in this PR.

Shared/Samples/Animate 3D graphic/Animate3DGraphicView.Model.swift
```diff
@@ -137,7 +137,7 @@ extension Animate3DGraphicView {
         }
         
         deinit {
```
yo1995, line 139: See discussion in this thread: https://forums.swift.org/t/cannot-access-property-with-a-non-sendable-type-from-non-isolated-deinit/57877

Since the 2 `deinit` related changes are just cleaning up, I decided to loosely handle them with a separate `nonisolated` method that actually runs on MainActor or the inherited context.

Shared/Samples/Create and save KML file/CreateAndSaveKMLView.swift
```diff
@@ -70,7 +70,8 @@ struct CreateAndSaveKMLView: View {
 
 extension CreateAndSaveKMLView {
     /// A KMZ file that can be used with the native file exporter.
-    final class KMZFile: FileDocument {
+    @MainActor
+    final class KMZFile: @preconcurrency FileDocument, Sendable {
```
yo1995, line 74: Not sure if this is the right fix for `FileDocument` conforming type, but it works.

Shared/Samples/Show viewshed from geoelement in scene/ShowViewshedFromGeoelementInSceneView.swift
```diff
@@ -46,7 +46,7 @@ struct ShowViewshedFromGeoelementInSceneView: View {
 
 private extension ShowViewshedFromGeoelementInSceneView {
     /// The view model for the sample.
-    class Model: ObservableObject {
+    class Model: ObservableObject, @unchecked Sendable {
```
yo1995, line 49: This is particularly bad - the timer assumes the block is Sendable, which ask for the captured model to be Sendable.

Blindly conform the model to Sendable is a bad practice 🤮 , but since the only use of it being sendable is to allow a scheduled animation, is that OK to just disregard potential data races in it? Animation isn't very important after all.

Would like to hear your thoughts on this and the `FileNMEASentenceReader` type. Should I instead log issues to refactor those code?

Shared/Supporting Files/Views/WebView.swift
```diff
@@ -45,15 +45,14 @@ struct WebView: UIViewRepresentable {
...
-                    await Self.openURL(url)
+                    Self.openURL(url)
                 }
                 return .cancel
             default:
                 return .allow
             }
         }
         
-        @MainActor
```
yo1995, line 56: #439

README.md
```diff
@@ -20,7 +20,7 @@ This repository contains Swift sample code demonstrating the capabilities of the
 
 * [ArcGIS Maps SDK for Swift](https://developers.arcgis.com/swift/) 200.5 (or newer)
```
philium, line 21: This should to be updated to 200.5.1.

```suggestion
* [ArcGIS Maps SDK for Swift](https://developers.arcgis.com/swift/) 200.5.1 (or newer)
```

Samples.xcodeproj/project.pbxproj
```diff
@@ -2960,7 +2960,6 @@
 				};
 			};
 			buildConfigurationList = 00E5400A27F3CCA100CF66D5 /* Build configuration list for PBXProject "Samples" */;
-			compatibilityVersion = "Xcode 15.0";
```
philium, line 2963: Normally Xcode would automatically update this to "Xcode 16.0". Did that not happen?

Shared/Samples/Animate 3D graphic/Animate3DGraphicView.Model.swift
philium: A better solution would be to move starting and stopping the animation from init and deinit to methods that the view could call on appear and disappear. That more closely ties lifetime of the animation to that of the view itself and eliminates the need for a task to stop it. Same elsewhere.

Shared/Samples/Create and save KML file/CreateAndSaveKMLView.swift
philium: No, this is not the right approach because `FileDocument` has adopted concurrency (as evidenced by the Sendable conformance and preconcurrency attribute). Instead, work through the issues. Some can be addressed by adding the `nonisolated` keyword, but others may be more involved. Same with `CreateMobileGeodatabaseView.GeodatabaseFile`. Let me know if you'd like some help!

Shared/Samples/Show viewshed from geoelement in scene/ShowViewshedFromGeoelementInSceneView.swift
philium: There isn't a simple solution here. I would suggest trying another method (instead of the block-based one).

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
```diff
@@ -15,7 +15,7 @@
 import Foundation
 
 /// An extension allowing an array to be used with the app storage property wrapper.
-extension Array: RawRepresentable where Element == String {
+extension Array: @retroactive RawRepresentable where Element == String {
```
philium, line 18: We really shouldn't be doing this. Could the adoption of the protocol itself be removed?

```swift
extension Array<String> {
    ...
}
```

Shared/Supporting Files/Views/WebView.swift
```diff
@@ -45,15 +45,14 @@ struct WebView: UIViewRepresentable {
             switch navigationAction.navigationType {
             case .linkActivated:
                 if let url = navigationAction.request.url {
-                    await Self.openURL(url)
+                    Self.openURL(url)
```
philium, line 48: Can the body of `openURL(_:)` just be inlined here now? If I remember right, this method was added purely to address the issue of needing to call the `UIApplication` APIs on the main thread.

Samples.xcodeproj/project.pbxproj
yo1995: I didn't receive the "Update to suggested project setting" warning on the first launch. So instead, I triggered one manually with Eskimo's [answer](https://forums.developer.apple.com/forums/thread/743932).

philium: This isn't related to that. Xcode should change this when the project format is changed:

image

yo1995: I didn't see that happen. Just now I tried to select 15.0, and there isn't a change in the project file after I select back to 16.0.

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
yo1995: This is used for #313 `FavoritesView`'s `AppStorage` to serialized an array of strings. If we don't add this extension, then we might need to roll our own UserDefault managing types instead. (which I tried by then and is possible, but involves manually insert/remove from user defaults so it is less clean than this solution)

Shared/Samples/Show viewshed from geoelement in scene/ShowViewshedFromGeoelementInSceneView.swift
yo1995: I've made the following changes:

- For the NMEA sentence reader, I move it onto Main Actor so the timer executes the block on main thread.
- For the Viewshed tank animation, I move it to the view level.

Let me know your thoughts.

Shared/Samples/Show device location with NMEA data sources/FileNMEASentenceReader.swift
```diff
@@ -93,10 +96,6 @@ extension ShowDeviceLocationWithNMEADataSourcesView {
             timer = nil
             messages = nil
         }
-        
-        deinit {
-            stop()
```
yo1995, line 98: This isn't needed because when the sample's main view disappears, it runs `reset` method that also stops the sentence generation.

Samples.xcodeproj/project.pbxproj
philium: Fascinating. Something must have changed in Xcode 16.

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
philium: I'm not suggesting getting rid of the functionality. It could still be used by changing the app storage type to `String` and then adding a computed property:

```swift
@AppStorage(.favoriteSampleNames) private var favoriteNamesString: String = ""

private var favoriteNames: [String] {
    get { .init(rawValue: favoriteNamesString) }
    nonmutating set { favoriteNamesString = newValue.rawValue }
}
```

I also suggest renaming `rawValue` to `jsonString`.

Shared/Samples/Create mobile geodatabase/CreateMobileGeodatabaseView.Model.swift
```diff
@@ -117,7 +117,8 @@ extension CreateMobileGeodatabaseView {
     // MARK: GeodatabaseFile
     
     /// A geodatabase file that can be used with the native file exporter.
-    final class GeodatabaseFile {
+    @MainActor
+    final class GeodatabaseFile: Sendable {
```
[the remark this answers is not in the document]
philium: It looks like all you need to do is mark the protocol requirements as non-isolated:

```swift
extension CreateMobileGeodatabaseView.GeodatabaseFile: FileDocument {
    /// The file and data types that the document reads from.
    nonisolated static var readableContentTypes: [UTType] { [.geodatabase] }
    
    /// Creates a document and initializes it with the contents of a file.
    nonisolated convenience init(configuration: ReadConfiguration) throws {
        fatalError("Loading geodatabase files is not supported by this sample.")
    }
    
    /// Serializes a document snapshot to a file wrapper.
    nonisolated func fileWrapper(configuration: WriteConfiguration) throws -> FileWrapper {
        return try FileWrapper(url: geodatabaseURL)
    }
}
```

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
yo1995: Thanks for the new perspective 👍  Haven't thought we could just store a string instead of an array.

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
```diff
@@ -15,10 +15,10 @@
 import Foundation
 
 /// An extension allowing an array to be used with the app storage property wrapper.
-extension Array: RawRepresentable where Element == String {
+extension Array where Element == String {
```
philium, line 18: This could be simplified:

```suggestion
extension Array<String> {
```

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
```diff
@@ -15,10 +15,10 @@
...
     /// Creates a new array from a given raw value.
     /// - Parameter rawValue: The raw value of the array to create.
-    public init(rawValue: String) {
+    init(rawValue: String) {
```
philium, line 21: This parameter should be renamed to `jsonString` as well. Also, please update doc.

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
philium, line 1: Please rename this file.

Shared/Supporting Files/Extensions/Array+RawRepresentable.swift
```diff
@@ -28,7 +28,7 @@ extension Array: RawRepresentable where Element == String {
     }
     
     /// The raw value of the array.
```
philium, line 30: Please update doc.

philium (commented):
I'm seeing a build failure here:

https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/1d207ee02238a6f00d2cf4754373152dc8d80654/Shared/Supporting%20Files/Views/SampleLink.swift#L64

philium (approved):

des12437 (approved):

PR #510, opened 2024-09-20
Revert "[Update] Fix linter warnings for string <-> data conversion"

CalebRas:
Reverts Esri/arcgis-maps-sdk-swift-samples#505

### DIFF
--- a/Shared/Samples/Navigate route with rerouting/NavigateRouteWithReroutingView.Model.swift
+++ b/Shared/Samples/Navigate route with rerouting/NavigateRouteWithReroutingView.Model.swift
@@ -122,7 +122,7 @@ extension NavigateRouteWithReroutingView {
             
             // Set up the data source's locations using a local JSON file.
             let jsonData = try Data(contentsOf: .sanDiegoTourPath)
-            let jsonString = String(decoding: jsonData, as: UTF8.self)
+            guard let jsonString = String(data: jsonData, encoding: .utf8) else { return }
             let routePolyline = try Polyline.fromJSON(jsonString)
             simulatedDataSource.setSimulatedLocations(with: routePolyline)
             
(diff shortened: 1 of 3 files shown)


yo1995 (approved):

philium (commented):
Why are these changes being reverted? I understand that listing rule has changed, but why does the code need to be changed back? In at least some of the cases, if not all of them, we know that the data will always be encoded as UTF8.

CalebRas:
> Why are these changes being reverted? I understand that listing rule has changed, but why does the code need to be changed back? In at least some of the cases, if not all of them, we know that the data will always be encoded as UTF8.

I see what your saying. Are you suggesting I use `swiftlint:disable` instead? 

Ting asked me to revert these and seems to be more knowledgeable about this than me. Thoughts @yo1995?

Shared/Samples/Navigate route with rerouting/NavigateRouteWithReroutingView.Model.swift
```diff
@@ -122,6 +122,7 @@ extension NavigateRouteWithReroutingView {
             
             // Set up the data source's locations using a local JSON file.
             let jsonData = try Data(contentsOf: .sanDiegoTourPath)
+            // swiftlint:disable:next optional_data_string_conversion
             let jsonString = String(decoding: jsonData, as: UTF8.self)
```
philium, line 126: Here, just use `String.init(contentsOf:encoding:)`:

```suggestion
            let jsonString = try String(contentsOf: .sanDiegoTourPath, encoding: .utf8)
```

philium (approved):

yo1995 (approved):

CalebRas:
Thank you both!

PR #539, opened 2024-11-01
[Fix] Samples unavailable on Mac Catalyst

dfeinzimer:
Toolkit AR components are no longer available on Catalyst

PR #541, opened 2024-11-05
[Fix] Samples unavailable on Mac Catalyst

dfeinzimer:
Re-do of #539 which was closed to apply these changes with a cleaner history.


Scripts/GenerateSampleViewSourceCode.swift
```diff
@@ -123,6 +122,7 @@ private let sampleStructs = sampleMetadata
         let portalItemIDs = (sample.offlineData ?? [])
             .map { #"PortalItem.ID("\#($0)")!"# }
         return """
+        \(sample.category == "Augmented Reality" ? "@available(macCatalyst, unavailable)" : "")
```
CalebRas, line 125: I am wondering if it would be simpler to generate a compiler directive here (and one at the bottom) instead of the `@available`:

```swift
#if !targetEnvironment(macCatalyst) || ! targetEnvironment(simulator)
```

That way, you wouldn't have to do any filtering like you did below, and it would be more efficient since it wouldn't get complied. You could also then use the "Target Membership" for the sample view files to exclude them on mac catalyst, so those wouldn't get complied either. Thoughts?

Shared/Samples/Augment reality to navigate route/AugmentRealityToNavigateRouteView.swift
```diff
@@ -16,6 +16,8 @@ import ArcGIS
 import CoreLocation
 import SwiftUI
 
+@available(macCatalyst, unavailable)
+@MainActor
```
CalebRas, line 20:
```suggestion
```

Is this needed? I believe conforming to `View` already puts this on the main actor.

Scripts/GenerateSampleViewSourceCode.swift
yo1995: This is an interesting solution. It would also need an `#endif` to wrap the block. It would generate one directive for each excluded struct (which looks kind of busy, but we won't see the generated code for structs after all, so no worries). I think this idea is worth a try.

Shared/Samples/Augment reality to navigate route/AugmentRealityToNavigateRouteView.swift
dfeinzimer: Yes, this was accidentally included when I resolved the rebase merge conflict in #539

Scripts/GenerateSampleViewSourceCode.swift
dfeinzimer: Prototyping this right now and unless I'm missing something we'd still need the `commonSamples` and `iOSSamples` arrays meaning we'd have two layers of the `#if !targetEnvironment...` directives: the ones you're suggesting on the structs and the one on `iOSSamples`. I'm not too keen on this.

As for changing the Target Membership on relevant files, I think that's a good idea so I've applied that suggestion.

CalebRas: I am not sure I see the need for `commonSamples` and `iOSSamples` if the we could have just one `samples` array where the AR samples are filtered out using the compiler directives. I'll trust you if you say they are necessary for some other reason, though.

Also, this currently doesn't compile on Mac Catalyst for me. Changing the target membership without the suggested compiler directives makes it so the sample views required for the AR `Sample` struct can't be found.

dfeinzimer: The reason is that compiler directives don't work in the middle of the array definitions. 

<p align="center">
  
</p>

The other alternative would be to have a computed array, and that's what we originally started with [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/539#discussion_r1826346050), but moved on from because all of the common samples are listed twice.

> Changing the target membership without the suggested compiler directives makes it so the sample views required for the AR Sample struct can't be found.

Good find thank you. I've reverted the Target Membership change.

yo1995: Ha, I think I was under the same wrong impression that the [compiler directives](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Conditional-Compilation-Block) are part of preprocessing. It turned out it is evaluated at compile time 🤔 rather than like in C-like languages, these directives are part of preprocessing…

yo1995 (dismissed):

CalebRas (approved):
Builds on all platforms.

philium (commented):
I know I wasn't asked to review this PR, but these are a couple things I noticed as I followed the discussion.

philium (approved):

dfeinzimer:
Screenshot 2024-11-06 at 2 08 35 PM

@philium this was accidental please disregard.

yo1995 (approved):

CalebRas (approved):

PR #548, opened 2024-11-14
[Fix] Search bugs

CalebRas:
## Description

This PR fixes some issues related to searching in the samples:

- A bug where some text wouldn't be bolded correctly due to a markdown issue (see screenshots).
- A bug where autocorrect would change the search results would change when un-focusing a search bar. Disabled autocorrect appears to be the default behavior for `.searchable` in iOS 18+, so this fix is for iOS 16 & 17.

## How To Test

- Verify the search bars perform as expected.

## Screenshots

|Before|After|
|:-:|:-:|
|    Simulator Screenshot - 1 - iPhone 16 Pro - 2024-11-13 at 16 38 47    |    Simulator Screenshot - 1 - iPhone 16 Pro - 2024-11-13 at 16 38 10    |

### DIFF
--- a/Shared/Supporting Files/Views/SampleLink.swift
+++ b/Shared/Supporting Files/Views/SampleLink.swift
@@ -131,11 +131,18 @@ private extension String {
     /// - Parameter substring: The substring to bold.
     /// - Returns: The attributed string with the bolded substring.
     func boldingFirstOccurrence(of substring: String) -> AttributedString {
-        if let range = localizedStandardRange(of: substring.trimmingCharacters(in: .whitespacesAndNewlines)),
-           let boldedString = try? AttributedString(markdown: replacingCharacters(in: range, with: "**\(self[range])**")) {
-            return boldedString
-        } else {
-            return AttributedString(self)
+        var attributedString = AttributedString(self)
+        
+        let trimmedSubstring = substring.trimmingCharacters(in: .whitespacesAndNewlines)
+        if let range = localizedStandardRange(of: trimmedSubstring) {
+            let substring = self[range]
+            
+            if let boldedSubstring = try? AttributedString(markdown: "**\(substring)**"),
+               let attributedRange = attributedString.range(of: substring) {
+                attributedString.replaceSubrange(attributedRange, with: boldedSubstring)
+            }
         }
+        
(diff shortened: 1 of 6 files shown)


Shared/Supporting Files/Views/SampleLink.swift
```diff
@@ -131,11 +131,18 @@ private extension String {
...
-        } else {
-            return AttributedString(self)
+        var attributedString = AttributedString(self)
+        
+        let trimmedSubstring = substring.trimmingCharacters(in: .whitespacesAndNewlines)
+        if let range = localizedStandardRange(of: trimmedSubstring) {
+            let substring = self[range]
+            
+            if let boldedSubstring = try? AttributedString(markdown: "**\(substring)**"),
+               let attributedRange = attributedString.range(of: substring) {
```
yo1995, line 141:
```suggestion
            if let boldedSubstring = try? AttributedString(markdown: "**\(substring)**"),
               let attributedRange = attributedString.range(of: self[range]) {
```
Since the local variable `substring` is only used here, move it here can avoid the potential confusion with the method argument.

yo1995, line 141: If the suggestion above is applied, then the 3 if-lets can be combined to one place

```swift
        if let range = localizedStandardRange(of: trimmedSubstring),
           let boldedSubstring = try? AttributedString(markdown: "**\(substring)**"),
           let attributedRange = attributedString.range(of: self[range]) {
            attributedString.replaceSubrange(attributedRange, with: boldedSubstring)
        }
```

CalebRas: It's actually needed in `AttributedString(markdown: "**\(substring)**")` as well, but I replaced both uses of it so I could apply your following suggestion.

yo1995 (approved):

Shared/Supporting Files/Views/SampleLink.swift
yo1995: Oops, apparently I misinterpreted which variable is which as well 😓

dfeinzimer (approved):

CalebRas:
Thank you!

PR #550, opened 2024-11-15
[Update] Add sample action menu

### LINKED ISSUE #407: Reconsider favorite category tile displaying condition
A few people left feedback on that an empty "Favorites" category tile may be confusing to first time users.

- Phil: https://github.com/Esri/arcgis-runtime-samples-ios/pull/1159#discussion_r866243170
- Lucas: When the app opens, the first thing I see to click is the Favorites category, so I clicked on that. When that opens, it’s just an empty list. I was thinking that maybe there would already be some default favorites in there, so I was a bit confused. I think favoriting a sample might be a “power user” type of workflow, so making that the very first item in the grid might be confusing for a new user that is exploring the app for the first time, as they will very likely just click the first thing they see. So, there are a few things you could consider:
  - don’t make favorites the first item
  - add some defaulted favorites
  - add some brief instructional text in the empty favorite list view that let’s the user know how to add favorites

My take is that the favorite tile will most likely be used by power user and internal colleagues, we can hide the tile when there is no favorited samples.

CalebRas:
## Description

This PR updates the sample row context menu and adds a new sample toolbar menu with following actions: 
- `View Info` (This replaces the sample info button to save space. Also, this is only present in the toolbar menu since it was much simipler to do it that way.)
- `View on Esri Developer`
- `View on GitHub` (This replaces the Mac Catalyst toolbar button.) 
- `Favorite`/`Unfavorite`

## Linked Issue(s)

Closes #407

## How To Test

- Ensure all of the menu actions work in the context menu (long press on a sample row) and the sample toolbar menu (circle ellipse button).

## Screenshots

|Before|After|
|:-:|:-:|
|    Simulator Screenshot - 1 - iPhone 16 Pro - 2024-11-15 at 12 27 26    |    Simulator Screenshot - 1 - iPhone 16 Pro - 2024-11-15 at 12 26 25    |
|    Simulator Screenshot - 1 - iPhone 16 Pro - 2024-11-15 at 12 23 56    |     Simulator Screenshot - 1 - iPhone 16 Pro - 2024-11-15 at 12 26 38    |

### DIFF
--- a/Shared/Supporting Files/Views/SampleMenuButtons.swift
+++ b/Shared/Supporting Files/Views/SampleMenuButtons.swift
@@ -36,7 +36,7 @@ struct SampleMenuButtons: View {
         Divider()
         Button {
             if sampleIsFavorite {
-                favoriteNames.removeAll { $0 == sample.name }
+                favoriteNames.removeAll(where: { $0 == sample.name })
             } else {
                 favoriteNames.append(sample.name)
             }
--- a/Shared/Supporting Files/Views/SampleMenuButtons.swift
+++ b/Shared/Supporting Files/Views/SampleMenuButtons.swift
@@ -1,0 +1,48 @@
+// Copyright 2024 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+
+struct SampleMenuButtons: View {
+    /// The sample to show the menu buttons for.
+    let sample: Sample
+    
+    /// The names of the favorite samples loaded from user defaults.
+    @AppFavorites private var favoriteNames
+    
+    /// A Boolean value indicating whether the sample is a favorite.
+    private var sampleIsFavorite: Bool {
+        favoriteNames.contains(sample.name)
+    }
+    
+    var body: some View {
+        Link(destination: sample.esriDeveloperURL) {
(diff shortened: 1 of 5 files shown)


pgruenler (dismissed):

philium:
Do we need the long press on the sample row? That seems redundant with the new menu inside the sample. What would be helpful on the row itself would be a swipe action to favorite/unfavorite the sample.

CalebRas:
> Do we need the long press on the sample row? That seems redundant with the new menu inside the sample.

I was modeling Apple Music here as it has a favorite/unfavorite both in the context menu and on the toolbar. From my understanding, context menus usually contain redundant items since the [HIG](https://developer.apple.com/design/human-interface-guidelines/context-menus) says: 

> Always make context menu items available in the main interface, too.

But I will remove it if you still think it is unneeded since I agree that it is not an essential feature.

> What would be helpful on the row itself would be a swipe action to favorite/unfavorite the sample.

I think this would be confusing. It seems to me that swiping is usually associated with destructive behavior (e.g., deleting in Files, Messages, Safari, etc.). I personally would only figure out that swiping favorites by accident.

philium:
I wonder if Apple did that for the Music app because there are so many menu options? Either way is fine I guess.

Shared/Supporting Files/Views/SampleMenuButtons.swift
```diff
@@ -0,0 +1,48 @@
...
+        Link(destination: sample.esriDeveloperURL) {
+            Label("View on Esri Developer", systemImage: "link")
+        }
+        Link(destination: sample.gitHubURL) {
+            Label("View on GitHub", systemImage: "link")
+        }
+        Divider()
+        Button {
+            if sampleIsFavorite {
+                favoriteNames.removeAll { $0 == sample.name }
```
philium, line 39: When a method takes a single unmuted (with an argument label) parameter, like this one, it is best to include the label for the sake of clarity:

```suggestion
                favoriteNames.removeAll(where: { $0 == sample.name })
```

philium (dismissed):

philium (approved):

pgruenler (approved):

CalebRas:
Thank you both!

PR #552, opened 2024-11-19
[Fix] `Browse OGC API feature service` alert bug

CalebRas:
## Description

This PR fixes a bug in `Browse OGC API feature service` where the alert wouldn't present when the sample was opened on iOS 18 (only iPhone), even when "Open Service" was pressed. I also changed the spelling of "textfield" to "text field" since it is [two words in SwiftUI](https://developer.apple.com/documentation/swiftui/textfield).

## How To Test

- Ensure the alert is presented when the sample is first opened on iOS 18.

### DIFF
--- a/Shared/Samples/Browse OGC API feature service/BrowseOGCAPIFeatureServiceView.swift
+++ b/Shared/Samples/Browse OGC API feature service/BrowseOGCAPIFeatureServiceView.swift
@@ -19,8 +19,8 @@ struct BrowseOGCAPIFeatureServiceView: View {
     /// The error shown in the error alert.
     @State private var error: Error?
     
-    /// A Boolean value indicating whether the textfield alert should be presented.
-    @State private var textfieldAlertIsPresented = false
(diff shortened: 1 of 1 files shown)


Shared/Samples/Browse OGC API feature service/BrowseOGCAPIFeatureServiceView.swift
```diff
@@ -63,8 +63,9 @@ struct BrowseOGCAPIFeatureServiceView: View {
                         }
                     }
                 }
-                .alert("Load OGC API feature service", isPresented: $textfieldAlertIsPresented) {
-                    // Textfield has a default OGC API URL.
+                .alert("Load OGC API feature service", isPresented: $textFieldAlertIsPresented) {
+                    // Text
+                    field has a default OGC API URL.
```
yo1995, line 68:
```suggestion
                    // Text field has a default OGC API URL.
```

yo1995 (approved):

njarecha (approved):
Looks good

PR #560, opened 2025-02-27
[New] Create KML multi-track

CalebRas:
## Description

This PR implements `Create KML multi-track` in `Edit and Manage Data` category.
URL to README: [Create KML multi-track](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-CreateKMLMultiTrackView/Shared/Samples/Create%20KML%20multi-track)

## Linked Issue(s)

- `swift/issues/6554`

## How To Test

- Tap "Record Track" to record track elements.
- Tap "Stop Recording" to convert the track elements to a track. 
- Tap the save button to save the track to a KML multi-track in a local KMZ file.
- Tap the delete button to remove the file and reset the sample.

## Screenshots

create-kml-multi-track


Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,185 @@
...
+                        Spacer()
+                        
+                        Button(asyncAction == .recordTrack ? "Stop Recording" : "Record Track") {
+                            if asyncAction == .recordTrack {
+                                asyncAction = nil
+                                model.addTrack()
+                            } else {
+                                asyncAction = .recordTrack
+                            }
+                        }
```
philium, line 76: For a toggle action like this, consider using a toggle button for an added visual cue:

```swift
Toggle(
    asyncAction == .recordTrack ? "Stop Recording" : "Record Track",
    isOn: .init {
        asyncAction == .recordTrack
    } set: { newValue in
        asyncAction = newValue ? .recordTrack : nil
    }
)
```

Screen Recording 2025-02-28 at 11 03 39 AM

However, you'll need to call `addTrack()` some other way. You could do that in the task modifier as a result of the state change.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,185 @@
...
+}
+
+private extension CreateKMLMultiTrackView {
+    /// A picker for selecting a track from a KML multi-track.
+    struct TrackPicker: View {
+        /// The KML tracks options shown in the picker.
+        private let tracks: [KMLTrack]
+        
+        /// The proxy for setting map view's viewpoint to the selected track's geometry.
+        private let mapViewProxy: MapViewProxy
```
philium, line 143: I suggest making these non-private:

```swift
/// The KML tracks options shown in the picker.
let tracks: [KMLTrack]
/// The proxy for setting map view's viewpoint to the selected track's geometry.
let mapViewProxy: MapViewProxy
```

And then just using the compiler synthesized initializer:

```swift
TrackPicker(tracks: multiTrack.tracks, mapViewProxy: mapViewProxy)
```

philium, line 143: Giving this view the whole map view proxy seems overkill. It might have been better to have it take a closure that is called when selection changes, then the caller could provide a closure that uses the map view proxy. This wouldn't even need to know about it.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,185 @@
...
+        
+        var body: some View {
+            Picker("Track", selection: $selectedTrack) {
+                Text("All Tracks")
+                    .tag(nil as KMLTrack?)
+                
+                ForEach(Array(tracks.enumerated()), id: \.offset) { offset, track in
+                    Text("KML Track #\(offset + 1)")
+                        .tag(track)
+                }
```
philium, line 161: This is fine for a list like this that doesn't change, but you wouldn't want to use indices/offsets for a mutable list.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+    
+    /// An asynchronous action associated with the sample.
+    enum AsyncAction {
+        case startNavigation
+        case recordTrack
+        case saveKMLMultiTrack
+        case reset
+    }
```
philium, line 180: It is odd to have this all the way down here when it is used by a state property of the view. It would be better to have this closer to where it is used, usually right before the stored property that uses it.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,185 @@
...
+
+import ArcGIS
+import SwiftUI
+
+struct CreateKMLMultiTrackView: View {
+    /// The view model for the sample.
+    @StateObject private var model = Model()
+    
+    /// The asynchronous action currently being run.
+    @State private var asyncAction: AsyncAction? = .startNavigation
```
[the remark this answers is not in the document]
CalebRas: Using `State` conflicted with the `@State` wrapper, so I used `SampleState` instead.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,185 @@
...
+    
+    /// The text shown in the status bar. This describes the current state of the sample.
+    @State private var statusText = ""
+    
+    /// The KML multi-track loaded from the KMZ file.
+    @State private var multiTrack: KMLMultiTrack?
+    
+    /// A Boolean value indicating whether the recenter button is enabled.
+    @State private var recenterIsEnabled = false
```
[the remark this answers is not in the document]
pgruenler: I think `recenterIsEnabled` is a better assertion than `isRecenterEnabled`. Not a bit deal though.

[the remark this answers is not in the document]
philium: Perhaps you're right. I'm just stuck on prepending `is` to Boolean properties, like we did in the Objective-C days for the their getters.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.Model.swift
```diff
@@ -0,0 +1,158 @@
...
+    @MainActor
+    final class Model: ObservableObject {
+        /// A map with a streets basemap.
+        let map = Map(basemapStyle: .arcGISStreets)
+        
+        /// The map view's location display.
+        let locationDisplay = LocationDisplay()
+        
+        /// The graphics overlays to display on the map view.
+        let graphicsOverlays: [GraphicsOverlay]
```
philium, line 29: This could be computed instead of stored.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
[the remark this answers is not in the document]
CalebRas: Starting with "is" seems to be more commonly used, such as in the Discussion in [sheet(ispresented:ondismiss:content:)](https://developer.apple.com/documentation/swiftui/view/sheet(ispresented:ondismiss:content:)), so I will leave it as is.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.Model.swift
```diff
@@ -0,0 +1,158 @@
...
+    final class Model: ObservableObject {
+        /// A map with a streets basemap.
+        let map = Map(basemapStyle: .arcGISStreets)
+        
+        /// The map view's location display.
+        let locationDisplay = LocationDisplay()
+        
+        /// The graphics overlays to display on the map view.
+        var graphicsOverlays: [GraphicsOverlay] {
+            return [trackElementGraphicsOverlay, trackGraphicsOverlay]
```
philium, line 30: Not a big deal, but we typically omit the `return` keyword for single line property getters.

```suggestion
            [trackElementGraphicsOverlay, trackGraphicsOverlay]
```

philium (dismissed):

pgruenler (dismissed):

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.swift
```diff
@@ -0,0 +1,186 @@
...
+}
+
+private extension CreateKMLMultiTrackView {
+    /// A picker for selecting a track from a KML multi-track.
+    struct TrackPicker: View {
+        /// The KML tracks options shown in the picker.
+        let tracks: [KMLTrack]
+        
+        /// The closure to perform when the selected track has changed.
+        let onSelectionChanged: (Geometry) async -> Void
```
yo1995, line 157: This closure doesn't need to be `async`, as the "onValueChanged" semantics usually happen synchronously. Since this is an internal component, no need to change.

Shared/Samples/Create KML multi-track/CreateKMLMultiTrackView.Model.swift
yo1995: Screenshot 2025-03-06 at 4 14 02 PM

With `return` removed, I'm seeing this warning. Either suppress the rule or add back `return`.

yo1995 (dismissed):

yo1995 (approved):

CalebRas:
@pgruenler Could you take another look at this when you get a second please?

philium (approved):

CalebRas:
Thank you!

PR #564, opened 2025-03-28
[New] Snap geometry edits with utility network rules

CalebRas:
## Description

This PR implements `Snap geometry edits with utility network rules` in `Utility Networks` category.
URL to README: [Snap geometry edits with utility network rules](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-SnapGeometryEditsWithUtilityNetworkRules/Shared/Samples/Snap%20geometry%20edits%20with%20utility%20network%20rules)

## Linked Issue(s)

- `swift/issues/6640`

## How To Test

- Tap on a feature to select it and press the edit button to start the geometry editor.
- Tap "Snap Sources" to enable and disable the snap sources.
- Drag to snap the vertex tool to a feature or graphic, and tap to place the feature.
- Press the save button to save the edits or the cancel button to discard them.
- Note that this requires `swift-toolkit-daily` to build. (There are some warning when building with this. I am fixing them in another PR.)

## Screenshots

snap-geometry-edits-with-utility-network-rules


Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,280 @@
...
+                        ForEach(Array(settings.enumerated()), id: \.offset) { _, settings in
+                            SnapSourceSettingsToggle(
+                                settings: settings,
+                                image: ruleBehaviorImages[settings.ruleBehavior]
+                            )
+                        }
+                    }
+                    
+                    Section("Legend") {
+                        ForEach(SnapRuleBehavior?.allCases, id: \.self) { behavior in
```
CalebRas, line 185: I originally had this legend on the main screen, similar to the design. It seemed too cluttered on iOS, though, so I moved it to the sheet.

Simulator Screenshot - 1 - iPhone 16 Pro - 2025-03-27 at 18 29 00

philium: I actually think it looks better overlaying the map. It actually takes up less room than the sheet does! However, I think the overlay could be made even smaller. Here' one idea:

image

<details>

<summary>Code</summary>

```swift
// In SnapGeometryEditsWithUtilityNetworkRulesView.body:

.overlay(alignment: .bottom) {
    SnappingLegend()
        .frame(maxWidth: .infinity, alignment: .center)
        .padding(8)
}

// Somewhere below SnapGeometryEditsWithUtilityNetworkRulesView:

private struct SnappingLegend: View {
    /// The display scale of this environment.
    @Environment(\.displayScale) private var displayScale
    
    /// The images for the snap rule behavior legend.
    @State private var ruleBehaviorImages: [SnapRuleBehavior?: Image] = [:]
    
    var body: some View {
        Group {
            if ruleBehaviorImages.count == SnapRuleBehavior?.allCases.count {
                LabeledContent("Snapping") {
                    HStack {
                        ForEach(SnapRuleBehavior?.allCases, id: \.self) { behavior in
                            Label {
                                Text(behavior.label)
                            } icon: {
                                ruleBehaviorImages[behavior]!
                            }
                        }
                    }
                }
                .font(.footnote)
            } else {
                EmptyView()
            }
        }
        .task(id: displayScale) {
            ruleBehaviorImages.removeAll(keepingCapacity: true)
            for behavior in SnapRuleBehavior?.allCases {
                // Creates an image from the rule behavior's symbol.
                let swatch = try? await behavior.symbol.makeSwatch(scale: displayScale)
                ruleBehaviorImages[behavior] = swatch.map(Image.init(uiImage:))
            }
        }
    }
}
```

</detail>

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,280 @@
...
+                }
+                .errorAlert(presentingError: $error)
+        }
+        .overlay(alignment: .top) {
+            VStack(alignment: .trailing, spacing: 0) {
+                Text(instructionText)
+                    .multilineTextAlignment(.center)
+                    .frame(maxWidth: .infinity, alignment: .center)
+                    .padding(8)
+                    .background(.ultraThinMaterial, ignoresSafeAreaEdges: .horizontal)
```
philium, line 99: `ultraThinMaterial` is pretty dark. Consider using `thinMaterial` for all overlays.

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,280 @@
...
+                
+                if let selectedElement = model.selectedElement {
+                    VStack(alignment: .leading) {
+                        LabeledContent("Asset Group:", value: selectedElement.assetGroup.name)
+                        LabeledContent("Asset Type:", value: selectedElement.assetType.name)
```
philium, line 104: The label is already set off from the content with a bold appearance. Using a colon as well is quite unusual.

```suggestion
                        LabeledContent("Asset Group", value: selectedElement.assetGroup.name)
                        LabeledContent("Asset Type", value: selectedElement.assetType.name)
```

Also, consider whether "Asset" is really needed. Could the labels just be "Group" and "Type"?

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,280 @@
...
+                    .disabled(model.selectedElement == nil)
+                }
+            }
+        }
+    }
+}
+
+// MARK: - Helper Views
+
+private extension SnapGeometryEditsWithUtilityNetworkRulesView {
```
philium, line 157: Putting these views in a private extension rather than making them private at the top level means that everything is indented by one level. I suggest getting rid of the extension and just making the views private.

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,280 @@
...
+            graphicsOverlay.id
+        case let layerContent as LayerContent:
+            layerContent.name
+        default:
+            "\(self)"
+        }
+    }
+}
+
+extension SnapRuleBehavior? {
```
philium, line 255: This was really surprising at first. Consider not using syntactic sugar here to improve readability:

```suggestion
extension Optional<SnapRuleBehavior> {
```

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,280 @@
...
+    fileprivate static var allCases: [Self] {
+        return [nil, .rulesLimitSnapping, .rulesPreventSnapping]
```
philium, line 257:
```suggestion
        return [.none, .rulesLimitSnapping, .rulesPreventSnapping]
```

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
CalebRas: I included colon because both the title and value appear the same on Mac Catalyst. See screenshot below (without colon).
Screenshot 2025-04-01 at 9 22 30 AM

Also, I would opt to keep "Asset" since it is used in the property names (`assetGroup`, `assetType`).

I honestly don't feel very strongly about either, though, so feel free to weigh in again if you still want these changes.

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
CalebRas: I went with your design but moved `ruleBehaviorImages` to the main view since they are also needed by `SnapSourcesList` and creating them in both places seemed redundant.

Note that in your screenshot, the legend is a different color than the toolbar because the legend is overlaying the attribution bar. I was going to hide the attribution bar whenever the legend is showing (e.i., when there is a selected feature), but can't because of `swift/issues/6728`. Hiding the attribution bar can be done once that issue is resolved.

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
```diff
@@ -0,0 +1,300 @@
...
+            layerContent.name
+        default:
+            "\(self)"
+        }
+    }
+}
+
+extension Optional<SnapRuleBehavior> {
+    fileprivate static var allCases: [Self] {
+        return [.none, .rulesLimitSnapping, .rulesPreventSnapping]
```
rolson, line 277: interesting that we made "none" the nil case, but have to add none back in here for the UI purposes. Do you think we made the right choice?

rolson (approved):

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
yo1995: In samples we tend to show all the options including `nil` in the picker, so we need to add `Optional.none` back for the UI.

In real world apps, I'm seeing a trend of "toggle + picker" UI (e.g. Apple's Reminders app), where toggle-on shows the picker, and the picker shows all non-optional cases. In this sort of UI, toggle-off means `nil`, which backs up our decision in issue 3380 to make some enums `none` case -> optional.

philium: > Do you think we made the right choice?

Yes, absolutely. The none in this list is actually `Optional.none`. Being able to use it when dealing with an optional value is pretty cool. You can actually switch over an optional value and have a `.none` case as if `none` was a case of the enum itself, making the switch exhaustive.

Shared/Samples/Snap geometry edits with utility network rules/SnapGeometryEditsWithUtilityNetworkRulesView.swift
philium: That's a bummer that the weights are different on macOS. I didn't even consider that. All well.

philium (approved):

CalebRas:
Thank you for your feedback everyone!

PR #583, opened 2025-04-28
[New] Add WFS layer

des12437:
## Description

This PR implements `Add WFS layer` in `Layers` category.
URL to README: [URL](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/destiny/Add-WFS-layer/Shared/Samples/Add%20WFS%20layer)

## Linked Issue(s)

- `swift/issues/6818`

## How To Test
Open the sample and verify that new features populate on the map after panning.

## Screenshots

add-wfs-layer


Shared/Samples/Add WFS layer/AddWFSLayerView.swift
```diff
@@ -0,0 +1,125 @@
...
+    
+    /// The error shown in the error alert.
+    @State private var error: Error?
+    
+    var body: some View {
+        MapView(map: map)
+            .onVisibleAreaChanged { visibleArea = $0 }
+            .onNavigatingChanged { isNavigating in
+                guard !isNavigating else { return }
+                Task {
```
rolson, line 57: this unstructured task will not get cancelled. It would be better to have a state variable for `isNavigating` and then use a .task modifier to run when isNavigating changes. That makes it a task that SwiftUI manages the lifetime of and will cancel previous tasks when it runs a new one.

Shared/Samples/Add WFS layer/AddWFSLayerView.swift
```diff
@@ -0,0 +1,127 @@
...
+
+@MainActor
+struct AddWFSLayerView: View {
+    /// A map with a topographic basemap centered on downtown Seattle.
+    @State private var map: Map = {
+        let map = Map(basemapStyle: .arcGISTopographic)
+        map.initialViewpoint = Viewpoint(
+            boundingGeometry: Envelope(
+                min: Point(x: -122.341581, y: 47.617207, spatialReference: .wgs84),
+                max: Point(x: -122.332662, y: 47.613758, spatialReference: .wgs84)
```
rolson, line 26: are the y values for min/max flipped?

des12437: The values match the order in the test design implementation.

philium: You could use the range-based initializer. This one is potentially confusing because `min` and `max` aren't truly that. It just uses the min and max x and y values from both points (the max can supply the min and the min could supply the max).

dfeinzimer (dismissed):

Shared/Samples/Add WFS layer/AddWFSLayerView.swift
```diff
@@ -100,12 +107,12 @@ struct AddWFSLayerView: View {
     
     /// Populates the feature table using queried features contained within a given geometry.
     /// - Parameter geometry: The geometry used to filter the results.
-    private func populateFeatures(within geometry: Geometry?) async throws {
+    private func populateFeatures(within geometry: Geometry) async throws {
```
rolson, line 110: The type here should be envelope as that is what we are always expecting to pass to it, and the name of the parameter should be `extent`.

rolson (dismissed):
I didn't mind the function being throwing, but only thought the type and arg needed to change. Up to you on where the throwing is.

rolson (approved):

dfeinzimer (approved):

PR #613, opened 2025-05-13
[New] Authenticate with token

des12437:
## Description

This PR implements `Authenticate with token` in `Cloud and Portal` category.
URL to README: [URL](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/3a649d75e3bcb634738575e1c5bcd350eb8b0d9d/Shared/Samples/Authenticate%20with%20token/README.md)

## Linked Issue(s)

- `swift/issues/6858`

## How To Test
- Open the sample and sign in with valid credentials, verify the secured layer loads.
- Sign in with invalid credentials and verify that an error message is shown.

## Screenshots

authenticate-with-token1 authenticate-with-token2


Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
```diff
@@ -0,0 +1,133 @@
...
+            return error.details
+        } else {
+            return error.localizedDescription
+        }
+    }
+    
+    /// Signs out from the portal by revoking OAuth tokens and clearing credential stores.
+    private func signOut() {
+        Task {
+            await ArcGISEnvironment.authenticationManager.revokeOAuthTokens()
```
des12437, line 104:
```suggestion
```
Is this needed for token authentication?

Shared/Samples/Authenticate with token/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Allows you to access a secure service with the convenience and security of ArcGIS token-based authentication. For example, rather than providing a user name and password every time you want to access a secure service, you only provide those creditials initially to obtain a token which then can be used to access secured resources.
+
+## How to use the sample
+
+Once you launch the app, you will be challenged for an ArcGIS Online login to view the protected map service. Enter a user name and password for an ArcGIS Online named user account (such as your ArcGIS for Developers account). If you authenticate successfully, the protected map service will display in the map.
+
+## How it works
+
+1. Create a toolkit component `Authenticator` object.
```
des12437, line 18: Should the order of these steps be changed to combine these two steps at the end?

```
5. Create a toolkit component `Authenticator` object.
6. Set authenticator object as ArcGIS and Network challenge handlers on authentication manager to handle authentication challenges.

**Review by mhdostal — APPROVED** [review]
Looks good once the questions are answered.

File: Shared/Samples/Authenticate with token/README.md (hunk 2 above)
Comment (njarecha) [reply Shared/Samples/Authenticate with token/README.md:18]: To answer above question, no. Actually, a step to set the authenticator view modifier is missing.

I see different style of `How it works` steps for IWA sample. What is the general guideline?

**Review by njarecha — CHANGES_REQUESTED** [review]
Done with first pass

File: Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift (hunk 1 above)
Comment (des12437) [reply Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift]: Done.

File: Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+                    Text("Error loading map: \(errorString(for: error))")
+                        .padding()
+                }
+            } else {
+                ProgressView()
+                    .task {
+                        mapLoadResult = await Result { @MainActor in
+                            let map = Self.makeMap()
+                            try await map.load()
+                            try await map.operationalLayers.first?.load()
```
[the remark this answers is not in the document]
des12437: No, the design does not specify to load the layer, but in my testing if this layer is not loaded, then the map will successfully load regardless of using valid or invalid credentials. Is this expected? The web map is public and the traffic layer is secured. When authentication fails the traffic layer won't show, but the map's load status is `loaded`.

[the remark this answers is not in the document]
njarecha: Ok, keep it then.

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
```diff
@@ -0,0 +1,132 @@
...
+        }
+        .onDisappear {
+            // Resetting the challenge handlers and clearing credentials here in `onDisappear`
+            // so user is prompted to enter credentials every time trying the sample. In real
+            // world applications, do these from sign out functionality of the application.
+            
+            // Resets challenge handlers.
+            ArcGISEnvironment.authenticationManager.handleChallenges(using: nil)
+            
+            signOut()
```
njarecha, line 85: Please follow IWA and PKI samples and keep the same code in separate functions. There is some discussion in PKI sample for keeping the revoke tokens function as well.

Shared/Samples/Authenticate with token/README.md
njarecha: Ok

njarecha (changes requested):
Make changes based on other auth samples for setting up and tearing down authenticator.

njarecha (approved):
Thank you for making all the changes

des12437:
Removed Mark as a reviewer as he is out of office today.

njarecha (dismissed):

njarecha (dismissed):

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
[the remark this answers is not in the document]
rolson: The map loads either way. What fails to load if you do not authenticate is the layer. When a map loads, it does not load all its layers. It only typically loads a base layer so that it can get its spatial reference. Your load result was failure because of the `try await map.operationalLayers.first?.load()`.

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+}
+
+private extension AuthenticateWithTokenView {
+    /// The string describing an error.
+    /// - Parameter error: The error.
+    func errorString(for error: Error) -> String {
+        return if error is ArcGISChallengeCancellationError {
+            "User cancelled error"
```
rolson, line 84: Where and how this is displayed, this string might make more sense:

```suggestion
            "User cancelled."
```

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
```diff
@@ -0,0 +1,144 @@
...
+    }
+    
+    var body: some View {
+        Group {
+            if let mapLoadResult = mapLoadResult {
+                switch mapLoadResult {
+                case .success(let value):
+                    MapView(map: value)
+                case .failure(let error):
+                    Text("Error loading map: \(errorString(for: error))")
```
rolson, line 48: Could be an error loading the map. But more than likely if not authenticated, then it's an error loading the traffic layer.

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
```diff
@@ -0,0 +1,144 @@
...
+                        .padding()
+                }
+            } else {
+                ProgressView()
+                    .task {
+                        mapLoadResult = await Result { @MainActor in
```
rolson, line 54: It's more than the "map load result". I don't know of a better name.

rolson (commented):
The only thing that needs to authenticate is the traffic layer. I'll have to think about a way to simplify this.

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
des12437: Yes its an error loading the traffic layer not the map.

rolson:
I think this sample can be simplified with something like this, which just looks at the layer view state and shows an error:

```swift
struct AuthenticateWithTokenView: View {
    /// The authenticator to handle authentication challenges.
    @StateObject private var authenticator = Authenticator()
    
    @State private var map = {
        // The portal to authenticate with named user.
        let portal = Portal(url: .portal, connection: .authenticated)
        
        // The portal item to be displayed on the map.
        let portalItem = PortalItem(
            portal: portal,
            id: .trafficMap
        )
        
        // Creates map with portal item.
        return Map(item: portalItem)
    }()
    
    @State private var error: Error?
    
    var body: some View {
        MapView(map: map)
            .onLayerViewStateChanged { layer, layerViewState in
                if layerViewState.status == .error {
                    error = layer.loadError
                }
            }
            .errorAlert(presentingError: $error)
            .authenticator(authenticator)
            .onAppear {
                setupAuthenticator()
            }
            .onDisappear {
                Task {
                    // Reset the challenge handlers and clear credentials
                    // when the view disappears so that user is prompted to enter
                    // credentials every time the sample is run, and to clean
                    // the environment for other samples.
                    await teardownAuthenticator()
                }
            }
    }
}
```

You might want to fixup the error alert to say what you are looking for it to say with the auth error. And you could also check the layer name or that it's the first operational layer before setting the error.

des12437:
@rolson Thanks I applied your suggestion.

> you could also check the layer name or that it's the first operational layer before setting the error.

This map has only one operational layer, so checking the name may not be necessary.


> You might want to fixup the error alert to say what you are looking for it to say with the auth error. 

Here is what the error alerts looks like when the user cancels authentication and when authentication fails due to invalid credentials respectively:

rolson:
You can fixup the error to show what you want if you like the strings you had before. I just didn't go through that exercise.

rolson:
> This map has only one operational layer, so checking the name may not be necessary.

Well, if you are only looking to show an error when the operational layer fails, then you might want to check that the layer is the first operational layer in your map before setting the error.

Shared/Samples/Authenticate with token/AuthenticateWithTokenView.swift
```diff
@@ -41,6 +41,7 @@ struct AuthenticateWithTokenView: View {
     var body: some View {
         MapView(map: map)
             .onLayerViewStateChanged { layer, layerViewState in
+                guard layer.name == map.operationalLayers.first?.name else { return }
```
rolson, line 44: why not check the layer instance?

des12437:
> You can fixup the error to show what you want if you like the strings you had before. I just didn't go through that exercise.

I used a string before since a standalone view with Text was shown for the error. The refactor uses an errorAlert, which requires an Error. I'm ok leaving this as-is and only using an alert for the error.

rolson (approved):

njarecha (approved):

PR #617, opened 2025-05-14
[New] Show geodesic path between two points

rolson:
## Description

This PR implements `ShowGeodesicPathBetweenTwoPoints` in `Edit and Manage Data` category.
URL to README: 

## Linked Issue(s)

- `swift/issues/6872`

## How To Test

- Open the sample
- Tap on the mapview in 2 different locations

## Screenshots

show-geodesic-path-between-two-points


Shared/Samples/Show geodesic path between two points/ShowGeodesicPathBetweenTwoPointsView.swift
```diff
@@ -0,0 +1,129 @@
...
+                switch state {
+                case .notStarted, .complete:
+                    // If the state is empty or complete, then start a new
+                    // path, adding the tap point as the first graphic.
+                    state = .startOnly(start: mapPoint)
+                case .startOnly(let start):
+                    // If the state was started, then add the end point
+                    // to complete it.
+                    state = .complete(start: start, end: mapPoint)
+                }
```
philium, line 52: You could use a switch expression here.

```suggestion
                state = switch state {
                case .notStarted, .complete:
                    // If the state is empty or complete, then start a new
                    // path, adding the tap point as the first graphic.
                    .startOnly(start: mapPoint)
                case .startOnly(let start):
                    // If the state was started, then add the end point
                    // to complete it.
                    .complete(start: start, end: mapPoint)
                }
```

Shared/Samples/Show geodesic path between two points/ShowGeodesicPathBetweenTwoPointsView.swift
```diff
@@ -0,0 +1,129 @@
...
+            }
+            .overlay(alignment: .top) {
+                VStack {
```
philium, line 55: Since this stack only ever contains one subview, it could just be a group.

```suggestion
                Group {
```

Shared/Samples/Show geodesic path between two points/ShowGeodesicPathBetweenTwoPointsView.swift
```diff
@@ -0,0 +1,129 @@
...
+                    switch state {
+                    case .notStarted, .startOnly:
+                        Text("Tap on the map to show a geodesic path")
+                    case .complete(_, _, _, let length):
+                        Text(length.formatted())
```
philium, line 60: Remember that when formatting text using SwiftUI, providing a format is preferred over using a `formatted()` method. This allows SwiftUI to efficiently format the value only when needed.

```suggestion
                        Text(length, format: .measurement(width: .abbreviated))
```

philium (commented):
Regions outside of the basemap can be tapped and a distance is even provided. That doesn't make sense to me, but I'm not sure there's anything that we can do about it.

image

rolson:
> Regions outside of the basemap can be tapped and a distance is even provided. That doesn't make sense to me, but I'm not sure there's anything that we can do about it.
> 

Right, nothing we can do.

philium (approved):

rolson:
@pgruenler please review

pgruenler (approved):

PR #629, opened 2025-05-23
[New] List geodatabase versions

yo1995:
## Description

This PR implements `List geodatabase versions` in `Edit and Manage Data` category.
[URL to README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/97f4aae4ba1fc103c7390ca294d8b517f06f5753/Shared/Samples/List%20geodatabase%20versions/README.md)

## Linked Issue(s)

- `swift/issues/6920`

## How To Test

Open the sample and see the versions from a GP server.

### DIFF
--- a/Shared/Samples/List geodatabase versions/ListGeodatabaseVersionsView.swift
+++ b/Shared/Samples/List geodatabase versions/ListGeodatabaseVersionsView.swift
@@ -50,9 +50,8 @@ struct ListGeodatabaseVersionsView: View {
         }
         .task {
             do {
-                if let resultFeatures = try await fetchGeodatabaseVersions() {
-                    geodatabaseVersionInfos = makeGeodatabaseVersionInfos(from: resultFeatures)
-                }
+                let resultFeatures = try await fetchGeodatabaseVersions()
+                geodatabaseVersionInfos = makeGeodatabaseVersionInfos(from: resultFeatures)
             } catch {
                 analysisError = error
             }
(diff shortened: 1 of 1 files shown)


dfeinzimer (dismissed):

Shared/Samples/List geodatabase versions/ListGeodatabaseVersionsView.swift
```diff
@@ -0,0 +1,133 @@
...
+                }
+            } catch {
+                analysisError = error
+            }
+        }
+        .errorAlert(presentingError: $analysisError)
+    }
+    
+    /// Fetches geodatabase versions using a geoprocessing job.
+    private func fetchGeodatabaseVersions() async throws -> FeatureSet? {
```
rolson, line 64: Throwing and also returning an optional might not always be the best choice. Should you throw some well-defined error instead of returning `nil` or does an optional really make sense here?

Shared/Samples/List geodatabase versions/README.metadata.json
```diff
@@ -0,0 +1,33 @@
+{
+    "category": "Analysis",
```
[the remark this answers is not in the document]
rolson: Edit and Manage Data makes more sense to me.

rolson (approved):

Shared/Samples/List geodatabase versions/ListGeodatabaseVersionsView.swift
yo1995: From core yaml:

> (GeoprocessingFeatures.features) It can be nil for services that have a map service, in which case calling `fetchOutputFeatures()` will retrieve the features from the service.

Thus, as long as the service provided supports versioning, the output `featureSet` won't be `nil` if the geoprocessing job successfully finishes with the results. I decided to `fatalError` in the case of `nil`.

dfeinzimer (approved):

PR #635, opened 2025-05-31
[New] Query table statistics

CalebRas:
## Description

This PR implements `Query table statistics` in `Search and Query` category.
URL to README: [Query table statistics](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/Caleb/New-QueryTableStatistics/Shared/Samples/Query%20table%20statistics)

## Linked Issue(s)

- `swift/issues/6930`

## How To Test

- Pan, zoom, and use the settings to adjust the query parameters.
- Tap "Query Statistics" to perform the query and view the statistics.

## Screenshots

query-table-statistics


Shared/Samples/Query table statistics/QueryTableStatisticsView.swift
```diff
@@ -0,0 +1,217 @@
...
+        map.addOperationalLayer(featureLayer)
+    }
+    
+    /// Queries the statistics of the data in the given extent.
+    /// - Parameters:
+    ///   - extent: The geometry to query within.
+    ///   - includesCitiesUnder5M: A Boolean value indicating whether queries
+    ///   will include cities with populations under 5 million.
+    /// - Returns: The statistic records resulting from the query.
+    func queryStatistics(
```
des12437, line 176:
```suggestion
    /// - Returns: The statistic records resulting from the query.
    @MainActor
    func queryStatistics(
```
The main actor annotation can be added to this method rather than to the class.

Shared/Samples/Query table statistics/QueryTableStatisticsView.swift
```diff
@@ -0,0 +1,217 @@
...
+
+private extension StatisticDefinition.StatisticType {
+    /// All the statistic types.
+    static var allCases: [Self] {
+        return [average, count, maximum, minimum, standardDeviation, sum, variance]
+    }
+}
+
+private extension URL {
+    /// The web URL to the "Cities" layer on the "SampleWorldCities" map service.
```
des12437, line 209:
```suggestion
    /// The URL to the "Cities" layer on the "SampleWorldCities" map service.
```

des12437 (dismissed):

Shared/Samples/Query table statistics/QueryTableStatisticsView.swift
```diff
@@ -0,0 +1,217 @@
...
+    
+    /// The parameters to use in the statistics query.
+    private let statisticsQueryParameters: StatisticsQueryParameters = {
+        // Creates a statistic definition for the "POP" field and each of the statistic types.
+        let statisticDefinitions = StatisticDefinition.StatisticType.allCases.map { statisticType in
+            StatisticDefinition(fieldName: "POP", statisticType: statisticType)
+        }
+        
+        return StatisticsQueryParameters(statisticDefinitions: statisticDefinitions)
+    }()
```
philium, line 161: You could simplify this by not using an immediately evaluated closure.

```suggestion
    private let statisticsQueryParameters = StatisticsQueryParameters(
        // Creates a statistic definition for the "POP" field and each of the statistic types.
        statisticDefinitions: StatisticDefinition.StatisticType.allCases
            .map { .init(fieldName: "POP", statisticType: $0) }
    )
```

Shared/Samples/Query table statistics/QueryTableStatisticsView.swift
```diff
@@ -0,0 +1,217 @@
...
+        map.addOperationalLayer(featureLayer)
+    }
+    
+    /// Queries the statistics of the data in the given extent.
+    /// - Parameters:
+    ///   - extent: The geometry to query within.
+    ///   - includesCitiesUnder5M: A Boolean value indicating whether queries
+    ///   will include cities with populations under 5 million.
+    /// - Returns: The statistic records resulting from the query.
+    @MainActor
```
philium, line 175: Since the issue has to do with the model not being sendable, instead of constraining this method to the main actor, just make the model adopt `Sendable`. Since all of the model's properties are already sendable, all that would need to be done is simply to declare that the model is sendable.

Shared/Samples/Query table statistics/QueryTableStatisticsView.swift
```diff
@@ -0,0 +1,217 @@
...
+        
+        return statisticRecords
+    }
+}
+
+private extension StatisticRecord {
+    /// The identifier for the statistic record object.
+    var objectID: ObjectIdentifier {
+        ObjectIdentifier(self)
+    }
```
philium, line 198:
```suggestion
    /// The object identifier for this statistic record.
    var objectID: ObjectIdentifier { .init(self) }
```

Shared/Samples/Query table statistics/QueryTableStatisticsView.swift
```diff
@@ -0,0 +1,217 @@
...
+    /// The action to dismiss the view.
+    @Environment(\.dismiss) private var dismiss
+    
+    var body: some View {
+        NavigationStack {
+            List(records, id: \.objectID) { record in
+                Section {
+                    let statistics = record.statistics.sorted(by: { $0.key < $1.key })
+                    ForEach(statistics, id: \.key) { name, value in
+                        LabeledContent(name, value: "\(value)")
```
philium, line 115: While this works, it isn't modeling best practices. `value` is turned into a string using `String(describing:)`, which is neither formatted nor localized. It would be better to switch over `value` with cases for the expected types and format them so as to set an example for others to follow.

CalebRas: It looks like everything in this case was an `Int` or `Double`, so I only handled those. Let me know if you had something else in mind.

philium (approved):

CalebRas:
Thank you!

PR #664, opened 2025-07-18
[Fix] Update frame on popover

chriswebb09:
## Screenshots

**Line of sight between geoelements**

Looks fine on iOS and Catalyst. 0.25 seemed like a reasonable amount but there is no reason to stick with it. 

|Before|After|
|:-:|:-:|
| Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-18 at 08 24 05 | Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-07-18 at 11 12 13 | 

**Browse WMS Layer**

|Before|After|
|:-:|:-:|
|  Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-07-18 at 09 38 32  | Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-07-18 at 15 09 41 | 

|Before|After|
|:-:|:-:|
|  Simulator Screenshot - iPhone 16 Pro - 2025-07-18 at 11 25 26 | Simulator Screenshot - iPhone 16 Pro - 2025-07-18 at 11 23 10 |


Shared/Samples/Browse WMS layers/BrowseWMSLayersView.swift
CalebRas, line 1: This isn't exactly related to frames, but it looks like the popup in this sample suffers from a list background issue similar to https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/pull/1211

Consider removing the `.listStyle(.plain)` line in `WMSLayerListView`, so that the default list style is used instead.

chriswebb09: Should it be set to inset or entirely removed?

CalebRas: Thanks for making that change. Should the sample's screenshot be updated?

chriswebb09: Gotcha, this is the new screenshot 

browse-wms-layers

yo1995 (approved):

PR #667, opened 2025-07-22
[Fix] Sample teardown issues

CalebRas:
## Description

This PR fixes an issue where sample teardown wouldn't always complete before another one appears. This would cause bugs when switching between samples that both access the same shared state, such as `ArcGISEnvironment`.

This solution blocks sample requiring teardown from appearing until the previous teardown sample has finished cleaning up. The previous sample lets the app know its teardown is complete by calling `onTearDownCompleted()` from the environment. The app identifies teardown samples via a new `has_teardown` metadata key. The samples without this key are not blocked from appearing at all. Sample with teardown in a view model required a little extra refactoring.

<details><summary>Affected Samples</summary>

- Add items to portal
- Add feature layers
- Analyze network with subnetwork trace
- Authenticate with Integrated Windows Authentication
- Authenticate with OAuth
- Authenticate with PKI certificate
- Authenticate with token
- Create and save map
- Create load report
- Display content of utility network container
- Edit with branch versioning
- Run valve isolation trace
- Set visibility of subtype sublayer
- Show utility associations
- Trace utility network
- Validate utility network topology

</details>

## Linked Issue(s)

- `swift/issues/7115`

## How To Test

- Clean the project to re-generate the `Sample` structs with the new key.
- Switch between the affected samples and ensure they load/prompt for credentials correctly.

<details><summary>Helper Code</summary>

- Replacing `FavoritesView.favoriteSamples` with the following code with populate the "Favorites" category with the samples to test.

```swift 
static let sampleNames = [
    "Add items to portal",
    "Add feature layers",
    "Analyze network with subnetwork trace",
    "Authenticate with Integrated Windows Authentication",
    "Authenticate with OAuth",
    "Authenticate with PKI certificate",
    "Authenticate with token",
    "Create and save map",
    "Create load report",
    "Display content of utility network container",
    "Edit with branch versioning",
    "Run valve isolation trace",
    "Set visibility of subtype sublayer",
    "Show utility associations",
    "Trace utility network",
    "Validate utility network topology"
].shuffled()

private var favoriteSamples: [Sample] = {
    sampleNames.map { name in
        SamplesApp.samples.first { $0.name == name }!
    }
}()
```

</details>


yo1995:
Welp, looks like need to make a change to ignore the `has_teardown` field if we go down this path. I can make a to update the checker script.

https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/636ec6056033935654df4e2c8750ffea92f3a6d6/Scripts/CI/metadata_checker.py#L49-L57

njarecha (dismissed):
Samples works as expected now. Nice Work!

Shared/Supporting Files/Views/CategoriesView.swift
```diff
@@ -149,3 +171,11 @@ private extension CategoriesView {
         }
     }
 }
+
+extension EnvironmentValues {
+    /// The action to run when a sample is done tearing down.
+    ///
+    /// - Note: Failure to call this in a ``Sample/hasTeardown`` sample will
+    /// block other teardown samples from appearing.
+    @Entry var onTearDownCompleted: () -> Void = {}
```
philium, line 180: `onTeardown` is a great name for a modifier, but an odd name for an environment value. Since this is an action, a better name would be something like `finishTeardown`.

philium (commented):
This is an interesting approach! Overall it looks good, but I do have a couple comments. First, "tear down" is a verb and "teardown" is a noun. So something like "on teardown" should be capitalized `onTeardown`, not `onTearDown`. Second, the code in each of the samples could be simplified by adding a custom modifier:

```swift
extension View {
    func onTeardown(perform action: @escaping () -> Void) -> some View {
        modifier(OnTeardown(action: action))
    }
}

private struct OnTeardown: ViewModifier {
    let action: () -> Void
    
    /// The action to run when the sample's teardown has completed.
    ///
    /// This is needed to prevent the authentication in this sample from interfering with other samples.
    @Environment(\.onTearDownCompleted) private var onTearDownCompleted
    
    func body(content: Content) -> some View {
        content
            .onDisappear {
                action()
                onTearDownCompleted()
            }
    }
}
```

Then it could be used in a sample like so:

```swift
.onTeardown {
    model.tearDown()
}
```

Shared/Supporting Files/Extensions/View+OnTeardown.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// The teardown action to perform.
+    let action: () async -> Void
+    
+    /// The action to run when the teardown action has completed.
+    @Environment(\.finishTeardown) private var finishTeardown
+    
+    func body(content: Content) -> some View {
+        content
+            .onDisappear {
+                Task {
```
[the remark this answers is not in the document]
CalebRas: I added a non-async version but modified your solution a bit so that the memberwise init could still be used (for simplicity). Let me know if you would rather it the other way.

CalebRas:
Sorry, I missed some extra colon spaces. Should be good to go now.

yo1995 (dismissed):
Nicely done! 👍

Shared/Samples/Add feature layers/AddFeatureLayersView.swift
```diff
@@ -53,9 +53,10 @@ struct AddFeatureLayersView: View {
                 // specified credentials and tokens for any challenges.
                 ArcGISEnvironment.authenticationManager.arcGISAuthenticationChallengeHandler = ChallengeHandler()
             }
-            .onDisappear {
+            .onTeardown {
                 // Resets the URL session challenge handler to use default handling.
```
njarecha, line 57:
```suggestion
                // Resets the URL session challenge handler to use default handling and remove all credentials.
```

Shared/Samples/Authenticate with OAuth/AuthenticateWithOAuthView.swift
```diff
@@ -53,24 +53,22 @@ struct AuthenticateWithOAuthView: View {
...
     
     /// Signs out from the portal by revoking OAuth tokens and clearing credential stores.
-    private func signOut() {
-        Task {
-            await ArcGISEnvironment.authenticationManager.revokeOAuthTokens()
-            await ArcGISEnvironment.authenticationManager.clearCredentialStores()
-        }
+    private func signOut() async {
+        await ArcGISEnvironment.authenticationManager.revokeOAuthTokens()
+        await ArcGISEnvironment.authenticationManager.clearCredentialStores()
```
njarecha, line 71: These are functions of extension in the toolkit. The new `signOut` function is exposed. Should all samples use it?

Shared/Samples/Display content of utility network container/DisplayContentOfUtilityNetworkContainerView.Model.swift
```diff
@@ -72,9 +72,11 @@ extension DisplayContentOfUtilityNetworkContainerView {
             ArcGISEnvironment.authenticationManager.arcGISAuthenticationChallengeHandler = ChallengeHandler()
         }
         
-        deinit {
+        /// Cleans up the model's setup.
+        func tearDown() {
             // Resets the URL session challenge handler to use default handling.
```
njarecha, line 77:
```suggestion
            // Resets the URL session challenge handler to use default handling and remove all credentials.
```

Shared/Samples/Run valve isolation trace/RunValveIsolationTraceView.Model.swift
```diff
@@ -116,6 +111,13 @@ extension RunValveIsolationTraceView {
             }
         }
         
+        /// Cleans up the model's setup.
+        func tearDown() {
+            // Resets the URL session challenge handler to use default handling.
```
njarecha, line 116:
```suggestion
            // Resets the URL session challenge handler to use default handling and remove all credentials.
```

Shared/Samples/Show utility associations/ShowUtilityAssociationsView.swift
```diff
@@ -144,6 +142,13 @@ private extension ShowUtilityAssociationsView {
             try await network.load()
         }
         
+        /// Cleans up the model's setup.
+        func tearDown() {
+            // Resets the URL session challenge handler to use default handling.
```
njarecha, line 147:
```suggestion
            // Resets the URL session challenge handler to use default handling and remove all credentials.
```

Shared/Samples/Trace utility network/TraceUtilityNetworkView.Model.swift
```diff
@@ -91,9 +91,11 @@ extension TraceUtilityNetworkView {
             ArcGISEnvironment.authenticationManager.arcGISAuthenticationChallengeHandler = ChallengeHandler()
         }
         
-        deinit {
+        /// Cleans up the model's setup.
+        func tearDown() {
             // Resets the URL session challenge handler to use default handling.
```
njarecha, line 96:
```suggestion
            // Resets the URL session challenge handler to use default handling and remove all credentials.
```

njarecha (changes requested):
Some suggestions

philium (approved):

Shared/Samples/Authenticate with OAuth/AuthenticateWithOAuthView.swift
njarecha: > Also, I notice that the AuthenticationManager.signOut() method is missing a Since tag in the toolkit. Does it need one?

Yes, good to add.

philium (approved):

CalebRas:
Thank you everyone for your feedback! I appreciate it.

Shared/Supporting Files/Views/CategoriesView.swift
```diff
@@ -76,6 +79,26 @@ struct CategoriesView: View {
...
+                            }
+                            .environment(\.finishTeardown) {
+                                // Allows the next teardown sample to appear.
+                                sampleNeedingTearDown = nil
+                            }
+                    }
+                } else {
+                    SampleDetailView(sample: sample)
+                }
+            }
```
yo1995, line 101: While ad-hoc testing, I realized this logic only accounts for switching between samples which require teardown, i.e., the outgoing and incoming samples both rely on `ArcGISEnvironment` and has the `has_teardown` field in metadata.

However, all the samples that use an API key should wait, if the outgoing sample resets the API key. Otherwise, the incoming sample will have an empty API key while the outgoing one hasn't finished tear down.

So, I think this logic needs adjustment. There are a few cases:

- Teardown to Self: no wait
- Teardown to another Teardown: wait till finish
- ❌ (currently not accounted for) Teardown to non-Teardown: wait till finish
- non-Teardown to Teardown: no wait
- non-Teardown to non-Teardown: no wait

Thoughts?

PR #668, opened 2025-07-23
[Setup] Adapt scripts to add has_teardown

yo1995:
## Description

This PR updates the CI checker scripts to add the "has_teardown" metadata field, for the reasons listed in #667 .

## How To Test

See that I tweaked a metadata file below to pass then reverted.

35853b8: [check 1](https://github.com/Esri/arcgis-maps-sdk-swift-samples/actions/runs/16458283449/job/46520280546): check fails if `"has_teardown" : false` was specified (when a sample doesn't need teardown, it shouldn't specify this field)

414105d: [check 2](https://github.com/Esri/arcgis-maps-sdk-swift-samples/actions/runs/16458439854/job/46520760400): check succeeds if `"has_teardown" : true`, which is used to check those that require teardown

47919fd: when there is no this field, the check succeeds

### DIFF
--- a/Scripts/CI/common.py
+++ b/Scripts/CI/common.py
@@ -374,7 +374,7 @@ class Metadata:
         data["snippets"] = self.snippets
         data["title"] = self.title
         if self.has_teardown:
-            # Only write offline_data when it is not empty.
+            # Only write has_teardown when it is not empty.
             data["has_teardown"] = self.has_teardown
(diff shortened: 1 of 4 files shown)


CalebRas (approved):
Looks good, thanks for doing this!

pgruenler (approved):

PR #673, opened 2025-07-24
[Setup] Add debug section to allow swapping API key

yo1995:
## Description

This PR adds a debug textfield to enter a different API key, in the case that a legacy key stops working but we still want to test an older daily build.

## Linked Issue(s)

- `swift/issues/7131`

## How To Test

- Put in an invalid key to see the app doesn't load basemap
- Put in a valid key to see the app works
- Use reset button to see the app works

## Screenshot

Simulator Screenshot - iPhone 16e - 2025-07-24 at 15 12 07

### DIFF
--- a/Shared/Supporting Files/Views/AboutView.swift
+++ b/Shared/Supporting Files/Views/AboutView.swift
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


saratchandrakarumuri (approved):
Tested with debug mode and this option is showing up as expected and hidden in release mode.

PR #683, opened 2025-07-31
[Fix] Slider causes time extent to constantly changing on Mac Catalyst

yo1995:
## Description

This PR fixes 2 issues for the `Add feature layer with time offset` sample on Mac Catalyst

- When you drag the slider quickly, the text will start flickering. Use `let _ = Self._printChanges()` and you'll see the `timeExtent` keeps causing the Text view to recompute. It seems to be some kind of debounce problem.
- On Mac Catalyst, when placed in the toolbar, the Text is clipped at the bottom

image

## Linked Issue(s)

- `swift/issues/7143`

## How To Test

- Build v.next to Mac Catalyst and see the problems
- Build this branch to see the changed UI and fixed slider

When review, check "Hide Whitespace" to see the diff better

## To Discuss

I couldn't figure out the root cause of `$timeExtent` not causing a problem on iOS while being buggy on Mac Catalyst. I feel like it might be a Mac Catalyst bug.

By changing it into a `.constant` binding, it ensures the map view's time extent is set-only - the state variable won't be updated by the core geoview's time extent changes.

Feel free to tweak and share your ideas if there is a better solution!


Shared/Samples/Add feature layer with time offset/AddFeatureLayerWithTimeOffsetView.swift
```diff
@@ -63,58 +63,54 @@ struct AddFeatureLayerWithTimeOffsetView: View {
...
+                          ) else {
+                        return
                     }
+                    timeExtent = TimeExtent(startDate: newStartDate, endDate: newEndDate)
                 }
-            }
+            Text(
+                """
+                \(timeExtent?.startDate?.formatted(dateFormat) ?? "") - \
+                \(timeExtent?.endDate?.formatted(dateFormat) ?? "")
```
philium, line 110: `timeExtent` is optional, but it looks like it doesn't need to be. Also, as far as I can tell, there will always be a start date and an end date, so those could just be force unwrapped. Lastly, formatting could (and should) be done using format styles:

```swift
\(timeExtent.startDate!, format: dateFormat) - \
\(timeExtent.endDate!, format: dateFormat)
```

Shared/Samples/Add feature layer with time offset/add-feature-layer-with-time-offset.png
philium, line 1: It's too bad that we can't use the toolbar anymore. The transparent background was looked really nice.

pgruenler (approved):

PR #688, opened 2025-08-05
[Fix] `Set up location driven geotriggers` monitoring

CalebRas:
## Description

This PR fixes a bug with `Set up location-driven geotriggers` where the first "Current Section" wouldn't always be available. The solution was to kick off the `GeotriggerMonitor.notifications` monitoring loop before calling `GeotriggerMonitor.start()`.

## How To Test

Open `Set up location-driven geotriggers` a few times and ensure the "Current Section" button is always enabled **before** the location display leaves the "Desert".


CalebRas:
Thank you!

Issue #448, opened 2024-07-10
Documentation: World scale AR


yo1995:
See https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/437#discussion_r1670329989

yo1995:
That is correct - for world tracking, the heading is the calibrated total heading; for geo tracking, the heading is the most recent heading when the geo tracking is localized. We can chat after tomorrow's scrum.

PR #693, opened 2025-08-08
[Fix] update documentation to clarify

chriswebb09:
## Description
Rewording documentation - related issue: [#448](https://github.com/Esri/arcgis-maps-sdk-swift-samples/issues/448)

### DIFF
--- a/Shared/Samples/Augment reality to navigate route/README.md
+++ b/Shared/Samples/Augment reality to navigate route/README.md
@@ -59,7 +59,7 @@ A digital elevation model is used to ensure that the displayed route is position
 
 **World-scale AR** is one of three main patterns for working with geographic information in augmented reality. Augmented reality is made possible with the ArcGIS Maps SDK Toolkit. See [Augmented reality](https://developers.arcgis.com/swift/scenes-3d/display-scenes-in-augmented-reality/) in the guide for more information about augmented reality and adding it to your app.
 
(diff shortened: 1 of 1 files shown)


Shared/Samples/Augment reality to navigate route/README.md
```diff
@@ -59,7 +59,7 @@ A digital elevation model is used to ensure that the displayed route is position
...
 
-Because most navigation scenarios involve traveling beyond the accurate range for ARKit positioning, this sample relies on **continuous location updates** from the location data source. Because the origin camera is constantly being reset by the location data source, the sample doesn't allow the user to pan to calibrate or adjust the altitude with a slider. The location data source doesn't provide a heading, so it isn't overwritten when the location refreshes.
+Since most navigation scenarios go beyond the range where ARKit can provide accurate positioning, this sample depends on **continuous location updates** from the location data source. Because the location data source constantly resets the origin camera, the sample disables user interactions like panning to calibrate or adjusting altitude with a slider. Additionally, because the location data source doesn’t include heading information, the heading remains unchanged during location updates.
```
yo1995, line 62: > Because the location data source constantly resets the origin camera, the sample disables user interactions like panning to calibrate or adjusting altitude with a slider.

Because we introduced geotracking to world-scale, this isn't accurate anymore.

> Additionally, because the location data source doesn’t include heading information, the heading remains unchanged during location updates.

This part still needs to be rephrased. location data source [includes](https://developers.arcgis.com/swift/api-reference/documentation/arcgis/locationdatasource/headings) headings, it is just we are not continuously picking it up. For world tracking, the heading is configured by the calibration sliders; for geo-tracking, the heading is only used to set the initial camera controller - once it is set it relies on the geotracking to align the camera.

---

Ideally, we can find a way to describe both modes in the same paragraph. But if we cannot, then maybe this paragraph needs to be broken into 2 for each mode. 😕

chriswebb09: That makes sense to me.

yo1995 (approved):

Issue #321, opened 2023-12-15
Don


yo1995:

CalebRas:
> I should clarify that in most (if not all) cases, Enabled shouldn't be needed as a suffix. While I couldn't find any explicit recommendations in the HIG, a cursory glance at how switches are used in iOS shows that toggle labels typically do not include the state as part of the label. The state is a function of the switch itself and including it in the label is odd, especially for VoiceOver. If it must be included in the label, it should be static, ie. it shouldn't change depending on the label's state.

On Mac Catalyst, a toggle in a toolbar using the default formatting appears as a plain button with no dynamic indication of whether it is enabled/disabled. Using a dynamic suffix was the best idea I could come up with to remedy this. I agree, though, that it is probably not an optimal solution. Do you have any other ideas, or do you recommend just leaving the buttons static? @philium

https://github.com/Esri/arcgis-maps-sdk-swift-samples/assets/34120440/d79c9e7f-833b-4ee3-b550-5aafebfb46dc

PR #698, opened 2025-08-14
[Update] Remove enable / disable from toggle label. 

chriswebb09:
## Description

Removes enabled from toggle label. Removes SwiftLint for force try.

## Screenshots

Simulator Screenshot - iPhone 16 Pro - 2025-08-14 at 16 45 17

Simulator Screenshot - iPhone 16 Pro - 2025-08-14 at 16 46 09

Simulator Screenshot - iPhone 16 Pro - 2025-08-14 at 16 47 36

Simulator Screenshot - iPhone 16 Pro - 2025-08-14 at 16 48 00


CalebRas:
It looks like the Mac Catalyst [issue](https://github.com/Esri/arcgis-maps-sdk-swift-samples/issues/321#issuecomment-1858242980) where the button toggles appear the same when on and off is still present on 15.5. I am also not sure if there is a better way to work around that than having a conditional label since the HIG specifically says to [only use the button style](https://developer.apple.com/design/human-interface-guidelines/toggles#macOS) when the toggle is in a toolbar on Mac Catalyst:

> Use switches, checkboxes, and radio buttons in the window body, not the window frame. In particular, avoid using these components in a toolbar or status bar.

As a result, my vote is that we leave the conditional labels when the toggle is shown in a toolbar. I don't feel too strongly about it, though. Thoughts @yo1995?

PR #701, opened 2025-08-18
[Setup] Add download offline resource view

CalebRas:
## Description

This PR adds a view allowing users to easily download the offline resources for multiple samples at once. This also involved updating `OnDemandResource` to use Observable.

## Linked Issue(s)

- `swift/issues/6797`
- `common-samples/issues/1481`

## How To Test

- Open the About view and tap on "Download Offline Resources".
- Tap on the sample names and "Download All" to ensure the on-demand resources download as expected.
- Use the following script to delete the on-demand resources on all of your booted simulators if needed:

<details><summary>Reset Simulator On-Demand Resources Script</summary>

```bash
xcrun simctl list devices -j | \
jq -r '.devices[][] | select(.state=="Booted") | .udid' | \
while IFS= read -r UDID; do
    echo "Clearing On-Demand Resources for simulator $UDID..."
    rm -rf "$HOME/Library/Developer/CoreSimulator/Devices/$UDID/data/Library/OnDemandResources"
done
```

</details>

**Note**: There is a bug where `NSBundleResourceRequest.conditionallyBeginAccessingResources()` sometimes returns `false` even though the resources were previously downloaded. This was present before these changes and seems to affect simulators.

## Screenshots

|Before|After|
|:-:|:-:|
| Simulator Screenshot - 1 - iPhone 16 Pro - 2025-08-18 at 14 23 26 | Simulator Screenshot - 1 - iPhone 16 Pro - 2025-08-18 at 14 19 31 |
|  | Simulator Screenshot - 1 - iPhone 16 Pro - 2025-08-18 at 14 20 01 |

### DIFF
--- a/Shared/Supporting Files/Models/OnDemandResource.swift
+++ b/Shared/Supporting Files/Models/OnDemandResource.swift
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import SwiftUI
+import Foundation
 
 /// A wrapper class that manages on-demand resource request.
 @MainActor
@@ -62,17 +62,17 @@ final class OnDemandResource {
     func cancel() {
         progress.cancel()
         request.endAccessingResources()
-        withAnimation { requestState = .cancelled }
+        requestState = .cancelled
     }
     
     /// Starts the on-demand resource request.
     func download() async {
         guard isDownloadable else { return }
         
-        withAnimation { requestState = .inProgress }
+        requestState = .inProgress
         do {
             try await request.beginAccessingResources()
-            withAnimation { requestState = .downloaded }
+            requestState = .downloaded
         } catch {
             if (error as NSError).code != NSUserCancelledError {
                 self.error = error


Shared/Supporting Files/Models/OnDemandResource.swift
```diff
@@ -35,54 +35,59 @@ final class OnDemandResource: ObservableObject {
...
-        request.progress
-            .publisher(for: \.fractionCompleted, options: .new)
-            .receive(on: DispatchQueue.main)
-            .map { .inProgress($0) }
-            .sink { [weak self] in self?.requestState = $0 }
-            .store(in: &cancellables)
+    }
+    
+    /// Sets up the `requestState` by checking if the resource is already on device.
+    func setUp() async {
```
philium, line 63: Having an asynchronous method that must be called immediately after creating the object is strong indication that the initializer itself should be asynchronous.

Shared/Supporting Files/Models/OnDemandResource.swift
```diff
@@ -35,54 +35,59 @@ final class OnDemandResource: ObservableObject {
...
-    /// A set of cancellable instances for the request progress subscription.
-    private var cancellables: Set<AnyCancellable> = []
+    /// The name of the sample associated with the resource request.
+    let sampleName: String
     
-    /// Initializes a request with a set of Resource Tags.
-    init(tags: Set<String>) {
-        request = NSBundleResourceRequest(tags: tags)
+    /// Initializes a request with a sample.
+    init(sample: Sample) {
```
philium, line 56: This type was intentionally designed to be sample agnostic. Figuring out the tags to use for a given sample shouldn't be part of its job. Instead, the tags should be passed to it like before and the name should be stored elsewhere, alongside the resource.

Shared/Supporting Files/Models/OnDemandResource.swift
```diff
@@ -35,54 +35,59 @@ final class OnDemandResource: ObservableObject {
...
     func download() async {
-        // Initiates download when it is not being/already downloaded.
-        // Checks if the resource is already on device.
-        let isResourceAvailable = await request.conditionallyBeginAccessingResources()
-        if isResourceAvailable {
+        guard isDownloadable else { return }
+        
+        requestState = .inProgress
+        do {
+            try await request.beginAccessingResources()
```
philium, line 83: It is worth pointing out this, from the [doc](https://developer.apple.com/documentation/foundation/nsbundleresourcerequest/endaccessingresources()#Discussion) for `endAccessingResources()` (emphasis mine):

> Call this method as soon as you have finished using the tags managed by this request. **If needed, this method will be called by the system when the resource request object is deallocated.**

For the resource downloader, this means that any of the downloaded resources could be purged as soon as the downloader dismisses and the resources are deallocated. I'd say that warrants a comment and perhaps even a message in the UI that downloads could be purged by the system at any time.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -28,6 +28,8 @@ struct AboutView: View {
     
     /// A Boolean value indicating whether the API key alert is presented.
     @State private var isAPIKeyAlertPresented = false
+    /// A Boolean value indicating whether the download offline resources cover is presented.
+    @State private var isOfflineResourcesCoverPresented = false
```
philium, line 32: `Cover` is an odd noun for this—SwiftUI calls this it just a modal view in the doc. You could just call it `isResourceDownloaderPresented`.

Shared/Supporting Files/Views/DownloadOfflineResourcesView.swift
```diff
@@ -0,0 +1,172 @@
...
+
+/// A view with controls for downloading the app's on-demand resources.
+struct DownloadOfflineResourcesView: View {
+    /// The action to dismiss the view.
+    @Environment(\.dismiss) private var dismiss
+    
+    /// The view models for the on-demand resource requests.
+    @State private var onDemandResources = SamplesApp.samples.compactMap { sample in
+        sample.hasDependencies ? OnDemandResource(sample: sample) : nil
+    }
```
philium, line 25: This doesn't really need to be a state property for a few reasons. One, the initial value is always the same. Two, computing the initial value isn't very expensive. And three, even if it were expensive, this view is unlikely if not impossible to be re-created by the parent while it is presented.

That being said, my suggestion would be to make this the samples that use on-demand resources and then have a state property for the resources:

```swift
let samples = SamplesApp.samples.filter(\.hasDependencies)

@State private var onDemandResources: [String: OnDemandResource] = [:]
```

And then create the resources in a task modifier:

```swift
.task {
    guard onDemandResources.isEmpty else { return }
    onDemandResources = await withTaskGroup { group in
        for sample in samples {
            group.addTask {
                let resource = await OnDemandResource(tags: sample.odrTags)
                return (sample.name, resource)
            }
        }
        var resources: [String: OnDemandResource] = [:]
        for await (name, resource) in group {
            resources[name] = resource
        }
        return resources
    }
}
```

Shared/Supporting Files/Views/DownloadOfflineResourcesView.swift
```diff
@@ -0,0 +1,172 @@
...
+    
+    /// The current state of the "download all on-demand resources" request.
+    @State private var downloadAllRequestState: OnDemandResource.RequestState?
```
philium, line 28: By making the initializer for `OnDemandResource` asynchronous, the request state would no longer need to be optional.

Shared/Supporting Files/Views/DownloadOfflineResourcesView.swift
```diff
@@ -0,0 +1,172 @@
...
+                    }
+                    .onChange(of: allResourcesAreDownloaded) {
+                        guard allResourcesAreDownloaded else { return }
+                        downloadAllRequestState = .downloaded
+                    }
+                }
+                
+                ForEach(onDemandResources, id: \.sampleName) { resource in
+                    DownloadOnDemandResourceView(resource: resource)
+                }
```
philium, line 68: This should be its own section, right?

```swift
Section {
    List(samples, id: \.name) { sample in
        if let resource = resource(for: sample) {
            DownloadOnDemandResourceView(name: sample.name, resource: resource)
        }
    }
}
```

Shared/Supporting Files/Views/DownloadOfflineResourcesView.swift
```diff
@@ -0,0 +1,172 @@
...
+                    .alert("Cancel Downloads?", isPresented: $confirmCancelAlertIsShowing) {
+                        Button("Resume", role: .cancel, action: {})
+                        
+                        Button("Confirm") {
+                            for resource in onDemandResources where resource.requestState == .inProgress {
+                                resource.cancel()
+                            }
+                            dismiss()
+                        }
+                    }
```
philium, line 90: I'm not sure this alert is entirely necessary. But if it is, here are some suggestions related to the HIG:

- The title shouldn't use title case since it is a question ("Cancel downloads?")
- The button titles should be more related to the question in the title: "Don't Cancel" and "Cancel" would be good

philium (commented):
I've made all the suggestions I gave as I went. I'll create a PR with those changes targeting your PR so you can take a look.

Shared/Supporting Files/Views/DownloadOfflineResourcesView.swift
```diff
@@ -0,0 +1,172 @@
...
+                        if onDemandResources.allSatisfy({ $0.requestState != .inProgress }) {
+                            dismiss()
+                        } else {
+                            confirmCancelAlertIsShowing = true
+                        }
+                    }
+                    .alert("Cancel Downloads?", isPresented: $confirmCancelAlertIsShowing) {
+                        Button("Resume", role: .cancel, action: {})
+                        
+                        Button("Confirm") {
```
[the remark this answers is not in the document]
philium: I'm not sure that cancelling downloads would be considered destructive. No data is lost or destroyed.

Shared/Supporting Files/Models/OnDemandResource.swift
```diff
@@ -12,18 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Combine
-import Foundation
+import SwiftUI
```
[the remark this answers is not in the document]
CalebRas: This was the only way I could figure out to animate the icon transitions. What you said makes sense, though, and I don't think the animations are a requirement, so I just removed them for now. They can be added later if needed.

philium (approved):

CalebRas:
Thank you both for you feedback on this!

PR #706, opened 2025-08-28
[Fix] `Validate utility network topology view` build error

CalebRas:
## Description

This PR fixes a build error in `Validate utility network topology view`, that was caused by a recent `ArcGIS Maps SDK for Swift` API change.

## Linked Issue(s)

- `swift/issues/7240`
- `runtimecore/pull/33545`

## How To Test

Ensure the project builds with the latest `swift-toolkit-daily` and the sample still loads as expected.


rolson:
do we want to also fix in lts.next so that if somebody wants to use 200.8.1 samples with 300 it will compile?

PR #710, opened 2025-10-03
[Fix] Update NavigationSplitView with Toggle Button 

chriswebb09:
Updates the NavigationSplitView to have a full screen button,

### DIFF
--- a/Shared/Supporting Files/Views/ContentView.swift
+++ b/Shared/Supporting Files/Views/ContentView.swift
@@ -23,7 +23,7 @@ struct ContentView: View {
     @Environment(\.requestReviewModel) private var requestReviewModel
     
     /// The visibility of the leading columns in the navigation split view.
-    @State private var columnVisibility: NavigationSplitViewVisibility = .doubleColumn
+    @State private var columnVisibility: NavigationSplitViewVisibility = .all
 
     /// A Boolean value indicating whether to present the about view.
     @State private var isAboutViewPresented = false
--- a/Shared/Supporting Files/Views/ContentView.swift
+++ b/Shared/Supporting Files/Views/ContentView.swift
@@ -24,16 +24,13 @@ struct ContentView: View {
     
     /// The visibility of the leading columns in the navigation split view.
     @State private var columnVisibility: NavigationSplitViewVisibility = .doubleColumn
-    
-//    @State private var fullScreen = false
-    
+
     /// A Boolean value indicating whether to present the about view.
     @State private var isAboutViewPresented = false
     
     var body: some View {
         NavigationSplitView(columnVisibility: $columnVisibility) {
             CategoriesView(columnVisibility: $columnVisibility)
-//                .toolbar(removing: fullScreen ? .sidebarToggle : nil)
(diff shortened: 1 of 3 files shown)


Shared/Supporting Files/Views/ContentView.swift
```diff
@@ -24,7 +24,7 @@ struct ContentView: View {
     
     /// The visibility of the leading columns in the navigation split view.
     @State private var columnVisibility: NavigationSplitViewVisibility = .all
-    
+
```
yo1995, line 27: Please revert the whitespace change.

Shared/Supporting Files/Views/SampleDetailView.swift
```diff
@@ -16,14 +16,20 @@ import SwiftUI
 
 struct SampleDetailView: View {
     /// The sample to display in the view.
-    private let sample: Sample
+    let sample: Sample
```
chriswebb09, line 19: Xcode was throwing 'using variable before it is set' when `isFullScreen` was directly in the init and the workarounds were convoluted. Let me know if the access level of the Sample is okay to be exposed or if there is a better way to handle the init.

Shared/Supporting Files/Views/SampleDetailView.swift
```diff
@@ -101,6 +103,12 @@ struct SampleDetailView: View {
         .navigationTitle(sample.name)
         .navigationBarTitleDisplayMode(.inline)
         .toolbar {
+            ToolbarItem(placement: .topBarLeading) {
+                Button("Full Screen", systemImage: screenExpansionToggleImage) {
+                    isFullScreen.toggle()
+                    screenExpansionToggleImage = isFullScreen ? "arrow.down.forward.and.arrow.up.backward" : "arrow.up.backward.and.arrow.down.forward"
```
yo1995, line 109: A more common pattern is like [this](https://stackoverflow.com/a/72858311/14369688), to use the label parameter and an [`Image(systemName:)` ](https://developer.apple.com/documentation/swiftui/image/init(systemname:)) to toggle the button icon.

chriswebb09: Yeah, I thought about it that way, these strings are absurd long - arrow.down.forward.and.arrow.up.backward, arrow.up.backward.and.arrow.down.forward - so I was thinking this would make it more readable. I don't have a strong opinion either way though.

Shared/Supporting Files/Views/SampleDetailView.swift
```diff
@@ -101,6 +104,15 @@ struct SampleDetailView: View {
         .navigationTitle(sample.name)
         .navigationBarTitleDisplayMode(.inline)
         .toolbar {
+            if ProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 26 && UIDevice.current.userInterfaceIdiom == .pad {
```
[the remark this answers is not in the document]
chriswebb09: Thanks Ting!

CalebRas (approved):

yo1995:
Sorry, I need to check one more thing…

PR #730, opened 2025-11-25
[Fix] Ensure labels update and maintain state instead of resetting 

chriswebb09:
## Description

Fixes issue where labels would not maintain state and would reset to original value after popover was closed.  Previously the selected radius and selected max scale would go back to 0 and 60.

## Linked Issue(s)

- `swift/issues/7561`


dfeinzimer (commented):
It looks like newly set values aren't reflected until the settings are re-opened. Marking the model `@Observable` might be one path forward.

For example, here I set the Cluster Radius to 30, but my choice isn't reflected accurately until I close and re-open:

Image

Samples.xcodeproj/project.pbxproj
```diff
@@ -5061,7 +5055,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				MARKETING_VERSION = 200.8.0;
+				MARKETING_VERSION = 200.8.1;
```
dfeinzimer: Should this be changed in a separate focused PR?

Samples.xcodeproj/project.pbxproj
```diff
@@ -5092,7 +5086,7 @@
...
```
dfeinzimer: Should this be changed in a separate focused PR?

Samples.xcodeproj/project.pbxproj
```diff
@@ -5135,7 +5129,7 @@
 			repositoryURL = "https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 200.8.0;
+				minimumVersion = 200.8.1;
```
dfeinzimer: Should this be changed in a separate focused PR?

Shared/Samples/Configure clusters/ConfigureClustersView.Model.swift
```diff
@@ -13,13 +13,15 @@
 // limitations under the License.
 
 import ArcGIS
+import SwiftUI
```
dfeinzimer: Is this needed?

CalebRas:
> Is the Samples project moving away from the pattern of multiple source files per sample? If not, the bug this is targeting can be resolved while keeping this file around.

I agree with David on this. You should still be able to use `SettingsView` if `Model` is made Observable. Does that not work for some reason?

As David also pointed out, you have `lts.next` changes in this PR, likely because you built your branch off of `lts.next` instead of `v.next`. Please rebase to prevent those commits from going into `v.next`.

PR #733, opened 2025-11-27
[Fix] Fix segment view 

chriswebb09:
## Description

Fixes issue where segment bar did not show up on iPhone

Simulator Screenshot - iPhone 17 Pro - 2025-11-26 at 20 43 35

Simulator Screenshot - iPad (A16) - 2025-12-01 at 08 13 08

## Linked Issue
- `swift/issues/7592`


yo1995 (commented):
Screenshot 2025-12-01 at 10 05 56 AM

This problem seems to only be present on iOS 26.0. 26.1 works fine with `.principal`.

After the change to `.automatic`, it looks a bit off on Mac Catalyst.

I'm slightly leaning towards keeping it principal and ignore it as a one version only iOS bug. What do you think?

PR #743, opened 2026-01-28
[Fix] Ensure labels update and maintain state instead of resetting

chriswebb09:
## Description

Fixes issue where labels would not maintain state and would reset to original value after popover was closed.  Previously the selected radius and selected max scale would go back to 0 and 60.

## Linked Issue(s)

- `swift/issues/7561`

### DIFF
--- a/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
+++ b/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
@@ -17,7 +17,7 @@ import SwiftUI
 extension ConfigureClustersView {
     struct SettingsView: View {
         /// The model for the sample.
-        @ObservedObject var model: Model
+        @Binding var model: Model
         
         /// The action to dismiss the settings sheet.
         @Environment(\.dismiss) private var dismiss: DismissAction
--- a/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
+++ b/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
@@ -25,13 +25,6 @@ extension ConfigureClustersView {
         /// The map view's scale.
         let mapViewScale: Double
         
-        /// The radius of feature clusters selected by the user.
-        @State private var selectedRadius = 60
-        
-        /// The maximum scale of feature clusters selected by the user.
-        @State private var selectedMaxScale = 0
-        
-        
         let scales: [Double] = [0, 1000, 5000, 10000, 50000, 100000, 500000]
         
         var body: some View {
@@ -54,10 +47,6 @@ extension ConfigureClustersView {
                                 Text("\(radius)")
                             }
                         }
-//                        .onChange(of: selectedRadius) {
-//                            model.radius = Double(selectedRadius)
-//                        }
-//                        
                         Picker(
                             "Cluster Max Scale",
                             selection: Binding(
@@ -69,9 +58,7 @@ extension ConfigureClustersView {
                                 Text("\(Int(scale))") // show as integer
                             }
                         }
-//                        .onChange(of: selectedMaxScale) {
-//                            model.maxScale = Double(selectedMaxScale)
-//                        }
+
                         
                         LabeledContent(
                             "Current Map Scale",
--- a/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
+++ b/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
@@ -17,7 +17,7 @@ import SwiftUI
 extension ConfigureClustersView {
     struct SettingsView: View {
         /// The model for the sample.
-        @Binding var model: Model
+        @Bindable var model: Model
         
         /// The action to dismiss the settings sheet.
         @Environment(\.dismiss) private var dismiss: DismissAction
@@ -31,31 +31,19 @@ extension ConfigureClustersView {
             NavigationStack {
                 Form {
                     Section("Cluster Labels Visibility") {
-                        Toggle("Show Labels", isOn: Binding(
-                            get: { model.showsLabels },
-                            set: { model.showsLabels = $0 }
-                        ))
+                        Toggle("Show Labels", isOn: $model.showsLabels)
                             .toggleStyle(.switch)
                     }
                     
                     Section("Clustering Properties") {
-                        Picker("Cluster Radius", selection: Binding(
-                            get: { Int(model.radius) },
-                            set: { model.radius = Double($0) }
-                        )) {
+                        Picker("Cluster Radius", selection: $model.radius) {
                             ForEach([30, 45, 60, 75, 90], id: \.self) { radius in
                                 Text("\(radius)")
                             }
                         }
-                        Picker(
-                            "Cluster Max Scale",
-                            selection: Binding(
-                                get: { model.maxScale },
-                                set: { model.maxScale = $0 }
-                            )
-                        ) {
+                        Picker("Cluster Max Scale", selection: $model.maxScale) {
                             ForEach(scales, id: \.self) { scale in
-                                Text("\(Int(scale))") // show as integer
+                                Text(scale, format: .number)
                             }
                         }
 
--- a/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
+++ b/Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
@@ -31,32 +31,47 @@ extension ConfigureClustersView {
         /// The maximum scale of feature clusters selected by the user.
         @State private var selectedMaxScale = 0
         
+        
+        let scales: [Double] = [0, 1000, 5000, 10000, 50000, 100000, 500000]
+        
         var body: some View {
             NavigationStack {
                 Form {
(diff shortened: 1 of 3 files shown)


Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,7 +17,7 @@ import SwiftUI
 extension ConfigureClustersView {
     struct SettingsView: View {
         /// The model for the sample.
-        @ObservedObject var model: Model
+        @Binding var model: Model
```
CalebRas: You made `Model` Observable, but are still using State variables and onChange modifiers to set its properties, so the bug still exists. Please test before putting up a PR.

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,46 +17,48 @@ import SwiftUI
...
         
         var body: some View {
             NavigationStack {
                 Form {
                     Section("Cluster Labels Visibility") {
-                        Toggle("Show Labels", isOn: $model.showsLabels)
+                        Toggle("Show Labels", isOn: Binding(
+                            get: { model.showsLabels },
+                            set: { model.showsLabels = $0 }
+                        ))
```
CalebRas: Nice approach, but I believe you can reference the model's binding here. (Please correct me if I am wrong.)

```suggestion
                        Toggle("Show Labels", isOn: $model.showsLabels)
```

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,46 +17,48 @@ import SwiftUI
...
                             .toggleStyle(.switch)
                     }
                     
                     Section("Clustering Properties") {
-                        Picker("Cluster Radius", selection: $selectedRadius) {
+                        Picker("Cluster Radius", selection: Binding(
+                            get: { Int(model.radius) },
+                            set: { model.radius = Double($0) }
+                        )) {
```
CalebRas:
```suggestion
                        Picker("Cluster Radius", selection: $model.radius) {
```

Same suggestion as above, but to get this one to work, you'll have to update the picker options. I personally would use a `tag` to do that, but you can also update the list to Doubles as you did with `scales`. Up to you.

Also, I know you didn't touch `Text("\(radius)")`, but it should probably be updated to use FormatStyle. Here is what I am thinking all together:

```swift
Picker("Cluster Radius", selection: $model.radius) {
    ForEach([30, 45, 60, 75, 90], id: \.self) { radius in
        Text(radius, format: .number)
            .tag(radius as Double)
    }
}
```

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,46 +17,48 @@ import SwiftUI
...
-                                Text(("\(scale)"))
+                        Picker(
+                            "Cluster Max Scale",
+                            selection: Binding(
+                                get: { model.maxScale },
+                                set: { model.maxScale = $0 }
+                            )
+                        ) {
+                            ForEach(scales, id: \.self) { scale in
+                                Text("\(Int(scale))") // show as integer
```
CalebRas:
```suggestion
                        Picker("Cluster Max Scale", selection: $model.maxScale) {
                            ForEach(scales, id: \.self) { scale in
                                Text(scale, format: .number)
```

Same suggestions binding and FormatStyle suggestions as above. You could also you `tag(radius as Double)`, but again, up to you.

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,46 +17,48 @@ import SwiftUI
...
                             }
                         }
-                        .onChange(of: selectedMaxScale) {
-                            model.maxScale = Double(selectedMaxScale)
-                        }
+
```
CalebRas: This added line is throwing a linter warning for me. Did you accidentally move it here from between the max scale and radius pickers?

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,46 +17,48 @@ import SwiftUI
 extension ConfigureClustersView {
     struct SettingsView: View {
         /// The model for the sample.
-        @ObservedObject var model: Model
+        @Binding var model: Model
```
CalebRas:
```suggestion
        @Bindable var model: Model
```

This works, but I think the expected way is to using `@Bindable`. See [Migrating from the Observable Object protocol to the Observable macro](https://developer.apple.com/documentation/SwiftUI/Migrating-from-the-observable-object-protocol-to-the-observable-macro) for an example.

Shared/Samples/Configure clusters/ConfigureClustersView.swift
```diff
@@ -58,7 +58,7 @@ struct ConfigureClustersView: View {
                             showsSettings = true
                         }
                         .popover(isPresented: $showsSettings) { [mapViewScale] in
-                            SettingsView(model: model, mapViewScale: mapViewScale)
+                            SettingsView(model: $model, mapViewScale: mapViewScale)
```
CalebRas:
```suggestion
                            SettingsView(model: model, mapViewScale: mapViewScale)
```

This will be needed to build with the `@Bindable` change above.

CalebRas (commented):
Thanks for fixing the bug. Just a few more suggestions:

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
chriswebb09: Yeah, it's accident.

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -40,23 +36,17 @@ extension ConfigureClustersView {
...
-                                Text(("\(scale)"))
+                        Picker("Cluster Max Scale", selection: $model.maxScale) {
+                            ForEach(scales, id: \.self) { scale in
+                                Text(scale, format: .number)
                             }
                         }
-                        .onChange(of: selectedMaxScale) {
-                            model.maxScale = Double(selectedMaxScale)
-                        }
+
```
dfeinzimer:
```suggestion
                        
```

Looks like this lines here are still off.

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -40,23 +36,17 @@ extension ConfigureClustersView {
                     }
                     
                     Section("Clustering Properties") {
-                        Picker("Cluster Radius", selection: $selectedRadius) {
+                        Picker("Cluster Radius", selection: $model.radius) {
                             ForEach([30, 45, 60, 75, 90], id: \.self) { radius in
```
dfeinzimer: In the model, `radius` is a `Double`, so these `Int` options aren't being persisted correctly when chosen.

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -40,23 +36,16 @@ extension ConfigureClustersView {
                     }
                     
                     Section("Clustering Properties") {
-                        Picker("Cluster Radius", selection: $selectedRadius) {
+                        Picker("Cluster Radius", selection: $model.radius) {
```
CalebRas, line 43: Did you ever address the feedback to fix the `Int` to `Double` conversion issue suggested [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/743#discussion_r2742859281) and [here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/743#discussion_r2747521493)? This picker still doesn't work for me on iOS 18.

CalebRas (dismissed):

Shared/Samples/Configure clusters/ConfigureClustersView.SettingsView.swift
```diff
@@ -17,19 +17,17 @@ import SwiftUI
...
         /// The map view's scale.
         let mapViewScale: Double
         
-        /// The radius of feature clusters selected by the user.
-        @State private var selectedRadius = 60
+        let scales: [Double] = [0, 1000, 5000, 10000, 50000, 100000, 500000]
         
-        /// The maximum scale of feature clusters selected by the user.
-        @State private var selectedMaxScale = 0
+        let radii: [Double] = [30, 45, 60, 75, 90]
```
dfeinzimer, line 29:
```suggestion
        /// The radius options for feature clusters.
        let radii: [Double] = [30, 45, 60, 75, 90]
        
        /// The maximum scale options for feature clusters.
        let scales: [Double] = [0, 1000, 5000, 10000, 50000, 100000, 500000]
```

dfeinzimer (commented):
Just one ordering and doc suggestion. Otherwise, it looks good to go!

dfeinzimer (approved):

CalebRas (approved):

PR #754, opened 2026-03-26
[Update] Bump version numbers for 300.0.0

yo1995:
## Description

This PR bumps the samples and toolkit version numbers to 300.0.0 in preparation for the upcoming release. Note that the project currently doesn't build since the 300.0.0 toolkit hasn't been released yet.

## Linked Issue(s)

- `swift/issues/7610`


README.md
```diff
@@ -18,9 +18,9 @@ This repository contains Swift sample code demonstrating the capabilities of the
...
 ## Requirements
 
-* [ArcGIS Maps SDK for Swift](https://developers.arcgis.com/swift/) 200.8 (or newer)
-* [ArcGIS Maps SDK for Swift Toolkit](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit) 200.8 (or newer)
-* Xcode 16.4 (or newer)
+* [ArcGIS Maps SDK for Swift](https://developers.arcgis.com/swift/) 300.0 (or newer)
+* [ArcGIS Maps SDK for Swift Toolkit](https://github.com/Esri/arcgis-maps-sdk-swift-toolkit) 300.0 (or newer)
+* Xcode 26.0 (or newer)
 
 The *ArcGIS Maps SDK for Swift Samples app* has a *Target SDK* version of *17.0*, meaning that it can run on devices with *iOS 17.0* or newer.
```
yo1995: Would you agree if I reword this sentence to 

> The *ArcGIS Maps SDK for Swift Samples app* has a *Minimum SDK* version of *17.0*, meaning that it can run on devices with *iOS 17.0* or newer.

because to my knowledge a "target SDK version" is different from a "min SDK version". In our case, the target SDK version is iOS 26, while min SDK version is iOS 17 (and the "compile SDK version" is iOS 26).

Not sure why I never thought about it before, but good to fix before a major release.

Samples.xcodeproj/xcshareddata/xcschemes/Samples.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1600"
+   LastUpgradeVersion = "2630"
```
yo1995: Image

Project recommended settings applied.

Samples.xcodeproj/project.pbxproj
```diff
@@ -4443,7 +4408,7 @@
 					StyleSymbolsFromMobileStyleFile,
 				);
 				LastSwiftUpdateCheck = 1330;
-				LastUpgradeCheck = 1600;
+				LastUpgradeCheck = 2630;
```
CalebRas: Should this be `2600` if the Xcode version in the README is 26.0? Honestly don't know enough about this to tell...

README.md
CalebRas: That change sounds right to me. Nice catch!

Samples.xcodeproj/project.pbxproj
yo1995: Yeah I've been on the fence on this one… here's my thought process:

The goal is to not let people see any warnings when they first open and build the project.
Our requirement is "Xcode 26.0 (or newer)", meaning that we *require* the users to build the project with at least Xcode 26.0. However, most people would likely to use the latest Xcode from App Store, which at this moment is Xcode 26.4. If we set it to LastUpgradeVersion = "2600", they would see an "update to recommended settings warning"; if we set it to 2640, no matter which Xcode 26 minor version they use, they won't see the warning.

So in short - setting it to 2640, no warnings for any Xcode 26 version; setting it to 2600 is consistent with our system requirements, but people using Xcode 26.1 - 26.4 would see a warning.

README.md
yo1995:
```suggestion
The *ArcGIS Maps SDK for Swift Samples app* has a *Minimum SDK* version of *17.0*, meaning that it can run on devices with *iOS 17.0* or newer.
```

Samples.xcodeproj/project.pbxproj
CalebRas: Cool, thanks for explaining!

CalebRas (approved):

zkline101 (approved):

PR #755, opened 2026-03-27
[Fix] Add in max scale on downloaded to prevent requesting too many tiles. 

chriswebb09:
## Description
This PR fixes an issue in the Generate Offline Maps sample where the request throws error because too many times are requested

## Linked Issue(s)

- `swift/issues/7706`

### DIFF
--- a/Shared/Samples/Generate offline map/GenerateOfflineMapView.swift
+++ b/Shared/Samples/Generate offline map/GenerateOfflineMapView.swift
@@ -159,6 +159,8 @@ private extension GenerateOfflineMapView {
         init() {
             // Initializes the online map.
             onlineMap = Map(item: napervillePortalItem)
+            // Sets the min scale to avoid requesting a huge download.
+            onlineMap.minScale = 1e4
         }
         
         deinit {
@@ -182,7 +184,8 @@ private extension GenerateOfflineMapView {
             let parameters = try await offlineMapTask.makeDefaultGenerateOfflineMapParameters(areaOfInterest: areaOfInterest)
             // Limit the levels of detail to reduce tile count.
             // This prevents requesting too many tiles from the service.
-            parameters.maxScale = 2e4
+            parameters.minScale = 2e4
+            parameters.maxScale = 1e4
             return parameters
         }
         
--- a/Shared/Samples/Generate offline map/GenerateOfflineMapView.swift
+++ b/Shared/Samples/Generate offline map/GenerateOfflineMapView.swift
@@ -160,7 +160,12 @@ private extension GenerateOfflineMapView {
             // Initializes the online map.
             onlineMap = Map(item: napervillePortalItem)
             // Sets the min scale to avoid requesting a huge download.
+            // Use a more conservative scale on iPad due to larger screen size.
+            #if os(iOS)
+            onlineMap.minScale = UIDevice.current.userInterfaceIdiom == .pad ? 5e4 : 1e4
+            #else
             onlineMap.minScale = 1e4
+            #endif
         }
         
         deinit {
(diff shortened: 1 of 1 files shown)


Shared/Samples/Generate offline map/GenerateOfflineMapView.swift
```diff
@@ -159,8 +159,6 @@ private extension GenerateOfflineMapView {
         init() {
             // Initializes the online map.
             onlineMap = Map(item: napervillePortalItem)
-            // Sets the min scale to avoid requesting a huge download.
-            onlineMap.minScale = 1e4
```
chriswebb09, line 163: @njarecha there was a min scale set initially. I tested with this in and out after the making the max scale change. The min scale prevents the map from zooming out too much doesn't prevent the time error or the requesting too many tiles. I will double check on iOS to make sure this doesn't have any secondary issues.

CalebRas:
Could you give steps on how to reproduce the original bug either in the issue or PR, so we can test please?

chriswebb09:
@CalebRas on an iPad zoom out to the largest area possible and then generate an offline map. 
Screenshot 2026-03-30 at 2 36 00 PM

It should consistently fail like so: 
IMG_0014

Shared/Samples/Generate offline map/GenerateOfflineMapView.swift
```diff
@@ -181,7 +179,11 @@ private extension GenerateOfflineMapView {
         /// - Returns: A `GenerateOfflineMapParameters` if there are no errors.
         private func makeGenerateOfflineMapParameters(areaOfInterest: Envelope) async throws -> GenerateOfflineMapParameters {
             // Returns the default parameters for the offline map task.
-            return try await offlineMapTask.makeDefaultGenerateOfflineMapParameters(areaOfInterest: areaOfInterest)
+            let parameters = try await offlineMapTask.makeDefaultGenerateOfflineMapParameters(areaOfInterest: areaOfInterest)
+            // Limit the levels of detail to reduce tile count.
+            // This prevents requesting too many tiles from the service.
+            parameters.maxScale = 2e4
```
CalebRas: Thanks for the instructions! A couple of things I noticed:

1. If I zoom out on my iPad further enough, it still fails too generate the offline map. Is that expected?

|Extent|Result|
|:-:|:-:|
| Image | Image |

2. If I just open the sample and press "Generate Offline Map" without adjusting the map at all, the basemap now generates at a much lower resolution. That isn't great since it is the most likely workflow. It is it possible to fix the bug without this happening?

|Before|After|
|:-:|:-:|
| Image | Image |

chriswebb09: 1. I'm taking a look at that one. 
2. I will see if it can be higher resolution.

rolson: Can we set a maxExtent on the map? Or even a minScale on the map? That would prevent zooming out too far. We could set the maxExtent to the extent of the feature data in the map.

chriswebb09: IMG_0017

With the new changes.

CalebRas: Issue 1 is resolved now, but the resolution still doesn't look great if you generate with the sample's default scale. My vote is we avoid messing with parameters and instead just set the map's minScale to something like 5000 prevent zooming out very far, which I believe is what Ryan is suggesting.

njarecha: Are we just trying to working around the issue by setting min and max scale? If user decide to change the values, will they run into the same issue?

CalebRas (dismissed):
Looks good and works for me on all platforms. Thanks for making those changes, Chris!

CalebRas (approved):
Looks good!

njarecha (approved):
Looks good

PR #752, opened 2026-03-24
[New] (300.0) Apply map algebra

yo1995:
## Description

This PR implements `Apply map algebra` in `Analysis` category.
[URL to README](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/Ting/ApplyMapAlgebra/Shared/Samples/Apply%20map%20algebra/README.md)

## Linked Issue(s)

- `swift/issues/7890`

## How To Test

- [ ] The ODR downloads correctly on iOS sim, iOS device.
- [ ] On Mac Catalyst, the sample runs fine.

## To Discuss

The sample requires an analysis extension license key, so applicable changes are made to accommodate that. Edit your ".secrets" file with these keys to make the app license correctly. Please ping me if you cannot find the keys.

- ADVANCED_EDITING_EXTENSION_LICENSE_KEY
- ANALYSIS_EXTENSION_LICENSE_KEY

PR #604, opened 2025-05-08
[Fix] Replace single raster tiff with a folder

yo1995:
## Description

This PR fixes `Apply hillshade renderer to raster` #592 in `Visualization` category. It replaces the single `srtm.tiff` file and put it in a folder, to ensure the build process passes. For some reason, single file raster doesn't play along very well with our samples…

## How To Test

1. Clean the portal data folder. Either clear all, or delete the "Portal Data/9c802ab0c93a4e45a16311c666fe81c1" and "Portal Data/ae9739163a76437ea02482e1a807b806" folder.
2. Build and run the app. There shouldn't be an error, whereas on v.next there is a build error.

PR #757, opened 2026-03-27
[Fix] Enclose the tiff file in a folder when the ODR is a standalone raster

yo1995:
## Description

This PR fixes a recent daily build error introduced by #752 . The sample needed an "arran.tif" raster file via ODR, that was incorrectly handled by Xcode, causing a build error of "Build input file cannot be found".

This type of problem affects all tiff files.
- To mitigate the problem specifically for a standalone tiff file, I added a logic to the download script, to enclose the tiff file in a folder when it is downloaded from AGOL.
- For rasters that contain more than just 1 tiff file, because typically they already come as a folder, I didn't make any change to handle them.
    - For instance, "Shasta.tif" from "Add raster from file" sample is already enclosed in a folder, so its folder isn't affected by this fix.

Upon a quick search, the only other same case was "srtm.tiff" used by "Apply hillshade renderer to raster" sample, which was fixed in #604. I've reverted that fix.

### Backstory

<details><summary>More backstory here</summary>

This dates back all the way to https://github.com/Esri/arcgis-runtime-samples-ios/pull/1036 . When Xcode 10 deprecated the legacy build system, there were some unclear changes to how Xcode handle the files that are tracked by the xcodeproj, but aren't part of a build phase's input/output. Later I discovered this undisclosed Xcode behavior that, for *certain* file types, Xcode reports the following error, even if they are added the same way as other files

> [!CAUTION]
error: Build input file cannot be found: 'path_to_build/arcgis-maps-sdk-swift-samples/Portal Data/aa97788593e34a32bcaae33947fdc271/arran.tif'. Did you forget to declare this file as an output of a script phase or custom build rule which produces it? (in target 'Samples' from project 'Samples')

So far, file types that are known to be affected are: .tif/.tiff, .plist

</details>

## Linked Issue(s)

Slack chat reported by swift-devops.

## How To Test

1. Have a clean build of this branch. Either download a zip archive of the branch, or clear your portal data folder before starting the test
2. Build the project. No download error should be reported
3. Run the app, open "Apply hillshade renderer to raster" and "Apply map algebra" samples.
4. The ODR progress bar shows for both samples, and the samples run correctly.


Samples.xcodeproj/project.pbxproj
```diff
@@ -2013,17 +2021,14 @@
 		00CCB8A6285D059300BBAB70 /* Portal Data */ = {
 			isa = PBXGroup;
 			children = (
-				9520B26A2E13555800B3BEF9 /* 3af5cfec0fd24dac8d88aea679027cb9 */,
```
yo1995, line 2016: I sorted the groups by name in the "Portal Data" group.

CalebRas (approved):
Looks good and builds for me

yo1995:
Will merge it ahead of daily build to see if it passes.

PR #769, opened 2026-04-10
[Fix] Offline geocoding

chriswebb09:
## Description 
This PR is to help fix the offline geocoding issue by making the error visible to the user and forcing the initializer to fail on the locator task instead of making it optional. This should be verified through TestFlight because the failure is not visible in local development environment.

### DIFF
--- a/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
+++ b/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import ArcGIS
+import Combine
 import UIKit
 
 extension GeocodeOfflineView {
--- a/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
+++ b/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
@@ -49,7 +49,7 @@ extension GeocodeOfflineView {
             return Graphic(symbol: markerSymbol)
         }()
         
-        /// The locator task used to preform the geocode operations, loaded from a local file.
+        /// The locator task used to perform the geocode operations, loaded from a local file.
         private let locatorTask = LocatorTask(url: .sanDiegoStreetAddressLocator)
         
         /// The placement of the callout on the map.
--- a/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
+++ b/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
@@ -25,7 +25,7 @@ extension GeocodeOfflineView {
         /// A map with a tiled layer of the streets in San Diego, CA, USA.
         let map = {
             // Create a tiled layer using the local tile package.
-            let tileCache = TileCache(fileURL: .streetMap)
+            let tileCache = TileCache(fileURL: .sanDiegoStreetAddressLocator)
             let tiledLayer = ArcGISTiledLayer(tileCache: tileCache)
             
             // Create a map with the tiled layer as base layer.
@@ -71,8 +71,6 @@ extension GeocodeOfflineView {
         /// - Parameter address: The given text address to geocode.
         /// - Returns: The extent of the result's display location.
         func geocodeSearch(address: String) async -> Envelope? {
-            guard let locatorTask else { return nil }
-            
             // Create geocode parameters.
             let geocodeParameters = GeocodeParameters()
             geocodeParameters.addResultAttributeName("Match_addr")
@@ -99,8 +97,6 @@ extension GeocodeOfflineView {
         /// Reverse geocodes a given location and adds a marker with the corresponding address at the result's location.
         /// - Parameter mapPoint: The point on the map to reverse geocode.
         func reverseGeocode(mapPoint: Point) async {
-            guard let locatorTask else { return }
-            
             //  Normalized the map point.
             guard let normalizedPoint = GeometryEngine.normalizeCentralMeridian(
                 of: mapPoint
@@ -173,4 +169,3 @@ private extension URL {
         )!
     }
 }
-
--- a/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
+++ b/Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
@@ -169,6 +169,7 @@ private extension URL {
         )!
     }
     
+    /// A URL to the local tile package of the streets in San Diego, CA, USA.
     static var streetMap: Self {
         Bundle.main.url(forResource: "streetmap_SD", withExtension: "tpkx")!
     }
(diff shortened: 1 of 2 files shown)


Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -98,9 +96,7 @@ extension GeocodeOfflineView {
         
         /// Reverse geocodes a given location and adds a marker with the corresponding address at the result's location.
         /// - Parameter mapPoint: The point on the map to reverse geocode.
-        func reverseGeocode(mapPoint: Point) async {
-            guard let locatorTask else { return }
-            
+        func reverseGeocode(mapPoint: Point) async throws {
```
yo1995, line 98: Please make this method actually throws by changing `try?` to `try`

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -71,8 +71,6 @@ extension GeocodeOfflineView {
         /// - Parameter address: The given text address to geocode.
         /// - Returns: The extent of the result's display location.
         func geocodeSearch(address: String) async -> Envelope? {
```
yo1995: I think this method should receive the same throwing treatment, as it also uses a `locatorTask` and get `geocodeResults`.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
dfeinzimer, line 16: While we're in this file, it seems like `Combine` is not needed.

Shared/Samples/Geocode offline/GeocodeOfflineView.swift
```diff
@@ -24,7 +24,7 @@ struct GeocodeOfflineView: View {
         center: Point(x: -13_042_250, y: 3_857_970, spatialReference: .webMercator),
         scale: 2e4
     )
-    
+
```
dfeinzimer: Please keep the consistent whitespacing.


```suggestion
    
```

dfeinzimer (commented):
Just a couple things while you're also working on Ting's feedback.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
chriswebb09: But even if it throws a non option, it's still geocodeResults.first array element which can't be guaranteed right?  If we make the Envelope non optional, what should the fallback be if the array is empty?

yo1995: Wasn't talking about optionality of the method. My comment was about making the method throwing because it is calling a throwing method inside.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
yo1995: `Combine` is _technically_ used, as [`ObservableObject`](https://developer.apple.com/documentation/combine/observableobject) is a `Combine` type.

For the samples project, because we haven't enabled [`MemberImportVisibility`](https://docs.swift.org/compiler/documentation/diagnostics/member-import-visibility/) upcoming feature yet, and `ArcGIS` is indirectly importing `Combine` (well, before we gradually phase it out for Swift Concurrency), removing Combine from imports won't produce a warning. However, eventually we want to enable `MemberImportVisibility`, so it is good to keep this import, until we refactor `ObservableObject` with `Observable`.

dfeinzimer: TIL, thanks Ting.

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
chriswebb09: Gotcha

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import ArcGIS
```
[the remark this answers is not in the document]
yo1995: As we discussed, please add back `import Combine`

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
yo1995:
```swift
            let geocodeResults = try await locatorTask.geocode(
```

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -83,7 +80,7 @@ extension GeocodeOfflineView {
                 forSearchText: address,
                 using: geocodeParameters
             )
-            
+        
             if let result = geocodeResults?.first,
```
yo1995:
```suggestion
            if let result = geocodeResults.first,
```

Shared/Samples/Geocode offline/GeocodeOfflineView.Model.swift
```diff
@@ -70,26 +70,23 @@ extension GeocodeOfflineView {
...
             
             // Perform geocode using the locator task with the text address and parameters.
-            let geocodeResults = try? await locatorTask.geocode(
+            let geocodeResults = try await locatorTask.geocode(
                 forSearchText: address,
                 using: geocodeParameters
             )
-            
-            if let result = geocodeResults?.first,
+
```
yo1995:
```suggestion
            
```
Please reindent to fix whitespaces

Shared/Samples/Geocode offline/GeocodeOfflineView.swift
yo1995, line 1: A comment on the error handling that doesn't need to be addressed.

Me personally feel like the error shouldn't be passed down to the child view via a binding.
Because the error in the main view is from `geocodeSearch` (text -> location), while the error in the child view `GeocodeMapView` is from `reverseGeocode` (location -> text). While the errors thrown by these 2 methods are likely both to be `GeocodeError` type, these errors originated from 2 different view interactions - one is from tapping on the map view, and the other from search bar or drop down text list.
Therefore, I feel like the views where the geocode/reverseGeocode operations were kicked off should be responsible to show the error alert.

However, it is just a design choice so I'll leave it up to you.

yo1995 (dismissed):

Shared/Samples/Geocode offline/GeocodeOfflineView.swift
chriswebb09: It makes sense. I've added it in.

yo1995 (approved):

dfeinzimer (approved):

PR #772, opened 2026-04-13
[Setup] Add button for verifying API key expiration date for the release

yo1995:
## Description

This PR adds a button to verify the API key's expiration date, by making a request to the ArcGIS REST endpoint, when the project is built in Debug configuration.

The non-expiring legacy keys would be retired in July 2026. As part of our future release checklist, we need to verify that we are using the latest API key for the daily builds before a release. This button would make verification easier.

Similar to #673 

## Linked Issue(s)

- `devops/issues/5590`

## How To Test

Build the app in debug vs release (scheme > run > info > release), and see if the button show up in debug and doesn't in release. Press the button to see if an alert shows the correct expiration date of the key you use.

## To Discuss

~~I went for the simplest way to verify the API key in use by printing it out. If the public release is handled correctly, it should not present any risk.~~

~~Edit: after consulting some security guidelines, I decided to use input verify partial key to reduce security risk.~~

Edit 2: after discussing with Nimesh, I decided to compare expiration date instead of the key itself.

<details><summary>Other alternatives considered:</summary>

- Print a truncated key for less risk of key exposure (like core's `truncate_API_key` method). Since this button will be stripped out in release config, I figured it is not necessary to truncate.
- Show it in an alert - could have done that, but wanted to keep it simple.
- Compare hash from an input - this would be the least risky, but would require a few additional steps

</details>

Please share your thoughts on which way do you prefer.

### DIFF
--- a/Shared/Supporting Files/Views/AboutView.swift
+++ b/Shared/Supporting Files/Views/AboutView.swift
@@ -222,10 +222,11 @@ private extension AboutView {
             // API key access token. The response contains an `appInfo` object,
             // which has the expiration date of the API key.
             // https://developers.arcgis.com/rest/users-groups-and-items/self/
-            let (data, _) = try await ArcGISEnvironment.urlSession.data(
-                from: URL(string: "https://www.arcgis.com/sharing/rest/Community/self")!,
-                queryParameters: ["f": "json", "token": apiKey.rawValue]
+            var request = URLRequest(
+                url: URL(string: "https://www.arcgis.com/sharing/rest/Community/self?f=json")!
             )
+            request.setValue("Bearer \(apiKey.rawValue)", forHTTPHeaderField: "X-Esri-Authorization")
+            let (data, _) = try await ArcGISEnvironment.urlSession.data(for: request)
             
             let decoder = JSONDecoder()
             decoder.dateDecodingStrategy = .millisecondsSince1970
--- a/Shared/Supporting Files/Views/AboutView.swift
+++ b/Shared/Supporting Files/Views/AboutView.swift
@@ -177,7 +177,14 @@ private extension AboutView {
                 switch apiKeyExpirationDate {
                 case .success(let expirationDate):
                     // Shows the expiration date of the API key in-use.
-                    Text(expirationDate, format: .dateTime.year().month().day())
+                    if let monthsRemaining = Calendar.current.dateComponents([.month], from: .now, to: expirationDate).month,
+                       monthsRemaining <= 4 {
+                        // Reminder when the API key is expiring within 4 months,
+                        // which is the typical release cycle.
+                        Text("\(expirationDate, format: .dateTime.year().month().day()). Expires in ^[\(monthsRemaining) month](inflect: true).")
+                    } else {
+                        Text(expirationDate, format: .dateTime.year().month().day())
+                    }
                 case .failure(let error):
                     // Shows the error when fail to get the expiration date.
                     switch error {
--- a/Shared/Supporting Files/Views/AboutView.swift
+++ b/Shared/Supporting Files/Views/AboutView.swift
@@ -29,14 +29,16 @@ struct AboutView: View {
     /// A Boolean value indicating whether the download offline resources cover is presented.
     @State private var isResourceDownloaderPresented = false
     /// The API key entered in the alert.
-    @State private var apiKey = ""
+    @State private var apiKeyInput = ""
     
     // MARK: Debug
     
     /// A Boolean value indicating whether the API key alert is presented.
     @State private var isAPIKeyAlertPresented = false
-    /// A Boolean value indicating whether the API key verification alert is presented.
-    @State private var isAPIKeyVerificationAlertPresented = false
+    /// A Boolean value indicating whether the API key expiration date alert is presented.
+    @State private var isExpirationDateAlertPresented = false
+    /// The expiration date of the API key in-use.
+    @State private var apiKeyExpirationDate: Date?
     
     var body: some View {
         NavigationStack {
@@ -124,7 +126,7 @@ private extension URL {
     static let writeReview = URL(string: "https://apps.apple.com/app/id1630449018?action=write-review")!
 }
 
-extension AboutView {
+private extension AboutView {
     /// A section for debugging purposes.
     var debugSection: some View {
         Section {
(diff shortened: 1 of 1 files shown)


CalebRas (dismissed):

yo1995:
Screenshot 2026-04-14 at 9 45 50 AM

After a second thought, I'll probably change it to either input or compare suffix. To verify a key is the latest, the workflow would be

- Go to the Swift Sample Viewer Release API key page
- Click "Show partial API key 1" to reveal the last few chars of the key
- Compare it with the key in the sample viewer

So there isn't really a need to show the full key. Let me consult Nimesh or security council for some advice…

yo1995:
After discussing with Nimesh, he suggested the safer way is to verify the expiration date, instead of the key itself, since what we really care about is when the key in-use would expire.

He further suggested to make a request to check the expiration date of an API key - so there is no exposure of any part of the key during this process. I'll apply these changes and request another review later.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +126,78 @@ private extension URL {
...
                     .textInputAutocapitalization(.never)
                     .autocorrectionDisabled()
                     .textEditorStyle(.plain)
                 Button("Cancel", role: .cancel) {}
                 Button("Submit") {
-                    ArcGISEnvironment.apiKey = APIKey(apiKey)!
+                    ArcGISEnvironment.apiKey = APIKey(apiKeyInput)!
                 }
-                .disabled(apiKey.isEmpty)
+                .disabled(apiKeyInput.isEmpty)
```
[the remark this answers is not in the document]
yo1995: Only empty string will return `nil`, so no risk crashing.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +126,78 @@ private extension URL {
...
                 }
             }
+            
+            Button("Verify Expiration Date") {
+                // An alert shows the expiration date for the API key in-use.
+                // Compare it with the date in the Swift Sample Viewer Release
+                // portal item.
+                Task {
+                    apiKeyExpirationDate = try await getAPIKeyExpirationDate()
+                    isExpirationDateAlertPresented = true
```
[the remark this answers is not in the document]
yo1995: I don't care about the error in this response. This is for debugging and if I see the wrong date, I know something is wrong.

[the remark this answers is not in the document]
yo1995: I don't care about the error in this response. This is for debugging and if I see the wrong date, I know something is wrong.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +126,78 @@ private extension URL {
...
+        
+        // An endpoint to get the current authenticated user identified by the
+        // API key access token. The response contains an `appInfo` object,
+        // which has the expiration date of the API key.
+        // https://developers.arcgis.com/rest/users-groups-and-items/self/
+        let (data, _) = try await ArcGISEnvironment.urlSession.data(
+            from: URL(string: "https://www.arcgis.com/sharing/rest/Community/self")!,
+            queryParameters: ["f": "json", "token": apiKey.rawValue]
+        )
+
```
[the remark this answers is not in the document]
yo1995: Well, urlencoded doesn't really make it more secure. Also, all the requests nowadays are already sending the token with GET query parameters.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -19,19 +19,26 @@ struct AboutView: View {
...
+    @State private var apiKeyInput = ""
+    
+    // MARK: Debug
+    
+    /// A Boolean value indicating whether the API key alert is presented.
+    @State private var isAPIKeyAlertPresented = false
+    /// A Boolean value indicating whether the API key expiration date alert is presented.
+    @State private var isExpirationDateAlertPresented = false
+    /// The expiration date of the API key in-use.
+    @State private var apiKeyExpirationDate: Date?
```
CalebRas: It seems like these should be moved into their own subview for better separation of concerns, and so that it can also get excluded in Release. Something like:

```swift
#if DEBUG
private struct DebugAboutView: View {
    /// The API key entered in the alert.
    @State private var apiKeyInput = ""
    /// A Boolean value indicating whether the API key alert is presented.
    @State private var isAPIKeyAlertPresented = false
    /// A Boolean value indicating whether the API key expiration date alert is presented.
    @State private var isExpirationDateAlertPresented = false
    /// The expiration date of the API key in-use.
    @State private var apiKeyExpirationDate: Date?
    
    var body: some View {
        ...
    }
}
#endif
```

That isn't super related to this PR, though, so I think it is okay if you want to leave it as is. If you don't refactor, should `apiKeyInput` be moved below `MARK: Debug`?

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +126,78 @@ private extension URL {
...
+        // which has the expiration date of the API key.
+        // https://developers.arcgis.com/rest/users-groups-and-items/self/
+        let (data, _) = try await ArcGISEnvironment.urlSession.data(
+            from: URL(string: "https://www.arcgis.com/sharing/rest/Community/self")!,
+            queryParameters: ["f": "json", "token": apiKey.rawValue]
+        )
+        
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .millisecondsSince1970
+        let jsonResponse = try decoder.decode(APIResponse.self, from: data)
```
CalebRas: I get this error when trying it out. Is that expected?

```
DecodingError.keyNotFound: Key 'expirationDate' not found in keyed decoding container. Path: appInfo. Debug description: No value associated with key CodingKeys(stringValue: "expirationDate", intValue: nil) ("expirationDate").
```

Shared/Supporting Files/Views/AboutView.swift
CalebRas:
```suggestion
    @State private var apiKeyExpirationDate: Result<Date?, any Error>?
```

Follow up on the last comment: Consider storing a result here and displaying errors thrown from `getAPIKeyExpirationDate()` in the expiration date alert for better debugging. Up to you, though.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +126,78 @@ private extension URL {
...
                 }
-                .disabled(apiKey.isEmpty)
+                .disabled(apiKeyInput.isEmpty)
                 Button("Reset") {
                     ArcGISEnvironment.apiKey = .iOS
+                    apiKeyInput.removeAll()
                 }
             }
+            
+            Button("Verify Expiration Date") {
```
CalebRas:
```suggestion
            Button("View API Key Expiration Date") {
```

Maybe?

CalebRas: Should this button be disabled while `getAPIKeyExpirationDate()` is running to prevent button spamming? Not sure how important that is since this is not release code, so up to you.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +126,78 @@ private extension URL {
...
+                    apiKeyExpirationDate = try await getAPIKeyExpirationDate()
+                    isExpirationDateAlertPresented = true
+                }
+            }
+            .alert("API Key Expiration Date", isPresented: $isExpirationDateAlertPresented) {
+            } message: {
+                Text(
+                    apiKeyExpirationDate == nil
+                    ? "Failed to get expiration date"
+                    : apiKeyExpirationDate!.formatted(date: .abbreviated, time: .omitted)
```
CalebRas:
```suggestion
                    apiKeyExpirationDate.map { $0.formatted(date: .abbreviated, time: .omitted) }
                    ?? "API key is not set"
```

Consider using map to prevent force unwrapping.

Shared/Supporting Files/Views/AboutView.swift
yo1995: Good call. I'll move them to a separate view.

Shared/Supporting Files/Views/AboutView.swift
philium: I suggest that an even better way would is to conditionalize the text view:

```swift
if let apiKeyExpirationDate {
    Text(apiKeyExpirationDate, format: Date.FormatStyle(date: .abbreviated, time: .omitted))
} else {
    Text("API key is not set")
}
```

Shared/Supporting Files/Views/AboutView.swift
yo1995: I'll add some error handling for legacy key.

philium:
Keep in mind that any user-displayable text that is computed ahead of time (especially formatted data!) can become stale. It is always better to compute the text as needed based on the current state of the view. I strongly suggest making `apiKeyExpirationDateMessage` a computed property rather than computing it up front and then caching it.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +119,110 @@ private extension URL {
...
+                apiKeyExpirationDateMessage.removeAll()
+                verifyingAPIKey = true
+            }
+            .disabled(verifyingAPIKey)
+            .task(id: verifyingAPIKey) {
+                guard verifyingAPIKey else { return }
+                defer { verifyingAPIKey = false }
+                do {
+                    let apiKeyExpirationDate = try await getAPIKeyExpirationDate()
+                    apiKeyExpirationDateMessage = apiKeyExpirationDate.formatted(date: .abbreviated, time: .omitted)
```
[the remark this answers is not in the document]
yo1995: I think it would be unlikely that the device to verify the API key to be in a different timezone than the author who created the key? As all Swift team members are in Redlands… 🤔

CalebRas (dismissed):

yo1995:
I agree with you Phil. Localized and formatted strings should always be computed just-in-time. I've applied the changes.

IMHO the code for nested conditional views look a bit messy. I guess that's why I tend to go down the "string-first" path.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +119,120 @@ private extension URL {
...
+        
+        /// Gets the expiration date of the API key in-use by making a request to
+        /// the ArcGIS REST API.
+        /// - Returns: The expiration date of the API key if available.
+        func getAPIKeyExpirationDate() async throws -> Date {
+            guard let apiKey = ArcGISEnvironment.apiKey else {
+                throw ArcGISAuthenticationError.invalidAPIKey
+            }
+            
+            struct APIResponse: Decodable {
```
njarecha: This is a very generic name. I would suggest changing it to something like, `APIKeyInfoResponse`.

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +119,120 @@ private extension URL {
...
+            }
+            
+            // An endpoint to get the current authenticated user identified by the
+            // API key access token. The response contains an `appInfo` object,
+            // which has the expiration date of the API key.
+            // https://developers.arcgis.com/rest/users-groups-and-items/self/
+            let (data, _) = try await ArcGISEnvironment.urlSession.data(
+                from: URL(string: "https://www.arcgis.com/sharing/rest/Community/self")!,
+                queryParameters: ["f": "json", "token": apiKey.rawValue]
+            )
```
njarecha: Send token in header which is recommended when possible.
```suggestion
            var request = URLRequest(url: URL(string: "https://www.arcgis.com/sharing/rest/Community/self?f=json")!)
            request.setValue("Bearer \(apiKey.rawValue)", forHTTPHeaderField: "X-Esri-Authorization")
            let (data, _) = try await ArcGISEnvironment.urlSession.data(for: request)
```

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -83,7 +79,11 @@ struct AboutView: View {
                 }
 #endif
 #if DEBUG
-                debugSection
+                Section {
+                    DebugView()
```
njarecha, line 83: If you use logger to check the request logs, it ends up sending 4 requests. We should find a way to prevent that.

```
            let logger = ConsoleNetworkLogger(
                requestOptions: [.method, .body, .headers],
                responseOptions: [.data, .headers]
            )
            logger.startLogging()
```

Shared/Supporting Files/Views/AboutView.swift
```diff
@@ -119,29 +119,120 @@ private extension URL {
...
+            }
+            .disabled(verifyingAPIKey)
+            .task(id: verifyingAPIKey) {
+                guard verifyingAPIKey else { return }
+                defer { verifyingAPIKey = false }
+                
+                apiKeyExpirationDate = await Result { try await getAPIKeyExpirationDate() }
+                isExpirationDateAlertPresented = true
+            }
+            .alert(
```
njarecha, line 168: It will be a good visual clue if the alert indicates that the expiration duration is less than few months (5-6) from current date. This may warrant generating new API key.

njarecha (changes requested):
Few suggestions

Shared/Supporting Files/Views/AboutView.swift
yo1995: I added a reminder for key expiring within 4 months (typical release cycle interval)

Shared/Supporting Files/Views/AboutView.swift
yo1995: With the latest changes I only see 1 request in the console 🤔 can you try again? Thanks

njarecha (approved):
Looks good

CalebRas (approved):

pgruenler (approved):

yo1995:
Thanks for the feedback!

PR #786, opened 2026-06-15
[New] Sample: Navigate map view and identify features with keyboard

chriswebb09:
## Description

This PR implements `Navigate map view and identify features with keyboard` in `Accessibility` category.
URL to README: 

[Here](https://github.com/Esri/arcgis-maps-sdk-swift-samples/blob/63ebd9a7625140e74d4ad70920abb271f33566ef/Shared/Samples/Navigate%20map%20view%20and%20identify%20features%20with%20keyboard/README.md)

## Linked Issue(s)

- `swift/issues/8255`

## Screenshots

navigate-map-view-and-identify-features-with-keyboard


Shared/Samples/Navigate map view and identify features with keyboard/navigate-map-view-and-identify-features-with-keyboard.png
yo1995, line 1: Please capture the screenshot in Light appearance (Command + shift + A)

Shared/Samples/Navigate map view and identify features with keyboard/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Use keyboard navigation to pan and zoom the map until restaurants appear inside the rectangle. Restaurants inside the rectangle are selected and labeled in reading order, from top-to-bottom and left-to-right.
+
+Press number keys `1` through `9` to show details for the matching restaurant. Press `Esc` to dismiss the callout and show the selection rectangle again.
+
+If more than nine restaurants are inside the rectangle, an overflow message is shown. Only the first nine restaurants can be identified with number keys.
+
+## How it works
+
+1. Create a `Map` with an `ArcGISLightGray` basemap centered on Redlands.
```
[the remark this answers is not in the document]
yo1995: This is a valid suggestion, good to apply.
```suggestion
1. Create a `Map` with an `arcGISLightGray` basemap style centered on Redlands.
```

Shared/Samples/Navigate map view and identify features with keyboard/README.md
yo1995, line 1: Can you double check if you have the latest version of the design README? For instance, when I check the README from the designs folder, the About the data section is

> This sample uses a [Redlands restaurants](https://www.arcgis.com/home/item.html?id=46119989eccd46a58b8f3d7aedadeb90) feature layer covering food establishments in Redlands, California. Each feature represents a single restaurant.

Also, the Additional information section is missing

yo1995 (changes requested):

yo1995 (commented):
I talked to Phil R and he'll review this PR instead of me.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import Foundation
+import SwiftUI
+import UIKit.UIColor
```
philium: There's really [no functional difference](https://nshipster.com/import/#importing-a-submodule) between importing this and importing UIKit:

> In practice, isolating imported declarations and submodules doesn’t confer any real benefit beyond signaling programmer intent. Your code won’t compile any faster doing it this way. And since most submodules seem to re-import their umbrella header, this approach won’t do anything to reduce noise in autocomplete lists.

That's why by and large, we've done away with importing submodules.


```suggestion
import UIKit
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+
+extension NavigateMapAndIdentifyFeaturesWithKeyboardView {
+    /// The model for the sample.
+    @MainActor
+    @Observable
+    final class Model {
+        /// The maximum number of features that can be identified with number keys.
+        private static let maximumNumberedFeatures = 9
```
philium: If you want to, you could put this on `Int`:

```swift
private extension Int {
    /// The maximum number of features that can be identified with number keys.
    static let maximumNumberedFeatures = 9
}
```

And then use it like so, without needing to reference `Self`:

```swift
hasMoreThanNineSelectedFeatures = orderedFeatures.count > .maximumNumberedFeatures
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+        
+        /// Colors used for the marker, selection halo, and label.
+        private static let markerFillColor = UIColor(
+            red: 11 / 255,
+            green: 79 / 255,
+            blue: 138 / 255,
+            alpha: 1
+        )
+        
+        static let selectionHaloColor = Color(
```
philium: Since this is not used by the model, I highly suggest moving it to the view where it is used.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+                labelSymbol.offsetY = -14
+                labelOverlay.addGraphic(
+                    Graphic(geometry: orderedFeature.anchor, symbol: labelSymbol)
+                )
+                numberedFeatures.append(orderedFeature.feature)
+            }
+        }
+        
+        /// The restaurant marker symbol.
+        private static var restaurantSymbol: SimpleMarkerSymbol {
```
philium: Since this is only used in one place, consider moving it there. Otherwise, this should really be a method since it returns a new value each time and the returned value has mutable state.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+        
+        private static let labelTextColor = UIColor(
+            red: 31 / 255,
+            green: 35 / 255,
+            blue: 40 / 255,
+            alpha: 1
+        )
+        
+        /// The map displayed in the map view.
+        let map: Map
```
philium: As far as I can tell, the map is only used by the model to add the operational layer to it. Instead, you could put the map on the view and then add the restaurants layer when the view appears:

```swift
.onAppear {
    guard map.operationalLayers.isEmpty else { return }
    map.addOperationalLayer(model.restaurantsLayer)
}
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+                    spatialReference: .wgs84
+                ),
+                scale: 4_000
+            )
+            map.addOperationalLayer(restaurantsLayer)
+            self.map = map
+        }
+        
+        /// Ensures the feature table is fully loaded before querying.
+        func ensureLayerLoaded() async throws {
```
philium: Needing to call this manually is most unfortunate. Instead, could it be done in `selectFeatures(in:screenPointFor:)`?

```swift
func selectFeatures(in polygon: Polygon?, screenPointFor: (Point) -> CGPoint?) async throws {
    if restaurantsTable.loadStatus != .loaded {
        try await restaurantsTable.load()
    }
    
    ...
}
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+            // Load the feature table to ensure metadata and features are available
```
philium: Remember to use proper punctuation in comments.


```suggestion
            // Load the feature table to ensure metadata and features are available.
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+        }
+        
+        /// Selects, numbers, and labels restaurant features that intersect a given polygon.
+        /// - Parameters:
+        ///   - polygon: The polygon used to query restaurant features.
+        ///   - screenPointFor: A closure that converts a map location to a screen point.
+        func selectFeatures(in polygon: Polygon?, screenPointFor: (Point) -> CGPoint?) async throws {
+            clearSelection()
+            
+            guard let polygon else { return }
```
philium: Taking an optional and then (almost) immediately guarding against non-optional is a bit of a code smell. Instead, I would suggesting changing `polygon` to non-optional and then change the caller to handle the `nil` polygon case:

```swift
if let polygon = makeSelectionPolygon(mapSize: mapSize, mapViewProxy: mapViewProxy) {
    do {
        try await model.selectFeatures(
            in: polygon,
            screenPointFor: { mapViewProxy.screenPoint(fromLocation: $0) }
        )
    } catch {
        self.error = error
    }
} else {
    model.clearSelection()
}
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+        func ensureLayerLoaded() async throws {
+            // Load the feature table to ensure metadata and features are available
+            try await restaurantsTable.load()
+        }
+        
+        /// Selects, numbers, and labels restaurant features that intersect a given polygon.
+        /// - Parameters:
+        ///   - polygon: The polygon used to query restaurant features.
+        ///   - screenPointFor: A closure that converts a map location to a screen point.
+        func selectFeatures(in polygon: Polygon?, screenPointFor: (Point) -> CGPoint?) async throws {
```
philium: Remember that on Mac Catalyst, `Polygon` is ambiguous because of [this](https://developer.apple.com/documentation/applicationservices/polygon) type, so it must be fully qualified everywhere that it is used: `ArcGIS.Polygon`. Please build on Mac Catalyst and address all build failures.

philium: Remember that parameter names should be nouns. A better name for `screenPointFor` might be `pointConverter`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+            guard let polygon else { return }
+            
+            let orderedFeatures = try await makeOrderedFeatures(
+                intersecting: polygon,
+                screenPointFor: screenPointFor
+            )
+            
+            hasMoreThanNineSelectedFeatures = orderedFeatures.count > Self.maximumNumberedFeatures
+            
+            // Only select the numbered features (first 9)
```
philium:
```suggestion
            // Only select the numbered features (first 9).
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+            let numberedOrderedFeatures = orderedFeatures.prefix(Self.maximumNumberedFeatures)
+            restaurantsLayer.selectFeatures(numberedOrderedFeatures.map(\.feature))
```
philium: We can take advantage of the fact that `selectFeatures(_:)` takes a sequence by _lazily_ mapping over the ordered features, thereby preventing the creation of an intermediate array.


```suggestion
            let numberedOrderedFeatures = orderedFeatures.lazy
                .map(\.feature)
                .prefix(Self.maximumNumberedFeatures)
            restaurantsLayer.selectFeatures(numberedOrderedFeatures)
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+            numberedFeatures.removeAll()
+            hasMoreThanNineSelectedFeatures = false
+        }
+        
+        /// Reads the feature's name attribute, returning the fallback when it is missing or blank.
+        /// - Parameters:
+        ///   - feature: The feature whose name is read.
+        ///   - fallback: The value to return when the feature has no name.
+        /// - Returns: The feature's name, or the fallback if no name exists.
+        func name(for feature: Feature, fallback: String?) -> String? {
```
philium: Consider giving `fallback` a default value of `nil` and then change `name(for: orderedFeature.feature, fallback: nil)` to `name(for: orderedFeature.feature)`.

Wait, I just noticed that the method returns an optional. Is `fallback` even needed then? The caller could just provide their own name if this method returns `nil` using the nil coalescing operator (`name(for: feature) ?? "My Name"`), correct?

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+            guard let name = feature.attributes[Self.nameAttribute] as? String,
+                  !name.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
+                return fallback
+            }
+            return name
```
philium: Did you mean to return the trimmed name?


```suggestion
            guard let name = feature.attributes[Self.nameAttribute] as? String else {
                return fallback
            }
            
            let trimmedName = name.trimmingCharacters(in: .whitespacesAndNewlines)
            
            guard trimmedName != "" else {
                return fallback
            }
            
            return trimmedName
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+        }
+        
+        /// Queries and sorts restaurant features from top-to-bottom, then left-to-right in screen space.
+        /// - Parameters:
+        ///   - polygon: The polygon used to query restaurant features.
+        ///   - screenPointFor: A closure that converts a map location to a screen point.
+        /// - Returns: The ordered restaurant features with their map and screen positions.
+        private func makeOrderedFeatures(
+            intersecting polygon: Polygon,
```
philium: Remember that for [factory methods](https://www.swift.org/documentation/api-design-guidelines/#init-factory-phrase-ends-with-basename), "The first argument…should not form a phrase starting with the base name." Therefore the first parameter shouldn't have an argument label and the name of the parameter should be something like `selection`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+            screenPointFor: (Point) -> CGPoint?
```
philium: The name of this parameter doesn't follow the naming guidelines—it should be a noun. Consider naming it `pointConverter`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,225 @@
...
+                    }
+                    return OrderedFeature(feature: feature, anchor: anchor, screenPoint: screenPoint)
+                }
+                .sorted { lhs, rhs in
+                    // Sort by visual reading order: top-to-bottom, then left-to-right.
+                    if lhs.screenPoint.y != rhs.screenPoint.y {
+                        return lhs.screenPoint.y < rhs.screenPoint.y
+                    }
+                    return lhs.screenPoint.x < rhs.screenPoint.x
+                }
```
philium: This code creates three separate arrays in order to return one:

1. Explicitly converting the sequence of features from the query result to an array
2. Compact mapping
3. Sorting

The first array is entirely unnecessary. And if we could lazily compact map over the features, then the second array wouldn't be necessary either. The problem is, lazily mapping captures the pointer converter closure. We could make the closure escaping, but we know that the closure will never escape the method. Thankfully, Swift has a solution for that: [`withoutActuallyEscaping(_:do:)`](https://developer.apple.com/documentation/swift/withoutactuallyescaping(_:do:)). We could take advantage of it to work with sequences until we need to create an array for the sorted elements.


```suggestion
            
            return withoutActuallyEscaping(screenPointFor) { escapableScreenPointFor in
                queryResult
                    .features()
                    .lazy
                    .compactMap { feature in
                        guard let anchor = feature.geometry as? Point,
                              let screenPoint = escapableScreenPointFor(anchor) else {
                            return nil
                        }
                        return OrderedFeature(feature: feature, anchor: anchor, screenPoint: screenPoint)
                    }
                    .sorted { lhs, rhs in
                        // Sort by visual reading order: top-to-bottom, then
                        // left-to-right.
                        return if lhs.screenPoint.y != rhs.screenPoint.y {
                            lhs.screenPoint.y < rhs.screenPoint.y
                        } else {
                            lhs.screenPoint.x < rhs.screenPoint.x
                        }
                    }
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Tips.swift
```diff
@@ -0,0 +1,70 @@
...
+    /// A TipKit tip that explains the centered rectangle.
+    struct AreaOfInterestTip: Tip {
+        var title: Text {
+            Text("Use the rectangle as the search area")
+        }
+        
+        var message: Text? {
+            Text(
+                 """
+                 To focus the map, press any arrow key. Then use the arrow keys ← → ↑ ↓ to pan the map until the restaurants you want to inspect are inside the rectangle.
```
philium: You might consider breaking up this line.


```suggestion
                 To focus the map, press any arrow key. Then use the arrow keys \
                 ← → ↑ ↓ to pan the map until the restaurants you want to inspect \
                 are inside the rectangle.
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Tips.swift
```diff
@@ -0,0 +1,70 @@
...
+        
+        var title: Text {
+            Text("Enable Full Keyboard Access")
+        }
+        
+        var message: Text? {
+            Text(
+                """
+                To use a hardware keyboard with your mobile device you need to enable Full Keyboard Access.
+                You can do this by opening the Settings app and navigating to:
```
philium:
```suggestion
                To use a hardware keyboard with your mobile device you need to \
                enable Full Keyboard Access. You can do this by opening the \
                Settings app and navigating to:
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    @State private var statusMessage = ""
+    
+    /// A task for the delayed selection refresh after navigation ends.
+    @State private var refreshTask: Task<Void, Never>?
+    
+    /// A token used to reactivate the UIKit key-command responder.
+    @State private var keyboardCommandCaptureActivation = 0
+    
+    /// The side length of the centered area-of-interest rectangle, in screen points.
+    private let selectionRectangleLength: CGFloat = 360
```
philium: Better to make this a computed property to get rid of an ivar.


```suggestion
    private var selectionRectangleLength: CGFloat { 360 }
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    
+    /// The number keys that can identify features.
+    private let featureNumberKeys = CharacterSet(charactersIn: "123456789")
+    
+    /// The fraction of the map size used for each keyboard pan step.
+    private let panStepRatio: CGFloat = 0.2
```
philium: This could be computed as well.

```suggestion
    private var panStepRatio: CGFloat { 0.2 }
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    
+    /// A tip explaining how to enable Full Keyboard Access.
+    private let enableKeyboardAccessTip = EnableKeyboardAccessTip()
```
philium: Is this necessary? Or could it just be created at the point of use?

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+                    .selectionColor(Model.selectionHaloColor)
+                    .callout(placement: $calloutPlacement.animation(.default.speed(2))) { _ in
+                        if let calloutFeature {
+                            makeCalloutContent(feature: calloutFeature)
+                        }
+                    }
+                    .onDrawStatusChanged { drawStatus in
+                        guard drawStatus == .completed, !initialDrawCompleted else { return }
+                        initialDrawCompleted = true
+                        Task {
```
philium: A good rule of thumb is to only use `Task` in a view when a task modifier won't work. In this case, you could use a task modifier with `initialDrawCompleted` as the identifier:

```swift
.onDrawStatusChanged { newDrawStatus in
    guard newDrawStatus == .completed, !initialDrawCompleted else { return }
    initialDrawCompleted = true
}
```

```swift
.task(id: initialDrawCompleted) {
    guard initialDrawCompleted else { return }
    do {
        try await model.ensureLayerLoaded()
        await refreshSelection(mapSize: mapSize, mapViewProxy: mapViewProxy)
        await focusMap()
    } catch is CancellationError {
        // Do nothing.
    } catch {
        self.error = error
    }
}
```

You may find that the sleep is no longer necessary (hopefully!). If it turns out to still be necessary, you could add it back in.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+                            }
+                        }
+                    }
+                    .onNavigatingChanged { navigating in
+                        refreshTask?.cancel()
+                        if navigating {
+                            dismissCallout()
+                        } else if initialDrawCompleted {
+                            // Debounce: wait for the map to fully settle after navigation.
+                            refreshTask = Task {
```
philium: Similarly hear, you could add an `isNavigating` state property, update it in this modifier, and then handle changes to it in a `task` modifier.

```swift
@State private var isNavigating: Bool?
```

```swift
.onNavigatingChanged { newIsNavigating in
    isNavigating = newIsNavigating
}
```

```swift
.task(id: isNavigating) {
    guard let isNavigating else { return }
    if isNavigating {
        dismissCallout()
    } else if initialDrawCompleted {
        await refreshSelection(
            mapSize: mapSize,
            mapViewProxy: mapViewProxy
        )
    }
}
```

Then `refreshTask` is not needed.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+                            RoundedRectangle(cornerRadius: 4)
+                                .fill(.pink.opacity(0.08))
+                                .overlay(
+                                    RoundedRectangle(cornerRadius: 4)
+                                        .stroke(.pink, lineWidth: 2)
+                                )
+                                .frame(
+                                    width: rectangleLength(for: mapSize),
+                                    height: rectangleLength(for: mapSize)
+                                )
```
philium: No need to compute `rectangleLength` multiple times.

```suggestion
                            let rectangleLength = self.rectangleLength(for: mapSize)
                            RoundedRectangle(cornerRadius: 4)
                                .fill(.pink.opacity(0.08))
                                .overlay(
                                    RoundedRectangle(cornerRadius: 4)
                                        .stroke(.pink, lineWidth: 2)
                                )
                                .frame(width: rectangleLength, height: rectangleLength)
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+                        // TipKit configuration is intended to happen once per process.
+                        try? Tips.configure([.displayFrequency(.immediate)])
+                    }
+                    .task {
+                        await focusMap()
+                    }
+                    .onChange(of: scenePhase) { _, newPhase in
+                        guard newPhase == .active else { return }
+                        Task { await focusMap() }
+                    }
```
philium:
```suggestion
                    .task(id: scenePhase) {
                        guard scenePhase == .active else { return }
                        await focusMap()
                    }
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,459 @@
...
+                    .task {
+                        await focusMap()
+                    }
+                    .errorAlert(presentingError: $error)
+            }
+        }
+    }
+    
+    /// Refreshes the selected and numbered restaurant features for the current rectangle.
+    private func refreshSelection(mapSize: CGSize, mapViewProxy: MapViewProxy) async {
```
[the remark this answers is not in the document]
philium: The proxy is a stand-in for the map view itself, so the name of the `mapViewProxy` parameter could just be `mapView`. Same for other methods.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+        
+        guard let targetCenter = mapViewProxy.location(fromScreenPoint: targetScreenPoint) else {
+            return
+        }
+        await mapViewProxy.setViewpointCenter(targetCenter)
+        await focusMap()
+    }
+    
+    /// Maps a pressed number key to a zero-based feature index.
+    private func featureIndex(for characters: String) -> Int? {
```
philium: A better name for this method would be `index(ofFeatureForCharacters:)`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+        guard let lastCharacter = characters.last,
+              let number = lastCharacter.wholeNumberValue,
+              (1...9).contains(number) else {
+            return nil
+        }
+        return number - 1
+    }
+    
+    /// Shows the details callout for the selected numbered feature.
+    private func showCalloutForFeature(at index: Int) {
```
philium: A better name for this method would be `showCallout(forFeatureAtIndex:)`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+        
+        let feature = model.numberedFeatures[index]
+        statusMessage = ""
+        calloutFeature = feature
+        calloutPlacement = .geoElement(feature, tapLocation: anchor)
+        if isKeyboardInputActive {
+            keyboardInputHasFocus = true
+        } else {
+            Task { await focusMap() }
+        }
```
philium: Please don't do work in the else clause of a guard statement. It breaks the mental model.


```suggestion
if model.numberedFeatures.indices.contains(index),
   let anchor = model.numberedFeatures[index].geometry as? Point {
    let feature = model.numberedFeatures[index]
    statusMessage = ""
    calloutFeature = feature
    calloutPlacement = .geoElement(feature, tapLocation: anchor)
    if isKeyboardInputActive {
        keyboardInputHasFocus = true
    } else {
        Task { await focusMap() }
    }
} else {
    statusMessage = "No restaurant is assigned to \(index + 1)."
    calloutFeature = nil
    calloutPlacement = nil
}
```

Also, I recommend making the method asynchronous and letting the caller handle that how they want. For example, the `onChange(of: keyboardInput)` modifier could be replaced with a task modifier instead.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    }
+    
+    /// Dismisses the details callout and restores the selection rectangle.
+    private func dismissCallout() {
```
philium: Please make this method asynchronous and let the caller handle the asynchronicity.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+        calloutFeature = nil
+        calloutPlacement = nil
+        statusMessage = ""
+        Task { await focusMap() }
+    }
+    
+    /// Gives keyboard focus to the map after SwiftUI finishes the current update.
+    @MainActor
+    private func focusMap() async {
+        guard !isFullKeyboardAccessEnabled else {
```
philium: Since work is being done in the else clause, please turn this into an if…else statement.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+            keyboardCommandCaptureActivation += 1
+            return
+        }
+        mapHasFocus = false
+        await Task.yield()
+        try? await Task.sleep(for: .milliseconds(100))
```
philium: This very much seems like a hack. It may be necessary (I don't know), but at the very least a comment explaining why is warranted. I presume you've exhausted all other options?

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    /// Dismisses the details callout and restores the selection rectangle.
+    private func dismissCallout() {
+        calloutFeature = nil
+        calloutPlacement = nil
+        statusMessage = ""
+        Task { await focusMap() }
+    }
+    
+    /// Gives keyboard focus to the map after SwiftUI finishes the current update.
+    @MainActor
```
philium: These `MainActor` annotations are unnecessary because the methods are on a view and views are inherently constrained to the main actor.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+        mapHasFocus = true
+    }
+    
+    /// Shows the software keyboard by focusing the hidden number input field.
+    @MainActor
+    private func showKeyboard() async {
+        isKeyboardInputActive = true
+        mapHasFocus = false
+        keyboardInputHasFocus = false
+        // Give SwiftUI time to add the text field to the view hierarchy.
```
philium: Based on this comment, could `keyboardInputHasFocus` be set using an `onAppear` modifier on the text field?

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    /// Hides the software keyboard input bar and returns focus to the map.
+    private func hideKeyboard() {
+        keyboardInput = ""
+        keyboardInputHasFocus = false
+        isKeyboardInputActive = false
+        dismissCallout()
+    }
+    
+    /// A status message styled to float above the map.
+    private func makeStatusText(_ text: String) -> some View {
```
philium: This might make more sense as a modifier.

```swift
Text("Blah")
    .statusText()
```

```swift
extension View {
    func statusText() -> some View {
        ...
    }
}
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+        Text(text)
+            .font(.footnote)
+            .multilineTextAlignment(.center)
+            .padding(8)
+            .background(.regularMaterial)
+            .clipShape(.rect(cornerRadius: 8))
+    }
+    
+    /// A hidden text field used to receive software keyboard input.
+    private func makeKeyboardInputBar() -> some View {
```
philium: A parameterless method like this that returns a view would make more sense as a property (`keyboardInputBar`). After all, a view's body is a property, not a method. For methods that take parameters and return a view, no need to prefix with `make`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+                    Spacer()
+                    Button("Done") {
+                        hideKeyboard()
+                    }
+                }
+            }
+    }
+    
+    /// The callout content for a restaurant feature.
+    private func makeCalloutContent(feature: Feature) -> some View {
```
philium: A better name for this would be `calloutContent(for:)`.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+                Text("Lat: \(wgs84Point.y, format: .number.precision(.fractionLength(6)))")
+                Text("Lon: \(wgs84Point.x, format: .number.precision(.fractionLength(6)))")
+            }
+        }
+        .padding(5)
+    }
+    
+    /// Opens the Settings app to an Accessibility feature when supported.
+    private func openAccessibilitySettings() {
+        Task {
```
philium: Having the entire body of a method be a `Task` is also a code smell. It indicates that the method itself should probably be asynchronous. Then the caller can decide how they want to handle the asynchronicity.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+            "App-prefs:ACCESSIBILITY",
+            "App-prefs:root=ACCESSIBILITY",
+            "prefs:root=ACCESSIBILITY"
+        ]
+        
+        if let url = candidates.compactMap(URL.init).first(where: UIApplication.shared.canOpenURL) {
+            UIApplication.shared.open(url)
+        } else {
+            UIApplication.shared.open(.appSettings)
+        }
```
philium: Please use the `openURL` [environment value](https://developer.apple.com/documentation/swiftui/environmentvalues/openurl).

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,556 @@
...
+    let onPan: (CGVector) -> Void
+    let onEscape: () -> Void
+    let onNumber: (Int) -> Void
+    
+    func makeUIView(context: Context) -> ResponderView {
+        let view = ResponderView()
+        view.backgroundColor = .clear
+        view.onPan = onPan
+        view.onEscape = onEscape
+        view.onNumber = onNumber
```
philium: No need to set these as `updateUIView(_:context:)` will be called before the view appears.


```suggestion
```

philium (commented):
Thank you for implementing this sample! I have reviewed almost everything (except for `KeyboardCommandCaptureView`). Here is my initial feedback.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
chriswebb09: I tried a few things. I'll see if there is a better way to approach this.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift
```diff
@@ -0,0 +1,209 @@
...
+            return trimmedName
+        }
+        
+        /// Queries and sorts restaurant features from top-to-bottom, then left-to-right in screen space.
+        /// - Parameters:
+        ///   - selection: The polygon used to query restaurant features.
+        ///   - pointConverter: A closure that converts a map location to a screen point.
+        /// - Returns: The ordered restaurant features with their map and screen positions.
+        private func makeOrderedFeatures(
+            _ selection: Polygon,
```
philium: Please remove the underscore (and fix the call sites) as each parameter of a factory function should have an argument label.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,570 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+import TipKit
+import UIKit
```
philium: This import isn't needed.


```suggestion
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,570 @@
...
+        }
+        // On iPhone, a focus system exists only when Full Keyboard Access is on.
+        // Caveat: on iPad, a connected hardware keyboard alone creates a focus
+        // system, so this reads as a false positive there.
+        return UIFocusSystem.focusSystem(for: window) != nil
+    }
+    
+    var body: some View {
+        MapViewReader { mapViewProxy in
+            GeometryReader { geometryProxy in
```
philium: Instead of using a geometry reader, you could use an `onGeometryChange` modifier instead. It has the advantage of re-rendering the view only when the desired information changes rather than when every property changes. First, add a state property for the map size:

```swift
@State private var mapSize: CGSize = .zero
```

Then add the modifier to the map view reader:

```swift
.onGeometryChange(for: CGSize.self, of: \.size) { newMapSize in
    mapSize = newMapSize
}
```

Then simply remove the geometry reader.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,570 @@
...
+        min(selectionRectangleLength, mapSize.width, mapSize.height)
+    }
+    
+    /// Refreshes the selected and numbered restaurant features for the current rectangle.
+    private func refreshSelection(mapSize: CGSize, mapViewProxy: MapViewProxy) async {
+        if let polygon = makeSelectionPolygon(mapSize: mapSize, mapViewProxy: mapViewProxy) {
+            do {
+                try await model.selectFeatures(
+                    in: polygon,
+                    pointConverter: { mapViewProxy.screenPoint(fromLocation: $0) }
```
philium: FYI: You could pass the method directly without an intermediate closure.


```suggestion
                    pointConverter: mapViewProxy.screenPoint(fromLocation:)
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,570 @@
...
+            return nil
+        }
+        return number - 1
+    }
+    
+    /// Shows the details callout for the selected numbered feature.
+    private func showCallout(forFeatureAtIndex index: Int) {
+        if model.numberedFeatures.indices.contains(index),
+           let anchor = model.numberedFeatures[index].geometry as? Point {
+            let feature = model.numberedFeatures[index]
```
philium, line 225: You could use [gosh darn if case let syntax](https://goshdarnifcaseletsyntax.com) to bind the non-optional feature as part of the if statement to avoid subscripting twice.


```suggestion
           case let feature = model.numberedFeatures[index],
           let anchor = feature.geometry as? Point {
```

philium (commented):
Don't forget to rename all uses of `mapViewProxy` with `mapView`. I think I mentioned that in one of my previous comments?

Samples.xcodeproj/project.pbxproj
```diff
@@ -345,9 +345,13 @@
...
 		95321B322E554B88002E9DE2 /* ApplyRenderersToSceneLayerView.swift in Copy Source Code Files */ = {isa = PBXBuildFile; fileRef = 95321B2F2E554B20002E9DE2 /* ApplyRenderersToSceneLayerView.swift */; };
 		9537AFD72C220EF0000923C5 /* ExchangeSetwithoutUpdates in Resources */ = {isa = PBXBuildFile; fileRef = 9537AFD62C220EF0000923C5 /* ExchangeSetwithoutUpdates */; settings = {ASSET_TAGS = (ConfigureElectronicNavigationalCharts, ); }; };
+		95398E6F2FDFAF1600E30CE8 /* NavigateMapAndIdentifyFeaturesWithKeyboardView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 95398E6E2FDFAF1400E30CE8 /* NavigateMapAndIdentifyFeaturesWithKeyboardView.swift */; };
+		95398E712FDFAF6300E30CE8 /* NavigateMapAndIdentifyFeaturesWithKeyboardView.swift in Copy Source Code Files */ = {isa = PBXBuildFile; fileRef = 95398E6E2FDFAF1400E30CE8 /* NavigateMapAndIdentifyFeaturesWithKeyboardView.swift */; };
+		95398E732FE0C78B00E30CE8 /* NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift in Sources */ = {isa = PBXBuildFile; fileRef = 95398E722FE0C78800E30CE8 /* NavigateMapAndIdentifyFeaturesWithKeyboardView.Model.swift */; };
```
CalebRas, line 187: Due to the changes in https://github.com/Esri/arcgis-maps-sdk-swift-samples/pull/800, you'll have to merge `v.next` into this branch and remove all of the lines added to `Samples.xcodeproj/project.pbxproj` that have "Copy Source Code Files" in them before this PR merged, just FYI. Let me know if you have any issues with that.

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,572 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ArcGIS
+import SwiftUI
+import TipKit
+
+
+struct NavigateMapAndIdentifyFeaturesWithKeyboardView: View {
```
philium, line 19: Remember to carefully double check any changes you make before you commit and push them. An extra blank line snuck in here.


```suggestion
import TipKit

struct NavigateMapAndIdentifyFeaturesWithKeyboardView: View {
```

Samples.xcodeproj/project.pbxproj
chriswebb09: Okay, I think this was done correctly, let me know if you see any issues.

CalebRas: It looks like there are still some extra lines in the project file diff. Every line with words "Copy Source Code Files" in it will have to be removed.

philium (dismissed):

Shared/Samples/Navigate map view and identify features with keyboard/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Use case
+
+Keyboard access is an important part of building inclusive GIS applications. Users who cannot or prefer not to use touch or pointer input need a way to move around the map, discover features, and read feature details from the keyboard.
+
+## How to use the sample
+
+Use keyboard navigation to pan and zoom until restaurants appear inside the rectangle. Restaurants inside the rectangle are selected and labeled in reading order, from top-to-bottom and left-to-right.
+
+Press number keys `1` through `9` to show details for the matching restaurant. Press `Esc` to dismiss the callout and show the selection rectangle again.
```
pgruenler, line 15: I couldn't get the `Esc` key to dismiss the callout. Does a button need to be added for that or does the readme need to be updated?

pgruenler (commented):
Looks good and is working well, just left one question.

Shared/Samples/Navigate map view and identify features with keyboard/README.md
chriswebb09: Updated

pgruenler (approved):

philium (approved):

yo1995 (approved):
Approve to get around GitHub bug, assuming it is approved by other reviewers

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,302 @@
...
+    private var selectionRectangleLength: CGFloat { 360 }
+    
+    /// The selection halo color for selected restaurant features.
+    private static let selectionHaloColor = Color(
+        red: 190 / 255,
+        green: 24 / 255,
+        blue: 93 / 255
+    )
+    
+    @State private var isNavigating: Bool = false
```
yo1995, line 68:
```suggestion
    /// A Boolean value indicating whether the map view is currently navigating.
    @State private var isNavigating = false
```

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,302 @@
...
+            calloutPlacement = .geoElement(feature, tapLocation: anchor)
+        } else {
+            statusMessage = "No restaurant is assigned to \(index + 1)."
+            calloutFeature = nil
+            calloutPlacement = nil
+        }
+    }
+    
+    /// Dismisses the details callout and restores the selection rectangle.
+    private func dismissCallout() async {
```
yo1995, line 242:
```suggestion
    private func dismissCallout() {
```
This method doesn't need to be async

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,302 @@
...
+                                .statusText()
+                        }
+                    }
+                    .padding(.bottom, 10)
+                }
+                .onAppear {
+                    // TipKit configuration is intended to happen once per process.
+                    try? Tips.configure([.displayFrequency(.immediate)])
+                    
+                    guard map.operationalLayers.isEmpty else { return }
```
yo1995, line 177: Why is this check needed?

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
```diff
@@ -0,0 +1,302 @@
...
+        min(selectionRectangleLength, mapSize.width, mapSize.height)
+    }
+    
+    /// Refreshes the selected and numbered restaurant features for the current rectangle.
+    private func refreshSelection(mapSize: CGSize, mapView: MapViewProxy) async {
+        if let polygon = makeSelectionPolygon(mapSize: mapSize, mapView: mapView) {
+            do {
+                try await model.selectFeatures(
+                    in: polygon,
+                    pointConverter: { mapView.screenPoint(fromLocation: $0) }
```
yo1995, line 198:
```suggestion
                    pointConverter: mapView.screenPoint(fromLocation:)
```

yo1995 (commented):
noticed a few things when I reviewed #802

Shared/Samples/Navigate map view and identify features with keyboard/NavigateMapAndIdentifyFeaturesWithKeyboardView.swift
chriswebb09: I'll open separate PR to address your comments.

PR #788, opened 2026-06-16
[Update] Changes for TipKit

chriswebb09:
## Description
This adds in the TipKit change that was initially part of the Sample PR. This change was requested. 

## Linked Issue(s)

- `swift/8270`

## How To Test

Test that hint text appears in Tip and then goes away after close button is tapped.

## Screenshot
tipkit-change


Shared/Samples/Update basemap for contrast accessibility/README.md
```diff
@@ -20,8 +20,9 @@ Switch to manual mode to choose Light, Dark, High Contrast Light, or High Contra
 2. Resolve which contrast appearance should be active based on the current mode and device settings.
     * In manual mode, use the appearance selected in the supporting pane.
     * In automatic mode, resolve the appearance from the device's current light, dark, and increased-contrast settings using SwiftUI environment values (for example, `@Environment(\.colorScheme)` and `@Environment(\.colorSchemeContrast)`).
-3. Map the resolved appearance to an ArcGIS Online `Basemap` or a `BasemapStyle` and create a `Map`.
-4. Apply the current reference-layer visibility setting to the basemap's labels and boundary layers.
+3. Display a TipKit tip when automatic mode is selected to guide users to test the sample by changing device appearance settings.
```
chriswebb09, line 23: @yo1995 added in here

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -183,6 +269,7 @@ private extension UpdateBasemapForContrastAccessibilityView {
             }
         }
         
+
```
yo1995:
```suggestion
```

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -86,54 +88,138 @@ private extension UpdateBasemapForContrastAccessibilityView {
...
+                        let feature: AccessibilitySettings.Feature = .assistiveTouch
+                        try await AccessibilitySettings.openSettings(for: feature)
+                    } else if #available(iOS 18.0, *) {
+                        let feature: AccessibilitySettings.Feature = .personalVoiceAllowAppsToRequestToUse
+                        try await AccessibilitySettings.openSettings(for: feature)
+                    } else {
+                        openAppSettings()
+                    }
+                } catch {
+                    print("Unable to open Accessibility Settings: \(error)")
```
yo1995: We don't print in the samples. Please use `errorAlert` if you want to display the error to the user.

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -86,54 +88,138 @@ private extension UpdateBasemapForContrastAccessibilityView {
...
+                    openAppSettings()
+                }
+            }
+        }
+        
+        /// Opens this app's page in the Settings app.
+        private func openAppSettings() {
+            UIApplication.shared.open(.appSettings)
+        }
```
yo1995, line 179: Mac Catalyst will throw an error and do nothing with the `appSettings` URL. Consider one of the following:

- Only show the button for iOS/iPadOS with compiler condition block `#if !targetEnvironment(macCatalyst)`
- Use one of the `x-apple.systempreferences` links specifically for Mac to open the system settings

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -12,17 +12,25 @@
...
 import ArcGIS
 import SwiftUI
+import TipKit
+import UIKit
 
 struct UpdateBasemapForContrastAccessibilityView: View {
+    /// The one-time TipKit configuration for this sample.
+    private static let tipConfiguration: Void = {
+        try? Tips.configure([.displayFrequency(.immediate)])
+    }()
```
yo1995: I don't think this is the correct way of calling `Tips.configure`. Where did you see this pattern?

The doc for [static func configure([Tips.ConfigurationOption]) throws](https://developer.apple.com/documentation/tipkit/tips/configure(_:)) says: 

> This function must be called before tips display in your app.

From the examples in Apple Developer Documentation that I see, it can either be called in the [initializer](https://developer.apple.com/documentation/tipkit/tips/configure(_:)), or the [`task` modifier](https://developer.apple.com/videos/play/wwdc2023/10229/?time=811). Presumably, it can also be called in `onAppear`.

```swift
.onAppear {
    do {
        try Tips.configure([.displayFrequency(.immediate)])
    } catch {
        self.error = error
    }
}
```

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -86,54 +88,138 @@ private extension UpdateBasemapForContrastAccessibilityView {
...
+        
+        var title: Text {
+            Text("Try changing device appearance")
+        }
+        
+        var message: Text? {
+            Text(
+                "Change Light/Dark Mode or Increase Contrast in the Settings app " +
+                "to see the basemap update automatically. Increase Contrast is in " +
+                "Accessibility > Display & Text Size."
```
yo1995:
```suggestion
                """
                Change Light/Dark Mode or Increase Contrast in the Settings app
                to see the basemap update automatically. Increase Contrast is in
                Accessibility > Display & Text Size.
                """
```
You can use [multiline string literals](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/#Multiline-String-Literals).

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -86,54 +88,138 @@ private extension UpdateBasemapForContrastAccessibilityView {
...
+        }
+        
+        var actions: [Action] {
+            return [
+                Action(
+                    id: Self.openSettingsActionID,
+                    title: "Open Accessibility Settings"
+                )
+            ]
+        }
```
yo1995, line 212:
```suggestion
        var actions: [Action] {
            Action(
                id: Self.openSettingsActionID,
                title: "Open Accessibility Settings"
            )
        }
```
The magic of builder pattern in Swift! 🪄✨

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -86,54 +88,138 @@ private extension UpdateBasemapForContrastAccessibilityView {
...
+                        Text("Visual Contrast Mode")
+                    } footer: {
+                        VStack(alignment: .leading, spacing: 8) {
+                            Text(model.contrastMode.detail)
+                            
+                            if model.contrastMode == .automatic {
+                                TipView(automaticModeTip) { action in
+                                    if action.id == AutomaticModeTip.openSettingsActionID {
+                                        openAccessibilitySettings()
+                                    }
```
yo1995:
```suggestion
                                openAccessibilitySettings()
```
Can this be simplified? As there is only 1 action for the tip.

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -86,54 +88,138 @@ private extension UpdateBasemapForContrastAccessibilityView {
...
+        /// Opens the Settings app to an Accessibility feature when supported.
+        private func openAccessibilitySettings() {
+            Task {
+                do {
+                    if #available(iOS 26.0, *) {
+                        let feature: AccessibilitySettings.Feature = .assistiveTouch
+                        try await AccessibilitySettings.openSettings(for: feature)
+                    } else if #available(iOS 18.0, *) {
+                        let feature: AccessibilitySettings.Feature = .personalVoiceAllowAppsToRequestToUse
+                        try await AccessibilitySettings.openSettings(for: feature)
```
yo1995: As both `openSettings(for: feature)` lead towards the same page, I don't see the need to call open settings on different features. Since there is no way to open the "Display & Text Size" sub menu, I would suggest both call `personalVoiceAllowAppsToRequestToUse`, and add a comment to explain that we call this feature to get to the accessibility menu.

more thoughts:

- `assistiveTouch` isn't a good option. When it opens that setting, it goes deep into the Accessibility > Touch > AssistiveTouch sub menu, which makes it harder to find the actual toggle we want Accessibility > Display & Text Size
- I tried on both my phone and iPad with OS 26, and I don't see the Accessibility > [Personal Voice](https://support.apple.com/en-us/104993) sub menu for `personalVoiceAllowAppsToRequestToUse`. I'm confused by that, but I think this is less bad than `assistiveTouch` as the page open is the root Accessibility menu.
- It would be great if Apple natively support opening the "Display & Text Size" sub menu without using private API or pref link 😬

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
chriswebb09: I'll update to the onAppear.

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
chriswebb09: Good catch, that was debugging and should have been removed.

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
chriswebb09: Awesome will update

Shared/Samples/Update basemap for contrast accessibility/UpdateBasemapForContrastAccessibilityView.swift
```diff
@@ -33,25 +36,33 @@ struct UpdateBasemapForContrastAccessibilityView: View {
...
     var body: some View {
         MapView(map: model.map)
-            .task(id: effectiveAppearance) {
+            .onAppear {
+                do {
+                    try Tips.configure([.displayFrequency(.immediate)])
+                } catch {
+                    self.error = error
+                }
+            }
```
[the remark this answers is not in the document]
yo1995: Well, I think AI has its point. But since the sample's view will only appear once, I don't think the problem it mentioned would have effect in our case.

yo1995 (approved):
LGTM! Thanks for adding these improvements.

des12437 (approved):

PR #802, opened 2026-08-18
[Fix] for mac catalyst

chriswebb09:
Fix for polygon namespace collision.


CalebRas (approved):
Builds on all platforms for me.

yo1995 (approved):

